PR #130, opened 2015-01-10
shutdown client resources

tonyvanriet:
Hi Zach,

I noticed that my application wasn't able to exit after using an aleph client. I found that I needed to call shutdownGracefully on the shared EventLoopGroup used by the clients. It looks like you had this call originally but it was removed when you changed to the shared event group.

This PR adds a convenience function for this shutdown, along with some documentation, if you're interested.

I have another branch that, instead, configures this shared event group to use daemon threads so they will be killed automatically on application exit. But, I'm assuming you would not want to preclude aleph from being used in applications where the clients are intended to continue after the main application exits. If I'm wrong about that, let me know and I'll submit that PR.

Thanks.
  -Tony


ztellman:
I think I prefer the daemon thread approach, it's what I used in the previous version.

PR #139, opened 2015-02-01
HTTP connection and request timeouts

bts:
I've added support for connection and request timeouts to `http/request`. In this branch there are now three timeouts:

* the existing `socket-timeout`, which limits pool generation time
* `connection-timeout`, which limits time to establish a connection post pool-generation
* `request-timeout`, which limits time for response arrival post connection-establishment

I added these because I was getting deferreds that never yielded a value for unavailable hosts, and I couldn't add a "hard" cumulative request timeout via middleware (without violating the pool contract).

In the cases of connection and request timeouts, I dispose the pooled connection because we won't know when data might be coming back from the server. I wasn't sure if there was a more intelligent way to handle this.

Additionally I've introduced reporting of a `:socket-time` separate from `:connection-time` which you may want to use.

I was also wondering if `socket-timeout` should possibly be renamed to something like `pool-timeout`? Since "socket timeout" typically refers to a limit on the amount of time without data flowing over the socket.

Let me know if there are any adjustments I can make.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -206,9 +206,12 @@
                     (-> (middleware conn')
                       (d/chain #(% req))
                       (maybe-timeout! request-timeout)
-                      (d/catch TimeoutException
-                               #(do (flow/dispose pool k conn) (throw %)))
-                      (d/catch #(do (flow/release pool k conn) (throw %)))
+                      (d/catch
+                        (fn [e]
+                          (if (instance? TimeoutException e)
+                            (flow/dispose pool k conn)
+                            (flow/release pool k conn))
+                          (throw e)))
                       (d/chain
                         (fn [rsp]
                           (d/chain (:aleph/complete rsp)
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -193,28 +193,31 @@
       (maybe-timeout! socket-timeout)
       (d/chain
         (fn [conn]
-          (-> (first conn)
-            (maybe-timeout! connection-timeout)
-            (d/catch TimeoutException
-                     #(do (flow/dispose pool k conn) (throw %)))
-            (d/chain
-              (fn [conn']
-                (let [connection-end (System/currentTimeMillis)
-                      connection-time (- connection-end start)]
-                  (-> (middleware conn')
-                    (d/chain #(% req))
-                    (maybe-timeout! request-timeout)
-                    (d/catch TimeoutException
-                             #(do (flow/dispose pool k conn) (throw %)))
-                    (d/catch #(do (flow/release pool k conn) (throw %)))
-                    (d/chain
-                      (fn [rsp]
-                        (d/chain (:aleph/complete rsp)
-                          (fn [_]
-                            (flow/release pool k conn)))
-                        (-> rsp
-                          (dissoc :aleph/complete)
-                          (assoc :connection-time connection-time))))))))))))))
+          (let [socket-end (System/currentTimeMillis)
+                socket-time (- socket-end start)]
+            (-> (first conn)
+              (maybe-timeout! connection-timeout)
+              (d/catch TimeoutException
+                       #(do (flow/dispose pool k conn) (throw %)))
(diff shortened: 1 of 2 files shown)


src/aleph/http.clj
```diff
@@ -164,35 +166,58 @@
...
+              (d/chain
+                (fn [conn']
+                  (let [connection-end (System/currentTimeMillis)
+                        connection-time (- connection-end socket-end)]
+                    (-> (middleware conn')
+                      (d/chain #(% req))
+                      (maybe-timeout! request-timeout)
+                      (d/catch TimeoutException
+                               #(do (flow/dispose pool k conn) (throw %)))
+                      (d/catch #(do (flow/release pool k conn) (throw %)))
```
bts: Is releasing here always safe in the case of streaming responses?

src/aleph/http.clj
```diff
@@ -164,35 +166,61 @@
...
+                  (let [connection-end (System/currentTimeMillis)
+                        connection-time (- connection-end socket-end)]
+                    (-> (middleware conn')
+                      (d/chain #(% req))
+                      (maybe-timeout! request-timeout)
+                      (d/catch
+                        (fn [e]
+                          (if (instance? TimeoutException e)
+                            (flow/dispose pool k conn)
+                            (flow/release pool k conn))
```
bts: Is releasing here always safe in the case of streaming responses?

ztellman: If we've timed out after acquiring the connection, we should close the connection.  Also, I'm a little unclear on the distinction between `socket` and `connection` timeouts.  Most of the time we'd expect that we're receiving a live connection from the pool.  Can you clarify your mental model for this?

ztellman: Oh, sorry, didn't read the top-level description closely enough.  I think `pool-timeout` is much better than `socket-timeout`.

bts: Definitely agree on closing the connection after timing out.

When it took a long time to get a connection, it was helpful for me to see the breakdown of how long it took for the pool to generate (>100 ms) vs the amount of time to acquire a usable connection subsequent to that. The mental model was that when the pool generates, that’s roughly when the connection was initiated vs when the connection is established.

On February 1, 2015 at 2:34:01 PM, Zach Tellman (notifications@github.com) wrote:

In src/aleph/http.clj:

> +            (-> (first conn)
> +              (maybe-timeout! connection-timeout)
> +              (d/catch TimeoutException
> +                       #(do (flow/dispose pool k conn) (throw %)))
> +              (d/chain
> +                (fn [conn']
> +                  (let [connection-end (System/currentTimeMillis)
> +                        connection-time (- connection-end socket-end)]
> +                    (-> (middleware conn')
> +                      (d/chain #(% req))
> +                      (maybe-timeout! request-timeout)
> +                      (d/catch
> +                        (fn [e]
> +                          (if (instance? TimeoutException e)
> +                            (flow/dispose pool k conn)
> +                            (flow/release pool k conn))
If we've timed out after acquiring the connection, we should close the connection. Also, I'm a little unclear on the distinction between socket and connection timeouts. Most of the time we'd expect that we're receiving a live connection from the pool. Can you clarify your mental model for this?

ztellman:
Can I get a single merged commit for all these changes?  I'm having a hard time getting a full picture of what's changed.

bts:
Yeah I will clean this up and make the suggested changes. I need to head out right now but I can probably take a look at this later today.

On February 1, 2015 at 2:51:36 PM, Zach Tellman (notifications@github.com) wrote:

Can I get a single merged commit for all these changes? I'm having a hard time getting a full picture of what's changed.

src/aleph/http.clj
ztellman: When starting a new connection, the pool will return a new connection immediately, before it's been fully established.  For subsequent calls, it will return a connection which is immediately usable.  I agree it's a useful distinction, but not one that will be important for every request.

bts:
See cleaned-up PR #140

PR #145, opened 2015-02-11
Fix bug in HeaderMap's keys implementation

bcobb:
The sequence `(keys (dissoc header-map "some-key"))` threw an exception because `removed` was a set, and not (as the old code seems to assume) a map.


ztellman:
Thanks for catching this.

PR #154, opened 2015-02-26
added body coercion copied from clj-http

shmish111:
Not sure how you feel about all this dynamic ns loading etc and how it fits in to what you want aleph to be but here it is, pretty much a copy and paste from clj-http.  It's what I and I guess a lot of others need.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -463,12 +463,12 @@
   (if-let [m (re-matches #"\s*(([^/]+)/([^ ;]+))\s*(\s*;.*)?" (str s))]
     {:content-type (keyword (nth m 1))
      :content-type-params
-                   (->> (str/split (str (nth m 4)) #"\s*;\s*")
-                        (identity)
-                        (remove str/blank?)
-                        (map #(str/split % #"="))
-                        (mapcat (fn [[k v]] [(keyword (str/lower-case k)) (str/trim v)]))
-                        (apply hash-map))}))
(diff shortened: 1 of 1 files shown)


ztellman:
Yeah, this is pretty reasonable.  However, there are a lot of whitespace changes.  I just added `cljfmt` to the project, can you merge with my changes, and run `lein cljfmt` on yours?  Thanks.

shmish111:
Is there anything else you would like me to do for this PR?

shmish111:
Also, I'm using https://github.com/hozumi/clj-det-enc locally to detect the character encoding, would you mind if I added that here?  The issue being it's another dependency for the project...

ztellman:
I'd prefer it be an optional dependency like the others, if possible.

ztellman:
Actually, looking at that code, I'm not sure how it works with HTTP requests, since it'll consume the head of the InputStream.  Can you expand on how that would be used here?

shmish111:
I've added the code to the PR.  It does mean you have to specify if you want an InputStream, rather than being the default.

ztellman:
I'd honestly prefer it be left out.  If it belongs anywhere, it belongs in `byte-streams`.  Let me give it some thought.

shmish111:
Indeed it would be nice in `bs/to-string`, I want that function to do as much for me as possible.  Would it adversely affect performance though?  If so it could be an option `:detect true`...  Also, I think I will change the default coercion to be none at all and require `:as :string` if you want a String.  What do you think?

ztellman:
Yes, I'd prefer no coercion to be the default.

PR #176, opened 2015-09-24
Do not generate Content-Length headers on 1xx and 204 responses

Leonidas-from-XIV:
Consider this example, running with the current git version https://gist.github.com/Leonidas-from-XIV/c4a6fb02bf8cd8288f5e, where Aleph returns an invalid header value. Basically, it takes the length of the response (9), adds it as a header and then I think Netty throws the body rightfully away, without adjusting for the headers. This creates issues with HTTP clients that read this header and wait indefinitely to get these 9 bytes that will never arrive (yes, that happened to us).

The proper behaviour (returning Content-Length: 0 in 204 requests or leaving the header out) is somewhat disputed on the internet, but RFC 7230, section 3.3.2 is relatively clear about it:

> A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content).

Hope this PR is of use to you, let me know if there are any changes necessary to get this in!

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -305,13 +305,18 @@
 (defn send-contiguous-body [ch ^HttpMessage msg body]
   (let [body (if body
                (DefaultLastHttpContent. (netty/to-byte-buf ch body))
-               empty-last-content)]
-    (let [code (-> msg .getStatus .code)]
-      (cond
-        (<= 100 code 199) nil
-        (= 204 code) nil
-        :else
-        (HttpHeaders/setContentLength msg (-> ^HttpContent body .content .readableBytes))))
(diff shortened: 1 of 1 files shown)


ztellman:
The `instance?` check should be for `HttpResponse`, not `DefaultHttpReponse`.  Also, using `cond` in this case seems a bit odd, why not just have a `when` with a predicate that does both checks?

Thanks for catching this, though.

Leonidas-from-XIV:
I kinda like using `cond` in a pattern matching with guards manner, handling the cases separately and allowing fallthrough instead of `or` logic, but I can see that is is a rather unusual, so I've replaced it with `when-not`.

PR #178, opened 2015-10-10
Fix spelling of operations in tcp.clj

cddr:
Minor typo I found while reading the docs


ztellman:
Thanks for catching this.

PR #180, opened 2015-10-25
Support setting netty channel options

jkytomak:
So that you can eg. use bigger UDP buffer size:
http://stackoverflow.com/questions/13086564/netty-increase-channelbuffer-size

### DIFF
--- a/src/aleph/udp.clj
+++ b/src/aleph/udp.clj
@@ -27,6 +27,14 @@
 
 (alter-meta! #'->UdpPacket assoc :private true)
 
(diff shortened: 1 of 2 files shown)


ztellman:
I'd rather just have a generic `bootstrap-transform` function which defaults to `idenitty`, since that mirrors the Netty configuration options elsewhere.  Any problem with that?

jkytomak:
Sure, should work fine. I have only used the udp part of aleph and did not know of that convention.

PR #192, opened 2015-11-16
close body after writing and flushing the final chunk

### LINKED ISSUE #191: Intermittent Incomplete Chunked Encoding Errors on various clients
Can reproduce with clj-http pretty consistently. It seems to fail probabilistically.

aamedina:
fixes  #191


ztellman:
I figured out the underlying problem, it should be fixed in beta3.

Issue #186, opened 2015-10-30
StringIndexOutOfBoundsException from aleph.http.client-middleware/generate-query-string


pbwolf:
user> (require '[aleph.http.client-middleware :as ahcm])
nil
user> (ahcm/generate-query-string {:a "3"})
StringIndexOutOfBoundsException String index out of range: -1  java.lang.AbstractStringBuilder.substring (AbstractStringBuilder.java:904)

Expected result

user> (ahcm/generate-query-string {:a "3"})
"a=3"

Looking at the code, I think there are two problems. (1) The "(if (sequential? v)" test is not needed: it has no "else", so it causes non-collection values to be skipped, and anyway the "true" branch has its own if-sequential test and seems to be able to deal with strings. (2) The concluding substring invocation uses the arity-1 version, which starts at an offset, but the generate-query-string function's intent is to select all but the last character, so it should use the arity-2 version with 0 as the first argument.

This is with respect to Aleph 0.4.1-beta1.

ztellman:
Since this was a major (stupid) bug, I've marked a new release for this fix.  Thanks again for the report.

PR #195, opened 2015-11-30
Fix HTTP client StringIndexOutOfBoundsException when empty query params are passed

gsnewmark:
Related to #186

`StringIndexOutOfBoundsException` is still thrown when empty `params` map is passed to the `generate-query-string`:

```
> (require '[aleph.http.client-middleware :as ahcm])
nil
> (ahcm/generate-query-string {})
StringIndexOutOfBoundsException String index out of range: -1  java.lang.AbstractStringBuilder.substring (AbstractStringBuilder.java:908)

```

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -313,16 +313,18 @@
     "UTF-8"))
 
 (defn generate-query-string [params & [content-type]]
-  (let [encoding (detect-charset content-type)
-        sb (StringBuffer.)]
-    (p/doit [[k v] params]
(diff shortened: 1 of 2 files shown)


Gonzih:
It's not fixed yet, right? So solution for now is to skip parameters completely?

ztellman:
The change has been merged, but I haven't cut a new release since then.  I'll get that done today, sorry for the delay.

ztellman:
I've just cut `0.4.1-beta4`.  I apologize this took so long, before I checked I had assumed this was already in a previous release.

Gonzih:
Thanks!

PR #209, opened 2016-02-04
add max-frame-payload option to http/websocket-connection (fixes #208)

### LINKED ISSUE #208: no ability to change netty's websocket maxFramePayloadLength
We're happily using aleph as a websocket server for a React Native client, however, React Native's WebSocket doesn't split large messages across multiple frames and we have a need to send large messages to the server.

It would be nice if we could dodge the more fundamental issue of our WebSocket library not being great at what it does by just bumping up netty's WebSocket `maxFramePayloadLength`, but there doesn't appear to be a user-accessible api to do so

nyoung-figly:
does what it says, little else


ztellman:
I'd like the doc-string for this parameter to read "maximum allowable frame payload length, in bytes" (all lower-case, for consistency with the other parameters).  Otherwise, looks good.

nyoung-figly:
done. amended the original commit.

PR #220, opened 2016-03-08
bootstrap-transform in client expects an io.netty.bootstrap.Bootstrap…

apa512:
… object, not io.netty.bootstrap.ServerBootstrap


ztellman:
Thanks for the correction.

PR #223, opened 2016-03-10
shutdown group if server start fails

thheller:
the group would otherwise leak threads when a start is retried.

See https://github.com/ztellman/aleph/issues/222


thheller:
Not sure if the `deref` of the future returned by `.shutdownGracefully` is required but it seemed logical.

thheller:
CI tests seem broken, tests pass fine locally.

PR #234, opened 2016-04-17
fix for aleph server dropping WebSocket continuation frames

nyoung-figly:
recently ran into an issue where aleph would drop continuation frames of WebSocket messages: the handler passed to `stream/consume` wasn't being invoked with the entire message and wasn't being applied to any frames other than the first one sent by the client.

Seems like aleph could solve this by handling `ContinuationWebSocketFrame`s in `http.server/websocket-server-handler`'s `:channel-read` callback, or by inserting a `WebSocketFrameAggregator` into the channel pipeline as part of `http.server/initialize-websocket-handler`.

The first approach would require aleph to allocate and manage some temporary storage before a final frame is received, while the second approach leans on Netty having already implemented that.

With `:pipeline-transform` its possible for users to add this handler themselves without any change to aleph, but it feels like this sort of behavior should be turned on out of the box. Happy to discuss further.


ztellman:
Hi, sorry for the delay in merging this.  Can you update it against the latest master?  Everything looks good, otherwise.

nyoung-figly:
done & force updated the branch.

PR #252, opened 2016-06-09
websocket client: max-frame-payload, continuation frame support

nyoung-figly:
almost identical set of changes to those made on the server side to support large frame payloads and aggregating continuation frames


ztellman:
Thanks!

PR #255, opened 2016-06-24
make PTR record lookups in UDP functionality optional

rwat:
This change maintains default behavior and backwards compatibility but allows for avoiding DNS PTR lookups per UDP packet.


ztellman:
I think I'd prefer to just have it never look it up, and expose the raw InetSocketAddress object as an additional key.

rwat:
Were you thinking something like this?

PR #258, opened 2016-07-24
Adds support for form param coercion

solatis:
As discussed in ztellman/aleph#257, here you will find a PR that integrates clj-http's support for json, edn and transit form param coercion.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,7 +9,8 @@
                  [byte-streams "0.2.2"]
                  [potemkin "0.4.3"]]
   :profiles {:dev {:dependencies [[org.clojure/clojure "1.8.0"]
-                                  [criterium "0.4.4"]]}}
+                                  [criterium "0.4.4"]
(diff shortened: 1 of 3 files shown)


solatis:
For what it's worth, the client-middleware module is out of sync with clj-http a bit. No more decorator macro magic, among others. I think it's a relatively small effort to backport this into aleph, and would add quite a few additional middleware, but will probably require some refactoring. What do you think?

ztellman:
Hey, sorry for the slow reply on this.  I support any work which brings client-middleware up to date with clj-http, if you're willing to do it.

solatis:
Okay, I've just aligned the client middleware with cli-http again. All decorator and related `defdecorator` macro code has been refactored into the most current cli-http middleware.

At one point, it's probably a good idea to move to "real" middleware, since the middleware is currently just only transforming the request object. This should be tightly coupled with manifold, though, and that definitely belongs in a different issue/PR.

PR #259, opened 2016-07-26
Fix :or map - keys are bound symbols, not lookup keys

puredanger:


ztellman:
Thanks Alex.  Since I see you're doing a number of these small things, are you testing out some sort of new code validation, maybe using spec?

puredanger:
Yeah, I'm running with a version of Clojure with spec'ed macros for let, ns, defn, etc. since these things will fail to compile after that's added, I'm trying to flush out some of that earlier.

Issue #262, opened 2016-08-01
Broken link in documentation


traviswsims:
Hello,

The [examples page](http://aleph.io/examples/literate.html) has a line that reads "Complete documentation for the aleph.http namespace can be found [here](http://ideolalia.com/aleph/aleph.http.html).", but the link gives a 404.

Thanks for your effort and a great library!

ztellman:
This has been fixed.

PR #263, opened 2016-08-01
Update broken link #262

Devinsuit:


ztellman:
This should be updated in the source under `/examples`, not the generated artifact.

Issue #273, opened 2016-09-22
Add nanotime to ring-request map before queueing?


jeroenvandijk:
We want to take the time a request is waiting into account when processing a request. Would you be open for a patch to add `:request-time (System/nanoTime)`[here](https://github.com/ztellman/aleph/blob/9cdd682bd8e1827ae560139f1b21c61640f7d17c/src/aleph/http/core.clj#L180)?

Measuring on ring level would mean we miss the first part of queueing.

ztellman:
The intended way to track this is via the executor stats, which do capture time spent on the queue, but this would be pretty harmless.  I'd prefer `:request-arrived` or something to that effect, though.

PR #275, opened 2016-10-06
Added :request-arrived field with nano time to request map

jeroenvandijk:
This is a very straight forward attempt to #273.

I'm had a look at the tests to see what I would have to add there, but I couldn't find something similar. 

Please let me know if you had more in mind.

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -194,6 +194,7 @@
                     (.substring uri (unchecked-inc question-mark-index))))
   :headers (-> req .headers headers->map)
   :request-method (-> req .getMethod .name str/lower-case keyword)
+  :request-arrived (System/nanoTime)
   :body body
   :scheme (if ssl? :https :http)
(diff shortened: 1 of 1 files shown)


ztellman:
This seems fine, though I'm tempted to namespace the key under `aleph`, just to differentiate between the Ring keys and non-Ring keys.  If I did this, I should also do it for `:keep-alive?`.  I'm going to give a little thought.

jeroenvandijk:
The namespace prefix sounds like a good idea. Without the prefixing it's not always clear where a map entry in the ring map comes from.

One other thing I considered is using a proxy function instead of `(System/nanoTime)` directly. This would help during testing as you would be able to override the nano time and test for specific behaviour. Not sure if it would add some unnecessary overhead. This could also be achieved in other ways so I'm not certain adding it on Aleph level would be very beneficial, but as an integration test it could be useful.

ztellman:
I've decided to namespace both of them.  Thanks for your contribution, apologies it took so long to merge.

PR #283, opened 2016-11-18
Support custom client SSL contexts.

csm:
My use-case is setting up client-side certificates for TCP connections. I haven't bothered to look at the tests to see if I should add anything there yet.

* src/aleph/netty.clj (check-ssl-args): new function.
  (ssl-client-context): add docstring, optional kwargs. Set up private key, certificate chain if requested.
* src/aleph/tcp.clj (client): add ssl-context argument.


ztellman:
Thanks, much appreciated.

ztellman:
Just a heads up: I've changed the keyword args to `ssl-client-context` to a single options map, since that's more consistent with the rest of Aleph.

PR #296, opened 2017-02-19
Fix netty/set-logger! to switch on logger parameter.

mahinshaw:


ztellman:
Thanks for catching that.

Issue #270, opened 2016-09-04
close codes for websocket client?


vspinu:
How to get more info when server closes websocket connection. Particularly, is there a way to retrieve the [close code](https://tools.ietf.org/html/rfc6455#section-7.4.1)? 

Thanks.

ztellman:
Right now there isn't a good means of conveying that information.  I could add it as a deferred in the stream's metadata, I guess.  Thoughts?

vspinu:
I would think that a natural place for those is as an argument to the `on-closed` callback, but that's not backward compatible, or is it?  If it's added as a metadata to the stream, it would be useful to have the stream itself as an argument to the callback.

eslachance:
I would like to +1 this, as receiving the connection code can be important. I'm dealing with the Discord API which returns specific codes that define clearly why the socket was closed: 

![image](https://cloud.githubusercontent.com/assets/643344/22757246/18db8c4a-ee18-11e6-9671-42e2622287e8.png)

Knowing the reason for the disconnect in this case is critical to know what action should be taken. Perhaps an `:on-error` event?

ubercow:
I think having it in the metadata or description of the stream would probably be best. Metadata being the change that'd require the fewest amount of changes to implement.

Aside from being a breaking change, adding an argument to on-closed would also require a bump of manifold too.

Gonna hack together a fork that uses metadata real quick and see how it looks/works.

ubercow:
https://github.com/ztellman/aleph/compare/master...ubercow:websocket-close-metadata
In theory by the time on-closed is received you'd be able to access this metadata.
Hacked this together, but I really don't like it.

I was thinking of adding a new field to `netty/ChannelSink`, an `(atom {})` that'd be merged into the description. But I'd love to hear @ztellman's opinions before I go crazy changing core APIs.

The other thing on my mind is how we could support close codes for websocket servers too. But for that I don't think there's any way around changing manifold to add an argument to close. Perhaps that's best for another PR.

PR #297, opened 2017-02-25
Set websocket close info in description

ubercow:
This is the first draft of adding the ability to get websocket close-codes when using `aleph.http.client/websocket-connection`.

This adds a new arity to `aleph.netty/sink` which allows you to pass a function that's called when the description is generated. This is then merged into the default description that `aleph.netty/ChannelSink` provides.

The existing arities of `aleph.netty/sink` simply provide a function that returns nil, ensuring backwards compatibility for anyone other implementations of sink in aleph or by a 3rd party.

`aleph.http.client/websocket-client-handler` was modified to take advantage of this new arity to provide additional information when a Netty `CloseWebSocketFrame` is passed.

On a graceful websocket close, the `statusCode` and `reasonText` will be set in `:websocket-close-code` and `:websocket-close-msg` respectively on the `:sink` of the stream.

See the related issue #270 for more details on this problem.

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -380,6 +380,7 @@
 (defn websocket-client-handler [raw-stream? uri sub-protocols extensions? headers max-frame-payload]
   (let [d (d/deferred)
         in (atom nil)
+        desc (atom {})
         handshaker (websocket-handshaker uri sub-protocols extensions? headers max-frame-payload)]
 
     [d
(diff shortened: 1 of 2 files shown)


ztellman:
Sorry it took so long to review this, this looks like a really reasonable solution to the problem, thank you.

ubercow:
👍 

Thanks for all your hard work!

PR #303, opened 2017-03-20
Fix memory leak with binary websocket frames (in websocket clients)

ryuuseijin:
The same fix as in https://github.com/ztellman/aleph/pull/302 but for websocket clients instead of websocket servers.

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -444,10 +444,10 @@
                (netty/put! ch @in (.text ^TextWebSocketFrame msg))
 
                (instance? BinaryWebSocketFrame msg)
-               (let [frame (netty/acquire (.content ^BinaryWebSocketFrame msg))]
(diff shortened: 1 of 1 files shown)


ryuuseijin:
We've had this fix in production for about a week, and it's looking good. We are still getting the occasional "LEAK" message, but now orders of magnitude less. Also, memory has been completely stable across several peak cycles (~2000 messages per second) where before there was a visible slope ending in a JVM OOM error.

The remaining "LEAK" seems to be either extremely uncommon or maybe even a false positive. I'm not sure where it's coming from. We're using websockets heavily, in binary mode to communicate between aleph clients and aleph servers and in text mode to communicate with browser clients and aleph servers. We are also using streams and deferreds heavily. Will continue to track this.

ztellman:
Thanks, I'll cut a new alpha release with this fix tomorrow.

PR #305, opened 2017-03-28
Fix example extending compojure Renderable with correct IDeferred interface

yonatane:
In the compojure example of delayed-handler, manifold.deferred.Deferred doesn't exist. Changed to IDeferred.


ztellman:
Thanks, I've pushed the changes to aleph.io.

Issue #272, opened 2016-09-12
Sending a request with Accept-Encoding: deflate, gzip won't produce a compressed response


uaalto:
A request using the mentioned header doesn't produce a response with Content-Encoding: gzip or deflate as expected. I've tried with this setting, but doesn't change much:

```
:pipeline-transform #(doto % (.addLast "deflater" (io.netty.handler.codec.http.HttpContentCompressor.)))
```

I don't see the deflater being added to the pipeline in Aleph's code anywhere. If this is supposed to be default in Netty, what does it need to be triggered otherwise?

Thanks @ztellman !

uaalto:
Also tried with:

```
:pipeline-transform #(doto %
                                                               (.addLast "decoder" (io.netty.handler.codec.http.HttpRequestDecoder.))
                                                               (.addLast "encoder" (io.netty.handler.codec.http.HttpRequestEncoder.))
                                                               (.addLast "inflater" (io.netty.handler.codec.http.HttpContentDecompressor.))
                                                               (.addLast "deflater" (io.netty.handler.codec.http.HttpContentCompressor.)))
```

uaalto:
In order for this to work, I had to use:

~~~
#(doto % (.addAfter "http-server" "deflater" (HttpContentCompressor. 1)))
~~~

Seems reasonable, but what about adding this to the pipeline by default. It's quite standard.
Also, do you think it would make sense to move https://github.com/ztellman/aleph/blob/master/src/aleph/http/server.clj#L398 up the pipeline before the handler?
Well, I see the benefits of having it this way too, so any function added to pipeline-transform can reference anything behind in the pipeline, and be placed wherever is required.

I would just add that the only inconvenience here is that for such an important feature, the way to enable it is not documented and not easily discoverable unless familiar with Netty.

ztellman:
Hi, sorry for the late reply to this.  I don't think this is something that should be enabled by default, but I think there's a strong argument to be made for adding a parameter to `start-server` which exposes this as an option.

Patches are welcome, or I'll get to it myself.

uaalto:
Hi @ztellman !
Thanks  for your reply. What about adding it simply as documentation? That would also work.

ztellman:
I'd be fine with a documentation patch, too.

kachayev:
Closing this one as compression is already supported since https://github.com/ztellman/aleph/commit/d137b5f4657478b5401bcb504e560e13bcfc5cff

PR #310, opened 2017-04-08
Enable server side compression

truemped:
Hi!

I'm am trying to learn more about aleph and netty but it's the first time I'm looking into that. So this might be more of a request for comments - if time allows - about my clojure code. I essentially saw #272 as a low hanging fruit that could help me understand the it.

Right now it still feels kind of hacky. Do you see obvious areas of improvement? Could I somehow combine the `compression?` key with the `pipeline-transform` without loosing the possibility to configure compression?

Daniel


ztellman:
This works well enough for now.  Sorry it took me so long to get to this.

PR #312, opened 2017-04-21
Support :multipart in HTTP clients requests

kachayev:
@ztellman Covering issue: https://github.com/ztellman/aleph/issues/159

Unit tests are included. 

API is slightly extended from the previous implementation to provided better compatibility with `clj-http` client.

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -315,12 +315,11 @@
                 (when-not (.get (.headers req') "Connection")
                   (HttpHeaders/setKeepAlive req' keep-alive?))
 
-                ;; TODO: uncomment this once the multipart implementation is validated
-                (let [body (if-let [parts (comment (get req :multipart))]
+                (let [body (if-let [parts (get req :multipart)]
                              (multipart/encode-body parts)
                              (get req :body))]
                   (netty/safe-execute ch
-                    (http/send-message ch true ssl? req' body))))
+                                      (http/send-message ch true ssl? req' body))))
 
               ;; this will usually happen because of a malformed request
               (catch Throwable e
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -36,11 +36,12 @@
                (when file?
                  (URLConnection/guessContentTypeFromName (.getName ^File content))))
         ;; populate file name when working with file object
-        filename (or name (when file? (.getName ^File content)))]
-    {:part-name part-name
+        filename (or name (when file? (.getName ^File content)))
+        ;; use "name" as a part name when the last is not provided
+        part-name-to-use (or part-name name filename)]
+    {:part-name part-name-to-use
      :content (bs/to-byte-buffer content)
-     ;; do not need to specify charset when sending file
-     :mime-type (mime-type-descriptor mt (when-not file? charset))
+     :mime-type (mime-type-descriptor mt charset)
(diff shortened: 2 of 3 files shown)


ztellman:
Thanks for pushing this over the finish line.

royalaid:
Not sure if I am using it wrong but making a multipart call against httpbin.org doesn't seem to work. Adding the boundary by works for getting it to identify the files but the actual content never seems to propagate.

kachayev:
@royalaid Would you please share the code that you're executing?

royalaid:
Apologies for taking so long to respond, work got busy and had to set down my side project using this for a while, code is below:
```clojure
(http/post "http://httpbin.org/post" 
           {:multipart [{:content (java.io.File. "./test.jpg")}])})
```
I based it off the test cases that have been pushed with this PR. I would expect the return value to have files in the in the body.

kachayev:
@royalaid Actually, I cannot reproduce this. Using `file.txt` from `tests`:

```clojure
user=> (-> (http/post "http://httpbin.org/post" {:multipart [{:content (java.io.File. "./test/file.txt")}]}) deref :body bs/to-string)
"{\n  \"args\": {}, \n  \"data\": \"--e5ff6b0d683514bc\\ncontent-disposition: form-data; name=\\\"file.txt\\\"; filename=\\\"file.txt\\\"\\ncontent-type: text/plain\\nthis is a file\\n\\n--e5ff6b0d683514bc--\", \n  \"files\": {}, \n  \"form\": {}, \n  \"headers\": {\n    \"Connection\": \"close\", \n    \"Content-Length\": \"149\", \n    \"Host\": \"httpbin.org\"\n  }, \n  \"json\": null, \n  \"origin\": \"194.44.212.15\", \n  \"url\": \"http://httpbin.org/post\"\n}\n"```

**Comment by royalaid** [comment]
The result should look like
```json
{
  "args": {},
  "data": "",
  "files": {},
  "form": {
    "file": "Users/Mark/Downloads/test.txt"
  },
  "headers": {
    "Accept": "*/*",
    "Connection": "close",
    "Content-Length": "168",
    "Content-Type": "multipart/form-data; boundary=------------------------d6179dc74eff1298",
    "Expect": "100-continue",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.51.0"
  },
  "json": null,
  "origin": "68.4.160.177",
  "url": "http://httpbin.org/post"
}
```
according to curl via `curl -v -X POST -F file=~/Downloads/test.txt http://httpbin.org/post`
Versus the result from aleph.multipart:

```json
{
  "args" : { },
  "data" : "--2ab91ebd5b1e45a1\ncontent-disposition: form-data; name=\"file.txt\"; filename=\"file.txt\"\ncontent-type: text/plain\nthis is a file\n\n--2ab91ebd5b1e45a1--",
  "files" : { },
  "form" : { },
  "headers" : {
    "Connection" : "close",
    "Content-Length" : "149",
    "Host" : "httpbin.org"
  },
  "json" : null,
  "origin" : "68.4.160.177",
  "url" : "http://httpbin.org/post"
}
```

kachayev:
@royalaid Ah, actually you're right. It's not an issue by RFC, but most tools rely on "Content-Type" header which is not set here by `aleph` (and there's no easy way to set header manually as far random boundary generation is not visible for the user). Let me upgrade implementation slightly.

PR #318, opened 2017-05-10
Expose max-queue-size setting for a connection-pool factory method

kachayev:
Right now `http/connection-pool` function creates `flow/instrumented-pool` with default `max-queue-size` (65536, which is pretty significant for a queue size).

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -87,6 +87,8 @@
    | `total-connections` | the maximum number of connections across all hosts
    | `target-utilization` | the target utilization of connections per host, within `[0,1]`, defaults to `0.9`
    | `stats-callback` | an optional callback which is invoked with a map of hosts onto usage statistics every ten seconds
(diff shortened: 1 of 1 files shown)


ztellman:
When the queue overflows, trying to make a call to `http/get` may throw a `RejectedExecutionException`.  Is that really the behavior you're looking for?

kachayev:
@ztellman Correct. That's useful for example when implementing a proxy server with long-running requests to downstream: to have backpressure mechanism and not try to accept too many requests at the same time. Implementing any kind of a bounded queue/buffered stream as an additional layer seems like an unnecessary overhead as far as this part of thread pool functionality already.

ztellman:
Okay, I have no objection to making this configurable, I just wanted to make sure this was actually a good solution to your problem.  I also have no objection to sub-typing `TimeoutException` to differentiate between the different timeout types.

kachayev:
Cool! I'll implement `TimeoutException` sub-typing as a separate PR.

PR #321, opened 2017-05-20
For 204 response the body should be empty...

stoyle:
...not an empty ByteArrayInputStream.

When proxying through the server a body here will result in a
`Transfer-Encoding:chunked` in the aleph server, basically stating the body size
is unknown. Some caching products, like varnish, fails miserably here, e.g. gives
the calling client back an http 503.

With this change, at least when the body is empty it is removed, and the
`Transfer-Encoding:chunked` is not added.

This seems to be working. I guess there is a deeper root cause here, somewhere in Netty perhaps? I would be happy to change this in any way you'd like, or attack the problem differently. Let me know.

Cheers,
Alf

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -532,6 +532,17 @@
       (-> (client req)
         (d/chain' #(assoc % :request-time (- (System/currentTimeMillis) start)))))))
 
+(defn wrap-204-empty-body
+  "Middleware which removes a body from a 204 request when the body is actually not there.
+  If the 204 response erroneously has a body with data, the body will be preserved."
+  [client]
+  (fn [req]
+    (-> (client req)
+        (d/chain' (fn [resp]
+                    (if (and (= 204 (:status resp)) (= 0 (.available (:body resp))))
+                      (dissoc resp :body)
+                      resp))))))
+
 (defn parse-content-type
   "Parse `s` as an RFC 2616 media type."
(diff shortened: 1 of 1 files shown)


danielcompton:
I'd be interested in hearing @malcolmsparks thoughts on this, as Yada tries very hard to handle HTTP semantics well. I think Yada only sets a 204 on `yada.methods` when there isn't a body so it wouldn't hit this?

stoyle:
I am not too familiar with Yada, but if you are wrapping both aleph's client and server, and haven't done anything to prevent this, I think you will have the same issue.

I guess this is a small issue, but it was very confusing when I ran into it. If others run into it in the future, there is at least a solution on the alephs mailing list (https://groups.google.com/forum/#!topic/aleph-lib/00pJj6s_UE0).

ztellman:
This is only a problem when you're proxying through a `204` response from Aleph's HTTP client, I don't think it should affect Yada much.

However, this fix assumes the body is always an `InputStream`, which won't be true when `raw-stream?` is true.  It may be better to fix this a little further upstream, I'm happy to take care of it myself if you're not comfortable digging in there.

danielcompton:
Ah right, I completely missed that this is a client 204, not a server 204 response, sorry.

stoyle:
Go ahead, looking forward to seeing how to fix it properly. It was fun digging a bit in Aleph's code.

PR #323, opened 2017-05-29
Add IDN URLs support to HTTP client

gsnewmark:
Currently it's not possible to make HTTP request to [IDN URLs](https://en.wikipedia.org/wiki/Internationalized_domain_name) using the Aleph because it relies on `java.net.URI` to [extract domain from the URL](https://github.com/ztellman/aleph/blob/9dcfc60321700ce271a41fc4765671f6f3bef5eb/src/aleph/http/client.clj#L58-L79) and `java.net.URI` doesn't really support IDNs: depending on constructor used it either silently ignores such domain names (`.getHost` returns `null`) or fails with exception (current Aleph behavior): 
```
user> (require '[aleph.http.client :as client])
nil
user> (aleph.http.client/req->domain {:url "https://名がドメイン.com"})
URISyntaxException Illegal character in hostname at index 8: https://名がドメイン.com  java.net.URI$Parser.fail (URI.java:2848)
user> *e
#error {
 :cause "Illegal character in hostname at index 8: https://名がドメイン.com"
 :via
 [{:type java.net.URISyntaxException
   :message "Illegal character in hostname at index 8: https://名がドメイン.com"
   :at [java.net.URI$Parser fail "URI.java" 2848]}]
 :trace
 [[java.net.URI$Parser fail "URI.java" 2848]
  [java.net.URI$Parser parseHostname "URI.java" 3387]
  [java.net.URI$Parser parseServer "URI.java" 3236]
  [java.net.URI$Parser parseAuthority "URI.java" 3155]
  [java.net.URI$Parser parseHierarchical "URI.java" 3097]
  [java.net.URI$Parser parse "URI.java" 3053]
  [java.net.URI <init> "URI.java" 673]
  [aleph.http.client$eval38639$req__GT_domain__38646 invoke "client.clj" 75]
  <REPL-related stacktrace...>
  ]}
```

I've added UTF to ASCII translation for hostname, but I'm not sure it's best possible solution, as with this approach there is one additional URL transformation for each request. What do you think about it?

On a side note, do you really need `java.net.URI` instance as a domain? If I don't miss anything it's used only as a [key for connection pool](https://github.com/ztellman/aleph/blob/9dcfc60321700ce271a41fc4765671f6f3bef5eb/src/aleph/http.clj#L213), wouldn't protocol+domain+port _string_ be enough for this purpose?

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -60,8 +60,7 @@
                (URI.
                  (name (or (:scheme req) :http))
                  nil
-                 (when-some [host (or (:host req) (:server-name req))]
-                   (IDN/toASCII host))
+                 (some-> (or (:host req) (:server-name req)) IDN/toASCII)
                  (or (:port req) (:server-port req) -1)
                  nil
                  nil
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -14,7 +14,8 @@
     [java.net
      URI
      URL
-     InetSocketAddress]
+     InetSocketAddress
+     IDN]
     [io.netty.buffer
      ByteBuf
      Unpooled]
@@ -59,19 +60,20 @@
                (URI.
                  (name (or (:scheme req) :http))
                  nil
-                 (or (:host req) (:server-name req))
+                 (when-some [host (or (:host req) (:server-name req))]
+                   (IDN/toASCII host))
                  (or (:port req) (:server-port req) -1)
                  nil
                  nil
                  nil))]
 
-  (defn req->domain [req]
+  (defn ^java.net.URI req->domain [req]
     (if-let [url (:url req)]
       (let [^URL uri (URL. url)]
         (URI.
           (.getProtocol uri)
           nil
-          (.getHost uri)
+          (IDN/toASCII (.getHost uri))
           (.getPort uri)
           nil
           nil
--- a/test/aleph/http/client_test.clj
+++ b/test/aleph/http/client_test.clj
@@ -1,0 +1,43 @@
+(ns aleph.http.client-test
+  (:require [aleph.http.client :as client]
+            [clojure.test :as t :refer [deftest is testing]]))
+
+(deftest test-domain-extracting
+  (testing "ASCII domain is extracted from URL as is."
+    (let [uri (client/req->domain {:url "https://domainname.com:80"})]
+      (is (= "https" (.getScheme uri)))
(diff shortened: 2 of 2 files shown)


src/aleph/http/client.clj
```diff
@@ -59,19 +60,20 @@
                (URI.
                  (name (or (:scheme req) :http))
                  nil
-                 (or (:host req) (:server-name req))
+                 (when-some [host (or (:host req) (:server-name req))]
```
danielcompton: A (subjective) matter of taste, but `some->` might be easier to read here. Will let the author of Elements of Clojure make the call though :)

test/aleph/http/client_test.clj
```diff
@@ -0,0 +1,43 @@
...
+                                   :port 80})]
+      (is (= "https" (.getScheme uri)))
+      (is (= "domainname.com" (.getHost uri)))
+      (is (= 80 (.getPort uri)))))
+  (testing "IDN domain in URL is translated to ASCII."
+    (let [uri (client/req->domain {:url "https://名がドメイン.com:80"})]
+      (is (= "https" (.getScheme uri)))
+      (is (= "xn--v8jxj3d1dzdz08w.com" (.getHost uri)))
+      (is (= 80 (.getPort uri)))))
+  (testing "IDN domain from request host is translated to ASCII."
```
danielcompton: Could be good to add a test for an [Internationalized TLD](https://en.wikipedia.org/wiki/Internationalized_country_code_top-level_domain) too.

src/aleph/http/client.clj
gsnewmark: agree, `some->` would definitely be more readable, didn't think about it

ztellman:
Netty expects a URI, so that's why I use it.  This looks good to me.

gsnewmark:
@ztellman thanks!

ztellman:
Is this blocking you? I can cut an alpha release, if so.

gsnewmark:
@ztellman not really, it affects only a small part of code and we have a workaround in place (basically transforming URL before passing it to Aleph). So no need for a new alpha. 

By the way, are there any plans for a new "stable" release or there are not enough changes yet?

ztellman:
There are a few changes I'm contemplating in Manifold w.r.t. to upstream propagation of closing, which are delaying the release of the whole suite of libraries.  I'll have a clearer answer for that in a week.

PR #331, opened 2017-09-15
Prevent the parsing of empty streams

acron0:
When trying to parse an empty response as `transit+json` an internal EOFException is thrown, regardless of the `:throw-exceptions` flag. When the stream is empty, we shouldn't even attempt to parse it.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -64,9 +64,10 @@
   "Resolve and apply Transit's JSON/MessagePack decoding."
   [in type & [opts]]
   {:pre [transit-enabled?]}
-  (let [reader (ns-resolve 'cognitect.transit 'reader)
-        read (ns-resolve 'cognitect.transit 'read)]
-    (read (reader in type (transit-read-opts opts)))))
+  (when (pos? (.available in))
+    (let [reader (ns-resolve 'cognitect.transit 'reader)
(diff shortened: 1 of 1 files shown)


ztellman:
The code which handles this is taken from clj-http, which is the basis for all the client semantics.  It's possible that I'm out of sync, but unless an empty string really is valid Transit (I know it's not valid JSON), this isn't necessarily the wrong thing to do.  I'd like any changes that are made here to first be vetted by @dakrone and merged into clj-http.

acron0:
@ztellman Fair enough, I shall await some guidance from @dakrone then :) Thanks

dakrone:
I think that empty responses should be treated as non-existent, I'm not sure, however, whether they should throw an exception or not. I can see merit in either direction. @ztellman do you have a preference in this case?

ztellman:
I think the "transforming nothing yields nothing" pattern is pretty well
established in Clojure, even though it's not the greatest thing for
debuggability. I think the expected behavior that prompted this issue is
reasonable.

acron0:
Would you like me to submit this to clj-http then?

dakrone:
@acron0 sure, please do

acron0:
@ztellman This change was merged into clj-http. Can it be merged here too?

ztellman:
Sure, thanks for chasing this down.

PR #334, opened 2017-11-07
fix in documentation

johannesloetzsch:


ztellman:
Thanks for catching that.

PR #336, opened 2017-11-20
Switch from using the netty-all JAR to specific dependencies

### LINKED ISSUE #335: Switch from netty-all to specifying Netty JARs?
I spent a bit of time tracking down a nasty coredumping issue which turned out to be caused because Aleph bundles netty-all, and the Google Cloud SDK depends on individual netty JARs.

As best as I can tell (https://github.com/netty/netty/issues/4671, [CASSANDRA-10984](https://issues.apache.org/jira/browse/CASSANDRA-10984)), netty-all is a kind of uberjar of all Netty classes, designed for people that just want to download a JAR and get running with Netty. If you're using Maven it is instead recommended to depend on specific Netty dependencies.

Now that I've figured out the problem I can just exclude the netty-all JAR to get the behaviour I want, but I thought it might save other people some anguish if Aleph moved to depending on explicit Netty JARs.

What are your thoughts on this? It looks like [eventually](https://github.com/netty/netty/issues/4671#issuecomment-257031451) netty-all may just become a dependency container for all of the other Netty JARs, in which case there would be no issue, however that seems like it may be a way off.

danielcompton:
This helps Maven handle dependency conflicts between different projects
that depend on Netty.

Fixes #335

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -4,7 +4,11 @@
                  "sonatype-oss-public" "https://oss.sonatype.org/content/groups/public/"}
   :license {:name "MIT License"}
   :dependencies [[org.clojure/tools.logging "0.3.1" :exclusions [org.clojure/clojure]]
-                 [io.netty/netty-all "4.1.11.Final"]
+                 [io.netty/netty-transport "4.1.11.Final"]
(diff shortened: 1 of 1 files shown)


ztellman:
Thanks!

PR #337, opened 2017-11-21
Leiningen 2.8.0 and JDK 9 compatibility

jgrodziski:
# Leiningen 2.8.0 introduces more strict checks on repositories
All repositories must now use TLS encryption (here the jboss one for Netty). Otherwise, we'll get the following error message when using leiningen:
```
Tried to use insecure HTTP repository without TLS.
This is almost certainly a mistake; however in rare cases where it's
intentional please see `lein help faq` for details.
```

# Class removed in JDK 9 was used in function encoding the basic auth header
The basic-auth-value function used the javax.xml.DatatypeConverter class (removed in JDK 9) for encoding the "username:password" value of the basic auth header. The fix use the Base64 class instead. Add a test case for ensuring the behavior was exactly the same before and after the fix (even if I'm doubtful about the original behavior on the limit case like nil or with missing arg ?).
**ONE CAVEAT**: the Base64 class was introduced in JDK 8. Zach if you think a lot of people use Aleph with JDK 6 or 7 we must introduce a check to get which class is available in the runtime env...Tell me in that case and I'll update the pull request accordingly.


ztellman:
I'd prefer to target the `Base64` class in Netty, which won't have any versioning issues.

jgrodziski:
Ok, done. The basic-auth-value function implementation now use Netty's Base64 implementation.
I bump Netty's patch version and dev's clojure a the same time.

ztellman:
Great, thanks for your contribution.

jgrodziski:
Thank you Zach for your huge and great work with all your libraries!

camsaul:
@ztellman we're running into issues with Java 9 support as well (unless you add the `--add-modules=java.xml.bind` to the JVM) because of the usage of `javax.xml.bind.DatatypeConverter` in aleph 0.4.4. It looks like it's fixed in this PR. Any chance you could push out a 0.4.5 or 0.4.4.1 release with this fix?

ztellman:
I've pushed `0.4.5-alpha1` with all the recent changes.

camsaul:
Thanks!

PR #339, opened 2017-11-25
Advanced timeout exceptions

kachayev:
`aleph.http/request` function accepts for now 3 types of timeouts: `pool-timeout`, `connection-timeout` and `request-timeout`.  The problem is that further during the request processing it's impossible to distinguish what actually happened when `j.u.c.TimeoutException` occurs (all 3 cases are handled in the same manner). It probably doesn't harm a lot with basic scenarios, but implementing something like a proxy server you have to know if the request was actually sent to a downstream or not.

Subclasses are used not to break existing code. Test cases are updated.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -29,6 +29,7 @@
             [lein-jammin "0.1.1"]
             [lein-marginalia "0.9.0"]
             [ztellman/lein-cljfmt "0.1.10"]]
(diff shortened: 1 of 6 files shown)


ztellman:
This looks useful, thank you.

kachayev:
@ztellman Sure!

PR #341, opened 2017-12-10
http/request read-timeout to prevent connections leak

kachayev:
`http/request` now accepts optional `read-timeout` param to throw `ReadTimeoutException` when it takes too long for the response to be completed preventing potential connections leak.


ztellman:
Looks good, thanks.

PR #343, opened 2017-12-12
Enhanced content-transfer-encoding handling for multipart requests

kachayev:
There are quite a few changes. What was done here:

1. Fixed `base64` encoding. Setting `:transfer-encoding` to `base64` right now throws an exception. And as far as I see it never worked properly. Conversion between Netty and NIO ByteBuf is not optimal in terms of performance in new implementation, but I did it following the same pattern as it's done for user/pass encoding (to prevent problems with different JDK versions).

2. `:binary` encoding could be specified in case we want to send data to the server "as is" explicitly marking “Content-Transfer-Encoding: binary” (allowed by the RFC).

There are also a few changes that might break someone's code. Which is highly unlikely as this functionality was essentially broken, but I’m open to a conversation here.

1. `aleph.http.encoding/encode` function throws an `IllegalArgumentException` when unsupported encoding is given (was: return `nil`) Silent fallback to `nil` as an output value might lead to very clumsy issues.

2. We do not do any encoding if `transfer-encoding` is not specified explicitly (was: default to `quoted-printable`) to match default behavior of most clients, like CURL. `quoted-printable` is rarely used nowadays and using it leads to almost doubling the size of the request containing binary data, such as pictures or video.

### DIFF
--- a/src/aleph/http/encoding.clj
+++ b/src/aleph/http/encoding.clj
@@ -1,11 +1,14 @@
 (ns aleph.http.encoding
   (:require
-    [byte-streams :as bs]
-    [primitive-math :as p]
-    [potemkin :refer [doary]])
+   [byte-streams :as bs]
+   [primitive-math :as p]
+   [potemkin :refer [doary]])
   (:import
-    [io.netty.handler.codec.base64
-     Base64]))
+   [io.netty.buffer
+    ByteBuf
+    Unpooled]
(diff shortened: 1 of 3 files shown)


ztellman:
Sorry it took me a bit to get to this.  It looks good, thanks.

kachayev:
@ztellman No worries!

jetmind:
I would love to see this released! :)

PR #347, opened 2018-01-08
HTTP Client middleware to deal with cookies, including cookies storage support

kachayev:
Effectively covers [this issue](https://github.com/ztellman/aleph/issues/188).

Implementation provides:
* mutable in-memory cookie store to maintain cookies across requests
* default cookies parsing, formatting, and matching specification (RFC6265 compliant with slightly extended domain matching procedure to cover older servers/clients)
* protocols to extend/change this behavior when necessary (i.e. when persistent storage is required or cookies matching should be changed due to proxying rules)

Heavily influenced by `CookieStore` and `CookieSpec` implementations from Apache HttpClient (used internally by `clj_http`) and Python's [cookielib](https://docs.python.org/2/library/cookielib.html).

How to use:

```clojure
(http/get "https://google.com" {:cookies [{:name "googleID" :value "42"}]})
```

Using cookie store:

```clojure
(def cookies (http.client-middleware/in-memory-cookie-store))
@(http/post "https://mysite.com/login" {:form-params {:user "me" :password "123456"}
                                        :cookie-store cookies})
;; the request below will be sent with cookies set previously
@(http/get "https://mysite.com/protected" {:cookie-store cookies})
```

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -336,7 +336,7 @@
    | `body` | an optional body, which should be coercable to a byte representation via [byte-streams](https://github.com/ztellman/byte-streams)
    | `multipart` | a vector of bodies
    | `cookie-store` | an optional instance of cookie store to maintain cookies across requests
-   | `cookie-spec` | an optional instance of a cookie management specification to enforce rules of parrsing, formatting and choosing cookies to be send with the request, defaults to the `middleware/default-cookies-spec`")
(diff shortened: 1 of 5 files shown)


kachayev:
@ztellman Reimplemented as client middleware, moved configuration settings from connection to request level, added support for explicitly set `:cookies`.

PR #350, opened 2018-01-18
Basic auth: Don't add line breaks to encoded string

ckarlsen84:
```io.netty.handler.codec.base64.Base64.encode()``` adds line breaks per default which causes basic auth for base64 encoded strings with more than 76 characters to break.

```clj
@(http/get "http://google.com" {:basic-auth (apply str (take 58 (repeat "a")))})
IllegalArgumentException only ' ' and '\t' are allowed after '\n': Basic YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFh
YQ==  io.netty.handler.codec.http.DefaultHttpHeaders$HeaderValueConverterAndValidator.validateValueChar (DefaultHttpHeaders.java:462)
```

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -437,7 +437,7 @@
                      basic-auth
                      (str (first basic-auth) ":" (second basic-auth)))
         input-bytebuf (Unpooled/wrappedBuffer (.getBytes ^String basic-auth "UTF-8"))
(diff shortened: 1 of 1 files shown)


ztellman:
Thanks for catching this, I'll cut a new alpha release.

PR #351, opened 2018-01-30
Added websocket Ping frame auto-answer

shilder:
According to RFC 6455 implementation MUST send Pong frame as response

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -41,6 +41,7 @@
      ChannelPipeline]
     [io.netty.handler.codec.http.websocketx
      CloseWebSocketFrame
+     PingWebSocketFrame
      PongWebSocketFrame
      TextWebSocketFrame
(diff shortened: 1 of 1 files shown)


ztellman:
I don't think we want to call `acquire` on the content, unless we plan to use it elsewhere.

shilder:
> I don't think we want to call acquire on the content, unless we plan to use it elsewhere.

I was thinking that way too, but omitting `netty/acquire` causes errors on `PARANOID` level because of `finally` block. I suppose netty decreases counter on `writeAndFlush` and then it again tries to decrease counter in `finally` causing `io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1` exceptions 

Maybe there is better solution to skip this finally section, but easiest way was to add `acquire`

ztellman:
Ah, I had forgotten about the finally clause, in that case it's fine.

Issue #182, opened 2015-10-27
Proxy support for http client


martinraison:
I've seen it mentioned somewhere but there didn't seem to be an issue open for it.

As a workaround for anyone looking for a solution, it might be possible to use the `pipeline-transform` argument along with `io.netty.handler.proxy.{HttpProxyHandler,Socks4ProxyHandler,Socks5ProxyHandler}`. I haven't got the time to actually try it, though.

ztellman:
`:pipeline-transform` works on the connection, so it can't be a per-request setting, it would have to be per-connection pool.  Other than that, though, that seems like a perfectly good workaround.  I haven't had to mess around with proxies much in the past, so I'd be happy to look at any code samples you may have.

martinraison:
If I get around to trying it I'll be glad to share.

neverfox:
Is there anything particular that makes per-request proxy support difficult to implement in Aleph or is it just a to-do?

ztellman:
I haven't looked at it, but I'm pretty sure it's straightforward.  I just haven't gotten to it yet.

blandinw:
This is the `:pipeline-transform` workaround mentioned above:
```
(def my-pool (aleph.http/connection-pool
               {:connection-options
                {:pipeline-transform
                 (fn [^ChannelPipeline x]
                   (doto x
                     (.addFirst (into-array ChannelHandler [(HttpProxyHandler. (InetSocketAddress. "myproxy" 8080))]))))}}))
(deref (aleph.http/get "https://github.com" {:pool my-pool}))
```

Atsman:
@blandinw, have it work for you? Because i just tried, and it doesn't work :(

kachayev:
It doesn't work for me as well. At least because of this: https://github.com/ztellman/aleph/blob/a69e066600e0e3961c4b41ac4c8b4bef8edb38d5/src/aleph/netty.clj#L723-L725 - when using http proxy you need to connect to a proxy server address, not your `remote-address`. As far as I see/understand `HttpProxyHandler` does not handle a connection flow, but you can do this passing your own handler:

```clojure
(defn add-proxy [^ChannelPipeline x]
  (doto x
     (.addFirst "proxy" (aleph.netty/channel-handler
                         :connect
                         ([_ ctx _ _ promise]
                          ;; this will connect you to the proxy first and handle req/response cycle
                          ;; as usual after that
                          (.connect ctx (InetSocketAddress. "myproxy" 8080) promise))))))

[...]

PR #352, opened 2018-01-30
Proxy support for HTTP client

kachayev:
Effectively covers #182.

How to use:

```clojure
(def pool (http/connection-pool {:connection-options {:proxy-options {:host "localhost"}}}))
(http/get "http://google.com" {:pool pool})
```

With more advanced settings:

```clojure
(def pool (http/connection-pool
           {:connection-options {:proxy-options {:host "localhost"
                                                 :port 8080
                                                 :protocol :http
                                                 :tunnel? true
                                                 :user "username"
                                                 :password "123456"
                                                 :keep-alive? true
                                                 :http-headers {"X-Via" "Proxy!"}}}}))

(http/get "http://google.com" {:pool pool})
```

Right now covers next protocols:

* HTTP proxy w/ tunneling (using CONNECT)
* HTTP proxy w/o tunneling
* Socks4
* Socks5

Proxy configuration is set as an additional (optional) key in `:connection-options` when creating connections pool. Minimal configuration is `:host`, all other settings are optional. Proxy connection timeout is suppressed in preference to `:connect-timeout` handling. "Proxy-Connection" is set to "Keep-Alive" if otherwise is not specified explicitly. 

The default protocol is HTTP w/o tunneling (same as `curl`). The client falls back to CONNECT if one the following is true:

* `tunnel?` is set to `true` explicitly 
* auth params are provided
* additional http headers are provided
* attempting to connect to HTTPS destination throw HTTP proxy

[Comment](https://github.com/ztellman/aleph/pull/352#issuecomment-361927226) on per-request configuration challenges.


kachayev:
`clj_http` client gives us an option to specify proxy settings like this:

```clojure
(client/get "http://google.com" {:proxy-host "localhost" :proxy-port "8080"})
```

There is a problem with implementing per-request configuration: we need to pass information to `create-connection` and right now this is not possible to accomplish (`flow/acquire` and `generate` functions take a single argument which is used also for `object-per-key` control). Creating another connections pool "silently" seems to be a terribly misleading choice.

What we can do: use a special "routes" objects (handle information on host and optional proxy configuration) instead of simple `URI`s to "register" connections in the pool. But it seems fairly complex and might potentially break someone's code (i.e. stats processing callbacks). Or we can simply skip over this and force users to setup a connection pool explicitly when working with proxies.

@ztellman what's your opinion?

kachayev:
@ztellman Updated PR with SSL handling.

PR #355, opened 2018-02-11
HTTP client to support async DNS resolver

kachayev:
So far, `aleph` relies on the domain name resolution mechanism provided by JDK (default in `netty`). Which brings a few well-known problems: names resolving in blocking manner (doesn't play well with i.e. crawlers), no way to employ an alternative cache/retry policy, set TTLs etc.

The solution provided gives the user ability to switch DNS resolving to async mechanism provided by `netty`. How it could be used:

```clojure
(def pool (http/connection-pool {:dns-options {:max-ttl 16 :query-timeout 200}}))
@(http/get "http://google.com" {:pool pool})
```

This code creates a shared instance of DNS resolver (`io.netty.resolver.dns.DnsAddressResolver`) which is passed to each created connection. 

In addition to that, you can now switch off names resolving at all. Might be helpful/desirable when working with [proxies](https://github.com/ztellman/aleph/pull/352). 

```clojure
(def pool (http/connection-pool {:connection-options {:name-resolver :noop}}))
@(http/get "http://127.0.0.1:3000" {:pool pool})
```

The only one question that I have is placed [here](https://github.com/ztellman/aleph/compare/master...kachayev:ft-resolver-settings?expand=1#diff-95720affa86696b46c5458a71ab3e8d7R773): when using a shared instance of the resolver we need to `.close` it when the connection pool is shutting down. To do that I need to extend `flow/instrumented-pool` with either a callback to be call after shutdown is performed or with some kind of `.terminationFuture` that I might use to add a listener. @ztellman  What do you think is more preferable?


src/aleph/http.clj
```diff
@@ -94,6 +95,7 @@
    | `stats-callback` | an optional callback which is invoked with a map of hosts onto usage statistics every ten seconds
    | `max-queue-size` | the maximum number of pending acquires from the pool that are allowed before `acquire` will start to throw a `java.util.concurrent.RejectedExecutionException`, defaults to `65536`
    | `control-period` | the interval, in milliseconds, between use of the controller to adjust the size of the pool, defaults to `60000`
+   | `dns-options` | an optional map with async DNS resolver settings, for more information check `aleph.netty/create-dns-resolver`. When set, ignores `name-resolver` setting from `connection-options` in favor of shared DNS resolver instace
```
ztellman: I think this should be just `dns-resolver`

ztellman:
Beyond the one doc fix I mentioned in the comments, this looks good.

src/aleph/http.clj
kachayev: @ztellman Do you mean to rename `dns-options` to `dns-resolver`?

kachayev: @ztellman 👍

kachayev:
@ztellman I fixed the documentation, thanks! But there's another issue that I want to discuss: https://github.com/ztellman/aleph/pull/355/files#diff-95720affa86696b46c5458a71ab3e8d7R773 - I close DNS resolver on when client-group is terminated, but, ideally, it should be closed when the appropriate connection pool is terminated. To do that, I need to be able to set a callback on a shutdown of the `flow/instrumented-pool`. Which can be done i.e. by subclassing `dirigiste.Pool`. Is there any better way?

ztellman:
Sorry, I guess I'm missing something, but I don't see how Dirigiste connection pools factor into this change.  Those are only used for HTTP connections, as far as I know.

kachayev:
@ztellman The problem that we create DNS resolver here: https://github.com/ztellman/aleph/pull/355/files#diff-f4435cedcc2c6307b45b191b2933fcbdR131, when creating connections pool. Meaning that the best moment to close the resolver is when connections pool is closed (destroyed). With the current implementation, it would be closed only on client group's termination, here: https://github.com/ztellman/aleph/pull/355/files#diff-95720affa86696b46c5458a71ab3e8d7R811

kachayev:
@ztellman Reimplemented once again using `DnsAddressResolverGroup` subclass to benefit from `InflightNameResolver` caches. Netty interfaces and class hierarchy for `dns` namespace is pretty clumsy. I will probably open a pull request to extend `DnsAddressResolverGroup` implementation to be more flexible, but for now, we can move on with this.

Please, disregard my comment about `.close`. As we use `AddressResolverGroup` it will close appropriate resolver when the executor is [terminated](https://github.com/netty/netty/blob/06dcca1dbcc048b6ce7d954e4ecdf1d3ed6af034/resolver/src/main/java/io/netty/resolver/AddressResolverGroup.java#L81).

PR #360, opened 2018-02-24
Websocket connection to support permessage-deflate compression

kachayev:
`http/websocket-client` and `http/websocket-connection` now except optional argument `:compression?` to enable [permessage-deflate](https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-28) extension. 

I also added support for an optional `pipeline-transform` as it's done for HTTP connections. The problem is that It was not even possible to activate extension "manually" as a websocket connect didn't provide any flexibility with channel pipelines. 

/cc https://github.com/ztellman/aleph/issues/304

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -164,7 +164,8 @@
    | `sub-protocols` | a string with a comma seperated list of supported sub-protocols.
    | `headers` | the headers that should be included in the handshake
    | `max-frame-payload` | maximum allowable frame payload length, in bytes, defaults to 65536.
-   | `max-frame-size` | maximum aggregate message size, in bytes, defaults to 1048576."
+   | `max-frame-size` | maximum aggregate message size, in bytes, defaults to 1048576.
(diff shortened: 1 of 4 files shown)


kachayev:
🤔 [this build](https://github.com/ztellman/aleph/pull/360/commits/38577efc7a83b06130b74d21bb735373a75c8f4f) failed on Travis CI with leak detector exception. Not sure why that happened, but I'm not able to reproduce the problem. Will try to dig deeper into that.

ztellman:
I screwed up the master merge.  I don't see anything in this change that would make this flaky (other than maybe the underlying Netty functionality, so I'm going to do a real merge and let you do some more investigation before I cut a full release with this functionality.

kachayev:
@ztellman Sure, I'll check!

ztellman:
Okay, this is actually very reliably failing in CI now, but I can't reproduce it on my machine.  I've tried targeting the latest Netty, just to see if that would help, and it doesn't.  I've commented out the tests for now.  If you're able to fix it soon, great, otherwise I'll disable the functionality for now and we can reactivate it in a future release.

PR #366, opened 2018-03-11
Proxy pending writer to handle flush properly 

kachayev:


kachayev:
There's also a problem with exceptions propagation and failing pending writes. We can implement it here from scratch, but I think there should be a better way to reuse netty's functionality. I'll try to spend more time on this.

kachayev:
@ztellman Okay, now I think this one is finished. 

[This](https://github.com/ztellman/aleph/pull/366/files#diff-e2d1dde5203e0962010d2fb176031b3cR352) issue was kinda tricky to find. It turned out that `VoidChannelPromise` might cause some problems from time to time.

kachayev:
Quick follow up here: I open [PR](https://github.com/netty/netty/pull/7780) for Netty project as well.

PR #373, opened 2018-03-18
Fix race condition on websocket handshake lead to skipping the first frame

kachayev:
The inial idea here was pretty simple: to extend websocket tests to check `raw-server?` set to `true`. Which undercovered a sneaky race condition that failed tests from time to time. Now [the build](https://travis-ci.org/ztellman/aleph/builds/355047654) is in green (I've already re-tested it dozens of times).

There's quite a lot here. A quick summary of notable changes:

1. Use [`processHandshake`](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-e2d1dde5203e0962010d2fb176031b3cR578) instead of `finishHandshake`. Hard to imagine this is the problem, but still safer.

2. Cleaning up the connection is updated both for the [server](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-649772f881cdedf671330e0bc28a5c56R504) and the [client](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-e2d1dde5203e0962010d2fb176031b3cR593). The approach is to check if a channel is not closed yet (might happen). Server connection does not perform `channel.close` (as it's already [covered](https://github.com/netty/netty/blob/2c90b6235d3c120c45ea2c52e4ada5141f3f3021/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker.java#L331) in Netty). Not sure why Netty implementation is not consistent here, will probably open Issue for this.

3. `raw-stream?` server [skips](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-649772f881cdedf671330e0bc28a5c56R385) empty buffers.

4. `raw-stream?` websocket server to [retain](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-649772f881cdedf671330e0bc28a5c56R543) `ButyBuf` from a message before putting onto the stream. 

5. Websocket connection pipeline is [reogranized](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-649772f881cdedf671330e0bc28a5c56R608) **before** handshake. Note, that `pipeline-transform` is still applied [after](https://github.com/ztellman/aleph/compare/fix-raw-stream-websockets-test?expand=1#diff-649772f881cdedf671330e0bc28a5c56R625) to make sure we do not break no one's existing transformations.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -54,8 +54,7 @@
    will be errors, and a new connection must be created."
   [^URI uri options middleware on-closed]
   (let [scheme (.getScheme uri)
-        ssl? (= "https" scheme)
-        response-executor (:response-executor options)]
+        ssl? (= "https" scheme)]
     (-> (client/http-connection
           (InetSocketAddress.
             (.getHost uri)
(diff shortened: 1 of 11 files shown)


kachayev:
Regarding connection close: it's actually [implemented](https://github.com/netty/netty/commit/cfd514d099fb41b2a467ca208fe1334bb04f8f6c#commitcomment-3500537) this way intentionally in Netty. Still, not compliant with [RFC](https://tools.ietf.org/html/rfc6455#section-5.5.1) though. RFC requires both sides to wait for the echoed closing frame which is not handled in any way right now. Fully compliant implementation is not trivial. So, I think it makes sense to leave things "as is", but I'll keep this on my radar.

ztellman:
Sorry for having ignored this PR for so long.  Is everything good to merge at this point?

kachayev:
@ztellman From what I can tell, - yes. At least the build is pretty stable now. It always hard to be sure that the underlying race condition is fixed as the absence of failures is not a strict proof. But all changes made seem to make sense.

kachayev:
🎉

PR #378, opened 2018-04-17
Split Inbound and Outbound handlers to take advantage of Netty's optimizations

kachayev:
As `ChannelPipeline` skips the evaluation of certain handlers to shorten the stack depth when possible.

There's also `pending-proxy-connection-handler` that might be created as `ChannelInboundHandler`, but I skipped it on purpose (not to mess up with another [pull request](https://github.com/ztellman/aleph/pull/377)).

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -97,7 +97,7 @@
               complete
               body)))]
 
-    (netty/channel-handler
+    (netty/channel-inbound-handler
 
       :exception-caught
       ([_ ctx ex]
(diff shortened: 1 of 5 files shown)


kachayev:
@ztellman I don't have benchmarks, I can do some tho' if it's necessary. It's just a good practice when working with Netty, keeping your pipeline as clean as possible. That gives Netty ability to apply internal optimizations, and keep things more straightforward when doing manipulations with pipeline (visitors, lifecycle observers etc).

ztellman:
No need, I agree this is the more idiomatic way to use Netty.  I just wanted to know if there were any concrete improvements I could expect from the change.

kachayev:
@ztellman 👍 I would also update `pending-proxy-connection-handler` as it's already merged.

PR #379, opened 2018-04-17
Fix cookie paths

julianwegkamp:
Fixes an issue with middleware/req->cookie-origin. Originally, requests that were previously processed by the wrap-url middleware always returned :path "/". This caused cookies not to be sent by the http client.
I also simplified the function a bit.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -615,19 +615,16 @@
         (= \/ (-> origin-path (subs (count norm-path)) first))))))
 
 (defn req->cookie-origin [{:keys [url] :as req}]
-  (if (some? url)
-    (let [{:keys [server-name server-port uri scheme]} (parse-url url)]
-      {:host server-name
-       :port server-port
-       :secure? (= :https scheme)
-       :path (cond
-               (nil? uri) "/"
-               (str/starts-with? uri "/") uri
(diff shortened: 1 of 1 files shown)


src/aleph/http/client_middleware.clj
```diff
@@ -615,19 +615,16 @@
...
-               (str/starts-with? uri "/") uri
-               :else (str "/" uri))})
-    {:host (some-> (or (:host req) (:server-name req)) IDN/toASCII)
-     :port (or (:port req) (:server-port req) -1)
-     :secure? (= :https (or (:scheme req) :http))
-     :path "/"}))
+  (let [{:keys [server-name server-port uri scheme]} (if (some? url)
+                                                       (parse-url url)
+                                                       req)]
+    {:host server-name
```
kachayev: @julianwegkamp Previous implementation actually followed [this logic](https://github.com/ztellman/aleph/blob/5e22db5f99f259d0b3d3f70d6eddcd4a59cd6588/src/aleph/http/client.clj#L68-L69) for "no URL" case, which is not handled now

julianwegkamp: Didn't see that... I reverted to the previous implementation and implemented a function to get the path

kachayev: @julianwegkamp 👍

PR #384, opened 2018-05-18
http: name-resolver is never invoked

greut:
Hi,

I was trying to use the async DNS option and hit a little wall. 

```clj
(http/connection-pool {:dns-options {:name-servers ["8.8.4.4"]}})
```

**result**

it continues to use my local setup.

screenshot from 2018-05-18 11-32-47

**fix**

in `create-connection`, `(InetSocketAddress. host port)` does the resolution and thus bypasses the name resolver from Netty.

With the following fix, I've got the expected network datagrams.

screenshot from 2018-05-18 16-08-49

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -57,7 +57,7 @@
   (let [scheme (.getScheme uri)
         ssl? (= "https" scheme)]
     (-> (client/http-connection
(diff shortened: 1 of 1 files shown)


ztellman:
Thanks for catching this.

PR #385, opened 2018-05-28
Replacing of header separator with CRLF in multipart

akosilov:
According to [RFC1521](https://www.ietf.org/rfc/rfc1521.txt) (see 7.2.1. Multipart:  The common syntax), header separator must be a CRLF char sequence, so `\newline` in multipart.clj should be replaced with `\r\n`.
Current implementation leads to a problem with e.g. [Apache Commons Multipart](https://git-wip-us.apache.org/repos/asf?p=commons-fileupload.git;a=blob;f=src/main/java/org/apache/commons/fileupload/MultipartStream.java;h=2c58e7e4138feb8dd743253ea68870f8bd7123a3;hb=refs/heads/master#l188). The entire body is interpreted as a header that results in [org.apache.commons.fileupload.MultipartStream$MalformedStreamException](https://git-wip-us.apache.org/repos/asf?p=commons-fileupload.git;a=blob;f=src/main/java/org/apache/commons/fileupload/MultipartStream.java;h=2c58e7e4138feb8dd743253ea68870f8bd7123a3;hb=refs/heads/master#l570).

### DIFF
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -60,12 +60,12 @@
 (defn part-headers [^String part-name ^String mime-type transfer-encoding name]
   (let [cd (str "Content-Disposition: form-data; name=\"" part-name "\""
              (when name (str "; filename=\"" name "\""))
-             \newline)
-        ct (str "Content-Type: " mime-type \newline)
+             "\r\n")
+        ct (str "Content-Type: " mime-type "\r\n")
         cte (if (nil? transfer-encoding)
               ""
-              (str "Content-Transfer-Encoding: " (cc/name transfer-encoding) \newline))]
-    (bs/to-byte-buffer (str cd ct cte \newline))))
+              (str "Content-Transfer-Encoding: " (cc/name transfer-encoding) "\r\n"))]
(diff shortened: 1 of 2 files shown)


danielcompton:
>  Note that the encapsulation boundary must occur at the beginning of a
   line, i.e., following a CRLF, and that the initial CRLF is considered
   to be attached to the encapsulation boundary rather than part of the
   preceding part.  **The boundary must be followed immediately either by
   another CRLF and the header fields for the next part, or by two
   CRLFs**, in which case there are no header fields for the next part
   (and it is therefore assumed to be of Content-Type text/plain).

Are both of the bolded cases tested for here? I wasn't quite sure, but thought I'd mention it just in case.

akosilov:
In the current imlpementation we always have at least Content-Disposition and Content-Type headers for each part. Thus, in our case only the first possibility exists, and two CRLFs may only mean that there are no more header fields. The second CRLF is being added [here](https://github.com/akosilov/aleph/blob/909adf1111b7b60f74c03c921aa090ffe54e20d3/src/aleph/http/multipart.clj#L68).

danielcompton:
Great, I thought that was probably being handled but wanted to mention it just in case. Thanks!

PR #386, opened 2018-06-01
Document aleph.http/request follow-redirects? usage

rborer:
Hi,

Browsing aleph source code, I've discovered that redirects accept extra parameters like `max-redirects` and `force-redirects`. Thus I propose to slightly enhance the documentation so new comers can more easily know about those extra parameters.

While at it, I've removed 2 unused namespaces and fixed a small typo :-)

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -235,7 +235,8 @@
      | `pool-timeout` | timeout in milliseconds for the pool to generate a connection
      | `connection-timeout` | timeout in milliseconds for the connection to become established
      | `request-timeout` | timeout in milliseconds for the arrival of a response over the established connection
-     | `read-timeout` | timeout in milliseconds for the response to be completed"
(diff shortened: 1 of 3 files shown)


gsnewmark:
Hi! Aleph actually allows passing client's boolean options with or without `?` as a last symbol :see_no_evil: Every read of a such variables is wrapped in `aleph.http.client-middleware/opt` (here's it for [`follow-redirects`](https://github.com/gsnewmark/aleph/blob/master/src/aleph/http/client_middleware.clj#L287)) which checks keywords both with and without `?` https://github.com/ztellman/aleph/blob/98e878946a97fd020a63cc43ee6e740c712ad188/src/aleph/http/client_middleware.clj#L138-L153

rborer:
Oh! I didn't bother checking the `opt` function actually, fool me!

Thanks, at least I've learned something 😄 .

I'll rework my PR to only improve the documentation since I believe it would help first time users discover the extra parameters available to tweak redirects.

PR #390, opened 2018-06-11
Remove Netty deprecated methods usage in aleph.http.client ns

rborer:
The following methods are deprecated:
- HttpHeaders/isTransferEncodingChunked
- HttpHeaders/setHost
- HttpHeaders/setKeepAlive
- (.getStatus HttpResponse)

I've only focused on aleph.http.client for now.


ztellman:
Thanks!

PR #392, opened 2018-06-21
Allow to set log4j2 logger

rborer:
For the sake of completeness, `set-logger!` now supports Log4J2 logger instance.


ztellman:
Thanks!

PR #393, opened 2018-06-21
[WIP] Modify client/client-handler to support HttpObjectAggregator

rborer:
Objective: limit response size to an upper bound and discard too big responses.

HttpObjectAggregator modifies the response to aggregate all HttpMessage(s) up to a maximum length (it throws a TooLongFrameException if the length exceed the limit).

When this handler is part of the pipeline, the channel receives an instance of FullHttpResponse with the whole aggregated response.

For now, this is a work in progress. It seems to work fine when doing manual testing, I'll now spend some time writing unit tests.

Is it the good approach to define a maximum allowed response size or is there a better approach available through aleph?


kachayev:
@rborer Maybe it makes sense to add an option to client configuration to setup `HttpObjectAggregator`?

rborer:
@kachayev definitely yes. For the moment it's more of a POC, there are things I have to check (for example, I'm not sure using a `buffered-source` of `buffer-capacity` may not mean that if the response is bigger than `buffer-capacity`, part of it will be skipped...).

ztellman:
Sorry for the late reply to this thread.  Actively discarding HTTP messages over a certain size seems like a pretty specialized case, I think it's reasonable to expect that people can accomplish this themselves via `pipeline-transform`.  I'm happy to hear any arguments to the contrary, though.

rborer:
Hi @ztellman , thanks for your feedback. This is still an early work in progress (I've changed the title in that direction).

My argument in favour of this integration is two-fold:

1. as `HttpObjectAggregator` is part of Netty it is already supported, so it's only a matter of wiring it into aleph. It is also probably the most efficient way of discarding too big content (think about PDF or videos).

1. with the current aleph release, whenever someone uses `HttpObjectAggregator` through `pipeline-transform`, it renders the http client unusable. Since `client-handler` receives an unexpected object (`FullHttpResponse`), it doesn't do anything and the client is stuck. I lost quite some time on this one ;-)

PR #396, opened 2018-07-02
http test: use throw-exceptions without ?

greut:
it seems that there is no `?` everywhere else


rborer:
Hi,

I was tricked by the same recently. Actually, the function which reads parameters (`aleph.http.client-middleware/opt`) look for the parameter both with & without the question mark `?`.

See https://github.com/ztellman/aleph/pull/386#issuecomment-393811755 for a deeper explanation :-)

greut:
@rborer thanks mate!

PR #398, opened 2018-07-12
Cleanups

kachayev:
Mostly cosmetic changes. There's some intersection with #390, so I tried to keep them non-overlapping.

A few questions to consider: 

1. Maybe it's better to use `ChannelInitializer` from Netty instead of manually defined `pipeline-initializer` handler? In case of using own handler here, we probably need to cover exceptions here (as at some point in time it's a single handler in the whole pipeline).

2. Do we really need [`Throwable`](https://github.com/ztellman/aleph/blob/46e3e3f70574dcbd3a0970d5a65a724e3ac2d27d/src/aleph/netty.clj#L639) here? Maybe relax to `Exception`?

3. Should [this](https://github.com/ztellman/aleph/blob/46e3e3f70574dcbd3a0970d5a65a724e3ac2d27d/src/aleph/netty.clj#L659-L710) be removed? I don't see it's used somewhere.


ztellman:
Using `Exception` rather than `Throwable` will ignore `assert` failures, among other things.  If we can't be more specific than `Exception`, I think using `Throwable` is almost always preferable.

kachayev:
1, 3. Will do! 

2. Got it. `AssertionError` is a good point, missed that.

kachayev:
@ztellman Okay, #3 is done. I will create a separate PR with `ChannelInitializer` instead of `pipeline-initializer` as it's a more of a standalone feature than just a "cleaning up" work.

Issue #256, opened 2016-06-28
clj-http-style debug options are ignored


lvh:
Examples are `:debug` and `:save-request?`, which are both very helpful for figuring out what actually went on when you made a request.

kachayev:
I think we can close this one as far as #401 was merged to the master.

PR #401, opened 2018-07-24
HTTP client & connections debugging

kachayev:
Partially covers #256.

1. `:save-request?` option for the `http/request` that saves initial request into `:aleph/request` key of the response. 

```clojure
user=> (clojure.pprint/pprint @(http/get "http://netty.io" {:save-request? true}))
{:request-time 7,
 :aleph/keep-alive? false,
 :headers
 {"x-cache" "HIT from mail.3f.com.ua",
  .... 
  "cache-control" "max-age=600"},
 :status 200,
 :connection-time 6,
 :body
 #object[java.io.ByteArrayInputStream 0x6a756c20 "java.io.ByteArrayInputStream@6a756c20"],
 :aleph/request
 {:save-request? true, :request-method :get, :url "http://netty.io"}}
```

It would be extremly useful to "save" the reference to the netty's `HttpMessage` instance that we're sending to the channel (e.g. to have `:aleph/req` and `:aleph/http-req` keys). But that's kinda tricky to do right now as this "hidden" by the requests/response stream machinery. So I'm still trying to wrap my head around the problem and find feasible solution.

2. `clj-http` has `:debug` option, which is hard to replicate. But we can achieve pretty much the same debuggability with netty's logging functionality. The key difference that it works on a connection level, not on a specific request level.

Introduced `:log-activity` option for connection pool that prepends `io.netty.handler.logging.LoggingHandler` instance into the connection pipeline to trace all events occuring. Looks like the following:

```clojure
user=> (def pool (http/connection-pool {:connection-options {:log-activity :info}}))
user=> @(http/get "http://netty.io" {:pool pool})
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329] CONNECT: netty.io/104.28.8.44:80
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329, L:/192.168.0.127:50738 - R:netty.io/104.28.8.44:80] ACTIVE
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329, L:/192.168.0.127:50738 - R:netty.io/104.28.8.44:80] WRITE: 53B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 47 45 54 20 2f 20 48 54 54 50 2f 31 2e 31 0d 0a |GET / HTTP/1.1..|
|00000010| 68 6f 73 74 3a 20 6e 65 74 74 79 2e 69 6f 0d 0a |host: netty.io..|
|00000020| 63 6f 6e 74 65 6e 74 2d 6c 65 6e 67 74 68 3a 20 |content-length: |
|00000030| 30 0d 0a 0d 0a                                  |0....           |
+--------+-------------------------------------------------+----------------+
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329, L:/192.168.0.127:50738 - R:netty.io/104.28.8.44:80] WRITE: 0B
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329, L:/192.168.0.127:50738 - R:netty.io/104.28.8.44:80] FLUSH
[aleph-netty-client-event-pool-1-2] INFO aleph-client - [id: 0x3ab83329, L:/192.168.0.127:50738 - R:netty.io/104.28.8.44:80] READ: 553B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 48 54 54 50 2f 31 2e 30 20 32 30 30 20 4f 4b 0d |HTTP/1.0 200 OK.|
....
```

Another option here would be to include (always) a special version of `LoggingHandler` that's flexible enough to log events only in case `:debug?` is set to `true` for a specific request, but that's not always informative, cause we're loosing a lot of information (when/how the connection is made etc). @ztellman what do you think?


kachayev:
Reimplemented `:save-request` feature as a client middleware. Not it's more functional:

* `:aleph/request` reflects all changes made by all middlewares 

* `:aleph/netty-request` holds a reference to `Netty`'s `DefaultHttpRequest` object that was sent to the channel

* `:aleph/request-body` key is introduces (when `:debug-body` set to `true` while sending request)

I don't really like underlying implementation with `atom`s, but that was the less intrusive way I found (not to break/rebuild requests/responses processing). Will add unit test here as well.

```clojure
(pprint @(http/get "https://github.com" {:save-request true :debug-body true}))
{:request-time 686,
 :aleph/request-body nil,
 :aleph/keep-alive? true,
 :headers
 {"server" "GitHub.com",
  "content-type" "text/html; charset=utf-8", 
  ...
  "cache-control" "no-cache"},
 :status 200,
 :aleph/netty-request
 #object[io.netty.handler.codec.http.DefaultHttpRequest 0x7495a23c "DefaultHttpRequest(decodeResult: success, version: HTTP/1.1)\nGET  HTTP/1.1\nhost: github.com\ncontent-length: 0"],
 :connection-time 357,
 :body
 #object[byte_streams.InputStream 0x72282fa6 "byte_streams.InputStream@72282fa6"],
 :aleph/request
 {:user-info nil,
  :request-url "https://github.com",
  :server-port nil,
  :uri "",
  :server-name "github.com",
  :query-string nil,
  :scheme :https,
  :request-method :get}}
```

kachayev:
@ztellman I think this one is ready for review now.

ztellman:
This looks good, but has a conflict with the multipart PR.

src/aleph/http/client.clj
```diff
@@ -345,6 +348,21 @@
...
+(defn coerce-log-level [level]
+  (if (instance? LogLevel level)
+    level
+    (let [netty-level (case level
+                        :trace LogLevel/TRACE
+                        :debug LogLevel/DEBUG
+                        :info LogLevel/INFO
+                        :warn LogLevel/WARN
+                        :error LogLevel/ERROR
+                        nil)]
```
rborer: Just as a side note, you could thrown the exception in the default case block to avoid a local variable. Something like this:

```clojure
(defn coerce-log-level [level]
  (if (instance? LogLevel level)
    level
    (case level
      :trace LogLevel/TRACE
      :debug LogLevel/DEBUG
      :info LogLevel/INFO
      :warn LogLevel/WARN
      :error LogLevel/ERROR
      (throw (IllegalArgumentException.
               (str "unknown log level given: " level))))))
```

It also makes the code slightly shorter :-)

src/aleph/http/client.clj
```diff
@@ -367,7 +386,11 @@
   (fn [^ChannelPipeline pipeline]
     (let [handler (if raw-stream?
                     (raw-client-handler response-stream response-buffer-size)
-                    (client-handler response-stream response-buffer-size))]
+                    (client-handler response-stream response-buffer-size))
+          logger (when (some? log-activity)
+                   (LoggingHandler.
+                    "aleph-client"
+                    ^LogLevel (coerce-log-level log-activity)))]
```
rborer: Oh, and to avoid the type hint here it could be included directly in the `coerce-log-level` function signature instead:

```clojure
(defn ^LogLevel coerce-log-level [level]
...
)
```

I know, I'm usually picky about such little details ;-)

PR #403, opened 2018-07-26
Support custom date header

michaelcameron:
Making a similar change that supported a custom server header to a support a custom date header. We use that in test environments to mimic running servers at different times without actually changing the system clock.


ztellman:
Thanks!

PR #410, opened 2018-08-06
Replace HTTP request multipart encoder with Netty's implementation

kachayev:
Pros:

* heavily tested encoder
* support for chunked input without a necessity to read everything into memory right away
* easily might be extended to support `multipart/mixed`, not only `multipart/form-data`

Cons:

* custom transfer encodings are not supported right now, only `binary` and plain text (to add `base64` or `qp` we need to extend Netty's implementation first)


ztellman:
I'm very happy to limit our multipart implementation to whatever Netty can offer out of the box.  Thanks for this.

PR #411, opened 2018-08-09
Multipart decoder

kachayev:
Implemented as a "helper" function `aleph.http.mutlipart/decode-request` to be called with `ring` request inside a handler. Works both with `ring` and raw stream handlers. Supports disk-based representation for payloads that exceed specified memory limit to ensure that we don't blow up heap processing large file uploads.


kachayev:
Weirdly enough, I could not replicate the same error locally 🤔

src/aleph/http/multipart.clj
```diff
@@ -146,3 +158,94 @@
...
+
+     ;; on each HttpContent chunk, put it into the decoder
+     ;; and resume our attempts to get the next attribute available
+     (s/connect-via
+      body
+      (fn [chunk]
+        (let [content (DefaultHttpContent. chunk)]
+          (.offer decoder content)
+          (read-attributes decoder parts)
+          ;; TODO(kachayev): I'm still not sure this is
```
ztellman: If we're feeling paranoid about this, we could just coerce it to a ByteBuffer and then back to a ByteBuf (now on-heap).  I don't think the memory copy overhead is worth worrying about.

src/aleph/http/multipart.clj
```diff
@@ -146,3 +158,94 @@
...
+     :memory? (.isInMemory attr)
+     :file? false
+     :file nil
+     :size (count content)}))
+
+(defmethod http-data->map InterfaceHttpData$HttpDataType/FileUpload
+  [^FileUpload data]
+  (let [memory? (.isInMemory data)]
+    {:part-name (.getName data)
+     :content (when memory?
```
ztellman: If it's in a file, why not just give a memory-mapped byte-buffer?

kachayev: Do you mean to use `ByteBuffer` in both cases?  We can do `InputStream` for both: right now content is converted to `InputStream` for in-memory uploads, but we can easily create an input stream directly from a temp file. I'm not sure what the most convenient way.

ztellman: Mostly I was confused why we were forcing the consumer of the API to check whether the data was in a file or not, but now that I think about it a bit more, I'm guessing it's because we don't want to allocate a file descriptor without them asking for it.  Is that right?

src/aleph/http/multipart.clj
kachayev: Nope, at all. I'm going just remove this comment as it was done initially. I've checked the implementation on the netty side and this sequence of operations seems pretty reasonable. More than that, I do not expect a lot of changes in those decoders in any foreseeable future.

Issue #346, opened 2017-12-21
websocket-client can hang indefinitely


AlexHill:
Hello,

I have a number of client devices connecting as websocket clients to a server. To handle network interruptions, I reconnect to the WS server after each disconnection, with some backoff, etc.

Some of these devices are on unreliable networks, and I'm sometimes seeing `websocket-client` fail to time out and simply hang indefinitely. I think what's happening is that the actual socket connection is succeeding, but then the client is losing connectivity during the handshake. Without a keepalive on the TCP connection, or a timeout handler on the Netty channel, the deferred never succeeds or fails.

I'm using `deferred/timeout!` to work around this but I'm not sure what's happening under the hood - i.e. whether or this means that sockets/resources are still allocated and hanging around forever in the background.

Adding an `IdleStateHandler` to the Netty channel while the handshake is in progress might fix this.

Alex

tendant:
Observed the same issue. Thanks @AlexHill for pointing out the use of manifold.deferred/timeout!.

shilder:
Related to #420

Don't know about timeouts, but handshake exceptions were ignored and left opened socket

matthewdowney:
I'm working with a similar situation (lots of unreliable connections, restarts with some backoff, etc.). I'm wondering what the idiomatic way to enforce a timeout would be?

Like, if I open the connection with something like the following, before setting any handlers, am I still creating a situation where the `aleph-netty-client-event-pool-<N>-<N>` threads could end up permanently in a `WAITING` state?

```clojure
(defn try-connect [uri timeout-ms]
  (let [pending-sock (http/websocket-client uri)]
    (-> pending-sock
        (d/timeout! timeout-ms ::timeout)
        (d/chain
          (fn [open-sock-or-timeout]
            (if (= open-sock-or-timeout ::timeout)
              (do
                ;; Make sure that if/when the socket does eventually open, it
                ;; gets closed
                (d/chain pending-sock (fn [open-sock] (s/close! open-sock)))
                nil)
              open-sock-or-timeout))))))
```

Then I'd do something along the lines of this to consume the connection:

```clojure
(d/let-flow [sock (try-connect "wss://foo.bar/" 10000)]
  (if sock 
    (do
      (println "Connection open.")
      (d/chain 
        (s/consume (fn [msg] (println ">" msg)) sock)
        (fn [_] (println "Connection closed."))))
    (println "Connection timed out, retrying...")))
```

Is there a better way to do this?

[...]

PR #422, opened 2018-10-14
Handle WebSocket handshake timeouts

shilder:
Added new option handshake-timeout for websocket client

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -60,12 +60,12 @@
      LoggingHandler
      LogLevel]
     [java.util.concurrent
-     Future
-     TimeoutException]
+     Future]
     [java.util.concurrent.atomic
      AtomicInteger]
     [aleph.utils
-     ProxyConnectionTimeoutException]))
+     ProxyConnectionTimeoutException
+     WebSocketHandshakeTimeoutException]))
 
 (set! *unchecked-math* true)
 
@@ -621,7 +621,7 @@
                         handshake-timeout
                         (fn []
                           ;; if there was no answer after handshake was sent - close channel
-                          (d/error! d (TimeoutException. "WebSocket handshake timeout"))
+                          (d/error! d (WebSocketHandshakeTimeoutException. "WebSocket handshake timeout"))
                           (netty/close ctx))))
 
            (.handshake handshaker ch))
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -671,7 +671,7 @@
 
           (let [handshake-future (.handshake handshaker ch)]
             ;; start handshake timeout timer if necessary only
-            ;; when handshake is written and flushedx
+            ;; when handshake is written and flushed
             (when (pos? handshake-timeout)
               (-> (netty/wrap-future handshake-future)
                   (d/chain'
@@ -684,7 +684,7 @@
                                       (when (not (.isHandshakeComplete handshaker))
                                         (d/error! d (WebSocketHandshakeTimeoutException.))
                                         (netty/close ctx)))
-                                    ^long (long handshake-timeout)
+                                    (long handshake-timeout)
                                     TimeUnit/MILLISECONDS)]
                        (reset! timeout-task timeout))))))))
         (.fireChannelActive ctx))
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -66,6 +66,9 @@
     [io.netty.handler.logging
      LoggingHandler
      LogLevel]
+    [io.netty.util.concurrent
+     EventExecutor
+     ScheduledFuture]
     [java.util.concurrent
      Future
      ConcurrentLinkedQueue]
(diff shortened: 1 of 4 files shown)


shilder:
Related to #346 More like request for discussion. Some opened questions

1. Is it ok to use somewhat private method `(.in time/*clock*)` to schedule timeout handler execution ?
2. There are already couple of specific Exception classes for different timeouts (like `ReadTimeoutException` and so on). Should i add another class ? Something like `WebSocketHandshakeTimeoutException` ?

kachayev:
@shilder Regarding `WebSocketHandshakeTimeoutException`: 👍. It's better to keep all timeouts separate when possible, so the user can distinguish them somehow later. Subclassing works just fine for that.

kachayev:
@shilder Regarding `WebSocketHandshakeTimeoutException`: 👍. It's better to keep all timeouts separate when possible, so the user can distinguish them somehow later. Subclassing works just fine for that.

kachayev:
@shilder Regarding `WebSocketHandshakeTimeoutException`: 👍. It's better to keep all timeouts separate when possible, so the user can distinguish them somehow later. Subclassing works just fine for that.

kachayev:
@shilder Regarding `WebSocketHandshakeTimeoutException`: 👍. It's better to keep all timeouts separate when possible, so the user can distinguish them somehow later. Subclassing works just fine for that.

src/aleph/http/client.clj
```diff
@@ -609,6 +614,16 @@
        ([_ ctx]
          (let [ch (.channel ctx)]
            (reset! in (netty/buffered-source ch (constantly 1) 16))
+
+           ;; Start handshake timeout timer
+           (reset! timeout-task
```
ztellman: I was trying to figure out why you used the `.in` accessor here, and realized that `manifold.time/in` doesn't allow it to be cancelled.  That's my bad, I'm going to merge these changes in and then fix the underlying method so we don't have to reach under the covers.

ztellman:
@kachayev it seems this no longer merges after I brought in the ping functionality, since you have more familiarity with that would you mind getting it into order?

src/aleph/http/client.clj
```diff
@@ -609,6 +614,16 @@
...
+                   (.in time/*clock*
+                        handshake-timeout
+                        (fn []
+                          ;; if there was no answer after handshake was sent - close channel
+                          (d/error! d (WebSocketHandshakeTimeoutException. "WebSocket handshake timeout"))
+                          (netty/close ctx))))
```
kachayev: We're not on I/O thread here, so we should probably close the channel, not the context. Right? @ztellman

src/aleph/http/client.clj
```diff
@@ -609,6 +614,16 @@
        ([_ ctx]
          (let [ch (.channel ctx)]
            (reset! in (netty/buffered-source ch (constantly 1) 16))
+
+           ;; Start handshake timeout timer
+           (reset! timeout-task
+                   (.in time/*clock*
```
kachayev: One another thing... Maybe it's better to use EventLoop executor to schedule the task (`ctx.executor().schedule`) instead of manifold's clock? We can also just trigger user event from the runnable and check if the handshake was complete in `:user-triggered`. It seems like this approach would leave fewer spaces for race conditions. What do you think? @shilder  @ztellman 

I still wonder why this functionality is not implemented in Netty, e.g. `SslHandler` supports handshake timeout as a parameter. I'm going to open an issue to start a discussion there.

shilder: >One another thing... Maybe it's better to use EventLoop executor to schedule the task (ctx.executor().schedule) instead of manifold's clock

Yes, we can use `EventExecutor` and Netty is using it internally in `IdleStateHandler` to schedule timeout handlers.

And about `:user-triggered` - yes, it will work, but I'm not sure how it will help with race conditions ? AFAIU `Future#cancel` provides synchronization point. Or maybe you are talking about situation where context is closed by timeout handler while another thread awaits on `Future#cancel` ? I'm not a Netty expert, so I'm not sure how it handles serialization of events for context

kachayev: Not only for IdleTimeoutHandler but for all timeout tasks. Future canceling is what I want to avoid here specifically because of a synchronization: it might hurt performance when you introduce sync primitives to the pipeline running on I/O thread. 

If you don’t mind, I can update your PR.

kachayev: On it!

kachayev:
The more I work with this functionality, the more I wonder why don't we use Netty's `WebSocket*ProtocolHandler` to deal with handshakes? It doesn't support timeouts out of the box (at least for now). Apart from that, it's a well-tested implementation that we can leverage injecting our own handler & realizing deferred with newly created `netty/sink` on `HANDSHAKE_COMPLETE` user event triggered by Netty's handler. @ztellman WDYT? Any objections?

kachayev:
@shilder I've reimplemented timeouts, please double check on your side. Notable changes: 

1. Avoid `.cancel` sync on I/O threads 
2. Schedule timeout on I/O threads
3. Schedule timeout only when the handshake was effectively flushed, not before
4. Cancel timeout when processing is done, avoid 2 timeouts

I've also added a test case to check that the solution works. Should we do the same for a server handler?

src/aleph/http/client.clj
```diff
@@ -692,33 +701,35 @@
...
-                          ;; handle handshake exception
-                          (d/error! d ex)
-                          (netty/close ctx)))))
+              (-> (.processHandshake handshaker ch msg)
+                  netty/wrap-future
+                  (d/chain'
+                   (fn [_]
+                     ;; cancel timeout if it was scheduled
+                     (when-let [^ScheduledFuture timeout-future @timeout-task]
+                       (.cancel timeout-future false)
```
shilder: Isn't there a race condition ? Client can get handshake completed and timeout event can get triggered on different thread ?

> Avoid `.cancel` sync on I/O threads

Well, that is `.cancel` call again. I don't think it will cause any performance issues though - it is single CAS call to mark future as cancelled

kachayev: It's a future listener, it executes not on I/O thread

kachayev: We rely on the fact that `(.isHandshakeComplete handshaker)` inside timeout callback returns `false` if  `handshakeFuture` realized.

kachayev:
From what I can see, there's no necessary to introduce the same for server's handshake, as it's just about flushing response to the client. 

@ztellman Now ready for review/merge.

kachayev:
@ztellman Merged with the latest changes from master, so now it's mergeable. 

@shilder I've also moved timeout cancelation to `d/finally'` instead of `d/chain'`. Does it make sense?

ztellman:
This has conflicts with the other WS changes I merged, but will merge once those are resolved.

kachayev:
@ztellman Done!

kachayev:
@ztellman Fixed conflicts here.

kachayev:
I'm going to rebase to 1.0.0 to merge with latest changes there. Substantial portion of the code is intertwined with websockets pipeline changes we've made so far.

kachayev:

PR #423, opened 2018-10-26
Add :ssl? option to HTTP start-server.

scramjet:
Hi Zach, I've just had a frustrating few hours trying to add SNI support to an Aleph-based HTTP server using `io.netty.handler.ssl SniHandler`. I was adding a `SslContext` via Aleph's `:ssl-context` option so that Aleph would configure correctly for SSL, then removing it in the pipeline-transform and adding the SNI handler.

Unfortunately if you do this the removed `SSLHandler` times out on the SSL handshake 10 seconds later and silently closes the connection! I'd say this is a bug in `SslHandler`, but it's still sub-optimal to be creating and then tossing `SslHandler`'s this way. So I thought a simple way to enable this was to allow clients to specify the `:ssl?` option directly, and then take responsibility for actually setting up the context themselves.

It'd be great if you think this is an good solution, but I'm also open to other ways to do this.

Cheers,

Matt.

PS. and thanks for Manifold and Aleph! They're amazingly clever and useful pieces of kit.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,7 +16,7 @@
     [byte-streams "0.2.5-alpha2"]
     [potemkin "0.4.5"]])
 
-(defproject aleph "0.4.7-alpha2"
+(defproject aleph "0.4.7-alpha2-mpp"
   :description "a framework for asynchronous communication"
   :repositories {"jboss" "https://repository.jboss.org/nexus/content/groups/public/"
                  "sonatype-oss-public" "https://oss.sonatype.org/content/groups/public/"}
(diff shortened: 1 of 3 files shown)


scramjet:
`:manual-ssl?` sounds fine to me. Conveys the meaning. I agree, in hindsight `:ssl?` isn't great.

ztellman:
Thanks!

scramjet:
No worries! Happy to be able to contribute something.

Issue #151, opened 2015-02-15
HTTP Client Example with chain does not work


darren-fhf:
Using aleph-0.4.0-beta2, Oracle Java 1.8.0_25

It could just be that I'm misunderstanding something here, but if I put the bs/to-string inside the chain, as it is in the README.md, it hangs forever.

````
(ns wr.play
  (:require [aleph.http        :as http]
            [manifold.deferred :as d]
            [byte-streams      :as bs]))

;;;; This works

(defn ex1 [url]
  (d/chain (http/get url) :body))

(-> @(ex1 "http://www.google.com") bs/to-string)

;;;; This hangs forever

(defn ex2 [url]
  (d/chain (http/get url) :body bs/to-string))

@(ex2 "http://www.google.com")
````

ztellman:
Are you using a specific version of Manifold, or just the transitive
dependency from Aleph?

darren-fhf:
Just the transitive dependency

ztellman:
Oh, I see the problem.  This is a duplicate of issue #137, which has been fixed but hasn't been put into a formal release.  I thought maybe I hadn't completely fixed this, but I expect if you target `0.4.0-SNAPSHOT` everything will work as expected.  I'm closing this issue, but will release `0.4.0-beta3` shortly.

darren-fhf:
Verified working with 0.4.0-SNAPSHOT. Looking forward to the -beta3 build.

ztellman:
beta3 is released

PR #427, opened 2018-11-17
Schedule only the first callback on executor in aleph.http/request

alexander-yakushev:
I'm not completely sure if this is a bug or an intended behavior, but submitting a PR just in case. *Note: see the diff with [?w=1](https://github.com/ztellman/aleph/pull/427/files?w=1).*

The current implementation of `aleph.http/request` uses `with-executor` to pin the response callback to the provided executor or `default-response-executor`. This causes multiple callbacks being rescheduled on the executor as a result of a single request. My expectation is that only the first callback will be scheduled on the executor, and the rest of the chain will happen in-place on the same thread.

Minimal reproducible example:

```clojure
user=> (require '[aleph.http :as http] '[manifold.deferred :as md])
nil
user=> (-> (http/get "http://example.com" 
                     {:response-executor (reify java.util.concurrent.Executor
                                           (execute [this task]
                                             (println "Execute:" task)
                                             (.execute http/default-response-executor task)))})
           (md/chain println))

Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x3f3d840a manifold.deferred.Deferred$fn__7249$fn__7250@3f3d840a]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x72d6304a manifold.deferred.Deferred$fn__7249$fn__7250@72d6304a]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x43ad2e11 manifold.deferred.Deferred$fn__7249$fn__7250@43ad2e11]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x76e9bac8 manifold.deferred.Deferred$fn__7249$fn__7250@76e9bac8]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x4d6ac814 manifold.deferred.Deferred$fn__7249$fn__7250@4d6ac814]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x33eb3a71 manifold.deferred.Deferred$fn__7249$fn__7250@33eb3a71]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x49554561 manifold.deferred.Deferred$fn__7249$fn__7250@49554561]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x7eec2265 manifold.deferred.Deferred$fn__7249$fn__7250@7eec2265]
Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x4d37d0b7 manifold.deferred.Deferred$fn__7249$fn__7250@4d37d0b7]

{:request-time 160, :aleph/keep-alive? true, ...}
```

I've traced it to this line: https://github.com/ztellman/aleph/blob/cf9540a741bc70072885a51a52aa3807059bcab6/src/aleph/http.clj#L257 . Since `with-executor` sets an executor for every deferred created in its lexical scope, every `chain` and `catch` inside it gets rescheduled onto the executor, triggering more context switches and strains the executor's queue more.

My suggested fix replaces `with-executor` with calling `onto` on the response deferred. This is conceptually the same as doing this:

```clojure
user=> (-> (http/get "http://example.com" {:response-executor nil})
           (md/onto (reify java.util.concurrent.Executor
                      (execute [this task]
                        (println "Execute:" task)
                        (.execute http/default-response-executor task))))
           (md/chain println))

Execute: #object[manifold.deferred.Deferred$fn__7249$fn__7250 0x702bd3c8 manifold.deferred.Deferred$fn__7249$fn__7250@702bd3c8]

{:request-time 128, :aleph/keep-alive? true, ...}
```

This works for me so far, but I wonder what I can be missing.

Yeah, and one more thing. One test is failing with my changes, the test that was added with https://github.com/ztellman/aleph/issues/249. With my limited understanding I can't tell whether it's testing some necessary behavior or just checks that `with-executor` is in place.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -254,92 +254,92 @@
            connection-timeout 6e4} ;; 60 seconds
       :as req}]
 
-    (executor/with-executor response-executor
-      ((middleware
-         (fn [req]
-           (let [k (client/req->domain req)
-                 start (System/currentTimeMillis)]
+    ((middleware
+      (fn [req]
+        (let [k (client/req->domain req)
+              start (System/currentTimeMillis)]
 
-             ;; acquire a connection
-             (-> (flow/acquire pool k)
-               (maybe-timeout! pool-timeout)
+          ;; acquire a connection
+          (-> (flow/acquire pool k)
+              (maybe-timeout! pool-timeout)
 
-               ;; pool timeout triggered
-               (d/catch' TimeoutException
-                 (fn [^Throwable e]
-                   (d/error-deferred (PoolTimeoutException. e))))
+              ;; pool timeout triggered
+              (d/catch' TimeoutException
+                  (fn [^Throwable e]
+                    (d/error-deferred (PoolTimeoutException. e))))
 
-               (d/chain'
-                 (fn [conn]
+              (d/chain'
+               (fn [conn]
 
-                   ;; get the wrapper for the connection, which may or may not be realized yet
-                   (-> (first conn)
+                 ;; get the wrapper for the connection, which may or may not be realized yet
+                 (-> (first conn)
 
                      (maybe-timeout! connection-timeout)
 
                      ;; connection timeout triggered, dispose of the connetion
                      (d/catch' TimeoutException
-                       (fn [^Throwable e]
-                         (flow/dispose pool k conn)
-                         (d/error-deferred (ConnectionTimeoutException. e))))
+                         (fn [^Throwable e]
+                           (flow/dispose pool k conn)
+                           (d/error-deferred (ConnectionTimeoutException. e))))
 
                      ;; connection failed, bail out
                      (d/catch'
-                       (fn [e]
-                         (flow/dispose pool k conn)
-                         (d/error-deferred e)))
+                         (fn [e]
+                           (flow/dispose pool k conn)
+                           (d/error-deferred e)))
 
                      ;; actually make the request now
                      (d/chain'
 
-                       (fn [conn']
+                      (fn [conn']
 
-                         (when-not (nil? conn')
-                           (let [end (System/currentTimeMillis)]
-                             (-> (conn' req)
-                               (maybe-timeout! request-timeout)
+                        (when-not (nil? conn')
+                          (let [end (System/currentTimeMillis)]
+                            (-> (conn' req)
+                                (d/onto response-executor)
+                                (maybe-timeout! request-timeout)
 
-                               ;; request timeout triggered, dispose of the connection
-                               (d/catch' TimeoutException
-                                 (fn [^Throwable e]
-                                   (flow/dispose pool k conn)
-                                   (d/error-deferred (RequestTimeoutException. e))))
+                                ;; request timeout triggered, dispose of the connection
+                                (d/catch' TimeoutException
+                                    (fn [^Throwable e]
+                                      (flow/dispose pool k conn)
+                                      (d/error-deferred (RequestTimeoutException. e))))
 
-                               ;; request failed, dispose of the connection
-                               (d/catch'
-                                 (fn [e]
-                                   (flow/dispose pool k conn)
-                                   (d/error-deferred e)))
+                                ;; request failed, dispose of the connection
+                                (d/catch'
+                                    (fn [e]
+                                      (flow/dispose pool k conn)
+                                      (d/error-deferred e)))
 
-                               ;; clean up the response
-                               (d/chain'
+                                ;; clean up the response
+                                (d/chain'
                                  (fn [rsp]
 
                                    ;; only release the connection back once the response is complete
                                    (-> (:aleph/complete rsp)
-                                     (maybe-timeout! read-timeout)
+                                       (maybe-timeout! read-timeout)
 
-                                     (d/catch' TimeoutException
-                                       (fn [^Throwable e]
-                                         (flow/dispose pool k conn)
-                                         (d/error-deferred (ReadTimeoutException. e))))
+                                       (d/catch' TimeoutException
+                                           (fn [^Throwable e]
+                                             (flow/dispose pool k conn)
+                                             (d/error-deferred (ReadTimeoutException. e))))
 
-                                     (d/chain'
-                                       (fn [early?]
-                                         (if (or early?
-                                               (not (:aleph/keep-alive? rsp))
-                                               (<= 400 (:status rsp)))
-                                           (flow/dispose pool k conn)
-                                           (flow/release pool k conn)))))
+                                       (d/chain'
+                                        (fn [early?]
+                                          (if (or early?
+                                                  (not (:aleph/keep-alive? rsp))
+                                                  (<= 400 (:status rsp)))
+                                            (flow/dispose pool k conn)
+                                            (flow/release pool k conn)))))
                                    (-> rsp
-                                     (dissoc :aleph/complete)
-                                     (assoc :connection-time (- end start)))))))))
+                                       (dissoc :aleph/complete)
+                                       (assoc :connection-time (- end start)))))))))
 
-                       (fn [rsp]
-                         (->> rsp
-                           (middleware/handle-cookies req)
-                           (middleware/handle-redirects request req)))))))))))
-        req))))
+                      (fn [rsp]
+                        (->> rsp
+                             (middleware/handle-cookies req)
+                             (middleware/handle-redirects request req)))))))))))
+     req)))
 
 (defn- req
   ([method url]


kachayev:
My opinion on that might be biased by the fact I have been using Aleph and Manifold libraries for many years already, but as an advanced user of those I know that to schedule chained callbacks on a specific executor you use `onto` for **any** deferred. `http/get` returns you just the same deferred and let you manage execution path in a way that you used to manage it. I have no assumptions that something different happend here.
 
@ztellman Even if we decide to change the semantic of `:response-executor` I would strongly encourage to mark this change as **BREAKING** one. From my own experience: we've spent a lot of time experimenting/tweaking with executors and `onto` scheduling, silent changes of the underlying semantic would obviously break those in the most inconvenient way: most probably our test suites will stay in green, but the performance of the system will downgrade with no visible reasons.

More than that, the desired behavior of the system is doable even now with no changes to Aleph just by tweaking `:response-executor` params and applying `onto`. So... if we think the problem described here is a real issue, we can try find more **friendly** way to handle it:

* better documentation for Aleph/Manifold
* another option that will suppress `:response-executor` and apply `onto` before returning deferred to the user

alexander-yakushev:
We've talked with @kachayev about this on another channel trying to reach a consensus, and I want to clarify here as well that I don't think this patch changes the semantics of `:response-executor` at all. The superfluous reschedules that happen now are not semantically different from being scheduled on the executor just once, because those reschedules are done to _the same executor_. In other words, even if one of the chain callbacks inside `request` function happen to hog a thread, it would still be a `:response-executor`'s thread that is hogged, no matter whether each callback is rescheduled or is executed linearly.

Now, if I'm wrong about this, and it's indeed a breaking change, I'd rather not merge it all.

alexander-yakushev:
No, we are not debating that (at least I am not), and the patch does not change that fact. Any callback chains connected to response deferred are still happening on the `response-executor`. The patch only changes the way how the callbacks inside the `aleph.http/request` function itself are scheduled onto the executor.

Here's my another attempt to show the problem with how `with-executor` interacts with chains:

```clojure
(require '[manifold.deferred :as md] '[manifold.executor :as mex])
; => nil

;; Defining a tracing executor that reports each task scheduled onto it.
(def tracing-exec (let [ex (mex/fixed-thread-executor 4)]
                    (reify java.util.concurrent.Executor
                      (execute [this task]
                        (println "Executing:" task)
                        (.execute ex task)))))
; => #'user/tracing-exec

;; This is what happens when no executor is used. The executor doesn't print
;; anything, obviously.
(-> (md/future (Thread/sleep 1000) 0)
    (md/chain inc)
    (md/chain inc)
    (md/chain inc)
    (md/chain println))
; => 3

;; Now the same but with md/onto. The executor prints once because only the
;; initial deferred is scheduled onto it. When it happens, the executor commits
;; one of its threads to handle all the chain callbacks through the end.
(-> (md/future (Thread/sleep 1000) 0)
    (md/onto tracing-exec)
    (md/chain inc)
    (md/chain inc)
    (md/chain inc)
    (md/chain println))
; Executing: #object[manifold.deferred.Deferred$fn__7243$fn__7244 0x43fa4a9a manifold.deferred.Deferred$fn__7243$fn__7244@43fa4a9a]
; => 3

;; But when with-executor is used, observe how each chained `inc` is scheduled
;; needlessly on the executor again.
(mex/with-executor tracing-exec
  (-> (md/future (Thread/sleep 1000) 0)
      (md/chain inc)
      (md/chain inc)
      (md/chain inc)
      (md/chain println)))

; Executing: #object[manifold.deferred.Deferred$fn__7243$fn__7244 0x11d701ca manifold.deferred.Deferred$fn__7243$fn__7244@11d701ca]
; Executing: #object[manifold.deferred.Deferred$fn__7243$fn__7244 0x27af4549 manifold.deferred.Deferred$fn__7243$fn__7244@27af4549]
; Executing: #object[manifold.deferred.Deferred$fn__7243$fn__7244 0x2968afcc manifold.deferred.Deferred$fn__7243$fn__7244@2968afcc]
; Executing: #object[manifold.deferred.Deferred$fn__7243$fn__7244 0x2d572e39 manifold.deferred.Deferred$fn__7243$fn__7244@2d572e39]
; => 3
```

ztellman:
I see.  The problem is this:

```clj
(-> (md/future (Thread/sleep 1000) 0)
    (md/onto tracing-exec)
    (md/chain inc)
    (md/chain inc)
    (md/chain inc)
    (md/chain println) 
    .executor)
; 3
; => nil
```

These will execute on the thread pool, but the deferred that we return will not enforce that for downstream computation (which was the entire point of the response-executor).  Your point seems to be that we're needlessly re-enqueueing onto the response-executor when we're already there, and can just do the work.  This is a very reasonable critique, but I think that's best fixed by changing the implementation of deferreds.

ztellman:
Yeah, back when I first wrote this I was afraid that if I allowed a `chain` over a deferred to inherit that executor, it would be very hard to ever escape.  The downside is, as @kachayev said, that you have to use `onto` a lot, but I feel like it's better to be explicit about these things.  The `with-executor` was a way to avoid that syntactic overhead, but as you point out it introduces a runtime overhead.  I think that's fixable, though.  Let me look into a bit more.

alexander-yakushev:
Thanks for the extra context and for looking into this. I suppose, if you compare Thread.currentThread().getThreadGroup() to the executor's ThreadGroup from its ThreadFactory, you can avoid redundant rescheduling in Manifold itself, and thus keep `with-executor` as is.

alexander-yakushev:
Another idea I just had: inside `chain` and `catch` implementation, if `.executor` of the source deferred is equal to `(manifold.executor/executor)`, then set nil executor for the chained deferred.

arnaudgeiser:
I spent some time to dig into the code and I'm not sure it requires
an Aleph change.

To me, the issue (if any) is more related to `manifold` internals. 

> Another idea I just had: inside chain and catch implementation, if .executor of the source deferred is equal to (manifold.executor/executor), then set nil executor for the chained deferred.

I don't know what the impact of this but I tend to concur.
If both the `.executor` on the source deferred and `manifold.executor/executor` are the same, that's
unlikely you want to submit another task to the executor your current thread comes from.

KingMob:
@arnaudgeiser I agree it's a Manifold change, if any. 

I need to think about where to change things. It could be done in chain and catch, which is more targeted. Alternatively, it could be a matter of updating the `wait-for`, `future-with`, and `set-deferred` macros to check the current executor/thread, which would be more global. Also not sure if it's better to check the thread group or check the threadlocal that tracks the executor. Hmmm.

More generally, the executor logic should be clarified. IIUC, it should be something along the lines of: 

1. Select an executor
    a. If an executor is specified on the deferred, choose that.
    b. Otherwise, if an executor is set on the `executor` threadlocal, choose that.
2. If the deferred is not realized, schedule the callback on the executor
3. If the deferred IS realized:
    a. Execute immediately on the current thread if it belongs to the selected executor
    b. Otherwise, schedule the callback to run on the executor

The tricky part will be in testing this. As kachayev mentioned, unless something hangs, tests should behave the same, but responsiveness may get worse if we get something wrong. Unfortunately, we don't currently have any tests to verify what executor something runs on.

alexander-yakushev:
The proposed Manifold changes are too heavy-handed. For example, there are legitimate cases where you wouldn't want:

> a. Execute immediately on the current thread if it belongs to the selected executor

But at the same time, it is probably a narrow set of cases where the user cares about executor overhead, so it doesn't warrant changing the current behavior. Feel free to close the PR until a better solution arrives.

arnaudgeiser:
The effort and risk is not worth it IMO (modify Manifold or Aleph).

However, as a Manifold user, the current behavior is something I wouldn't
expect compared to this:

> 1. Select an executor
    a. If an executor is specified on the deferred, choose that.
    b. Otherwise, if an executor is set on the `executor` threadlocal, choose that.
> 2. If the deferred is not realized, schedule the callback on the executor
> 3. If the deferred IS realized:
    a. Execute immediately on the current thread if it belongs to the selected executor
    b. Otherwise, schedule the callback to run on the executor

Regarding:

> But at the same time, it is probably a narrow set of cases where the user cares about executor overhead, so it doesn't warrant changing the current behavior. Feel free to close the PR until a better solution arrives.

I will close this PR but the discussions we had around the situation will remain a good source of knowledge.
Let's see if more Manifold/Aleph users have issues with this behavior before spending more time on this.
The solution was non-obvious for Zach and Oleksii back then and I'm not sure we (at least me) have the full picture of the impact.

Feel free to reopen @KingMob.

KingMob:
@alexander-yakushev Can you say a little more about "there are legitimate cases where you wouldn't want: Execute immediately on the current thread if it belongs to the selected executor"?

Because iiuc, at least in some cases, that's _already_ what's happening, so I'm wondering if we have another issue to address. See the comment above about #151. (Attaching listeners to realized deferreds causes them to execute in the current thread immediately.)

Regardless, I'm content to leave the behavior as is and close this. I agree it's high risk/low reward.

PR #429, opened 2018-11-26
Minor fixes in client_middleware.clj

alexander-yakushev:
1. Looks like there were some leftovers in the docstring from the time when Aleph used Slingshot.
2. I took the liberty of removing a few currently unused functions. I will drop this commit if they are still needed.
3. Since by default the requests don't have `:as` key (as I understand, it has to be provided explicitly), in common case `coerce-response-body` will do nothing, which makes it unnecessary to put it on the executor in such case.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -1,7 +1,6 @@
 (ns aleph.http.client-middleware
   "This middleware is adapted from clj-http, whose license is amenable to this sort of
    copy/pastery"
-  (:refer-clojure :exclude [update])
   (:require
     [potemkin :as p]
     [clojure.string :as str]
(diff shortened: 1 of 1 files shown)


kachayev:
@alexander-yakushev I don't think Aleph ever used `slingshot`, that's just a copy-paste from https://github.com/dakrone/clj-http/blob/eee52198d7776e3e6d965fbade9296918efc8cca/src/clj_http/client.clj#L248-L250 :) Thanks for catching that! 

Regarding removed functions. I don't worry a lot about `dissoc-in` and `update` because it's very unlikely you would use them in your code. But `parse-content-type` is kinda different. It's public, so someone could use it just as `aleph.http.client-middleware/parse-content-type`. If we want to remove them from the code, I think it would be fair at least to make function as deprecated as a first step and give users some time to adjust if necessary.

Issue #428, opened 2018-11-21
occasional ref count exception in websocket :channel-read


k7d:
This started happening in aleph 0.4.5. Couldn't yet reproduce locally but happens consistently in production (with server having few hundred open WebSocket connections).

```
2018-11-21 17:19:41.397 WARN  error in websocket handler [aleph.http.server]
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release0(AbstractReferenceCountedByteBuf.java:100)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:84)
	at io.netty.buffer.DefaultByteBufHolder.release(DefaultByteBufHolder.java:117)
	at io.netty.util.ReferenceCountUtil.release(ReferenceCountUtil.java:88)
	at aleph.http.server$websocket_server_handler$reify__3907.channelRead(server.clj:544)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:310)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:284)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
[...]

PR #430, opened 2018-12-01
Carefully release ByteBuf when processing WebSocketFrame

kachayev:
Trying to catch #428. Left a few notes/comments in the code.

@ztellman Seems more reasonable now, but it obviously requires a double-check.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -534,33 +534,44 @@
 
        :channel-read
        ([_ ctx msg]
-         (try
-           (let [ch (.channel ctx)]
-             (if-not (instance? WebSocketFrame msg)
-               (.fireChannelRead ctx msg)
-               (let [^WebSocketFrame msg msg]
-                 (cond
+         (let [ch (.channel ctx)]
+           (if-not (instance? WebSocketFrame msg)
+             ;; no need to release buffer when passing to a next handler
+             (.fireChannelRead ctx msg)
+             (let [^WebSocketFrame msg msg]
+               (cond
 
-                   (instance? TextWebSocketFrame msg)
-                   (netty/put! ch in (.text ^TextWebSocketFrame msg))
+                 (instance? TextWebSocketFrame msg)
+                 (let [text (.text ^TextWebSocketFrame msg)]
+                   ;; working with text now, so we do not need
+                   ;; ByteBuf inside TextWebSocketFrame
+                   ;; note, that all *WebSocketFrame classes are
+                   ;; subclasses of DefaultByteBufHolder, meaning
+                   ;; there's no difference between releasing
+                   ;; frame & frame's content
+                   (netty/release msg)
+                   (netty/put! ch in text))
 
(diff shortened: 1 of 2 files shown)


shilder:
I'm not sure how this request changes Binary frames processing - it's essentialy the same.

Bug #428 was introduced in https://github.com/ztellman/aleph/commit/64a8819d - it adds `.fireChannelRead` without acquiring buffer, hence double free.

kachayev:
@shilder This is fixed now, earlier `release` was done as a finalizer. Right now there's no "global" try-catch for all use cases, meaning you don't need `acquire` anything when simply passing to the next handler.

shilder:
Yes, i've seen that, but you mentioned in issue

> It looks like acquire for a message and message body is a bit messed up when dealing with binary messages, which leads to the error on finally....

And i wanted to point out that binary messages are handled just fine.

Also you don't need `if-not (instance? WebSocketFrame msg)` clause - `:else` branch of cond handles this case just fine.

kachayev:
> Also you don't need if-not (instance? WebSocketFrame msg) clause - :else branch of cond handles this case just fine.

That's a habbit of checking class before doing

```[^WebSocketFrame msg msg]```.

**Comment by shilder** [comment]
> That's a habbit of checking class before doing
> 
> `[^WebSocketFrame msg msg]`.

Yes, i understand that, but every clause in `cond` has `instance?` check, so that's unnecessary in this case

**Comment by kachayev** [comment]
@shilder Get rid both of `if` branching and type hint for a message (as we have type hint per each frame individually).

**Comment by shilder** [comment]
Thanks, looks cleaner now, and yes, that `finally` block was confusing. I guess we can get rid of it in `websocket-client-handler` too (it will be more complicated though)

File: src/aleph/http/server.clj [hunk 1]
```diff
@@ -534,33 +534,41 @@
...
-                   (instance? CloseWebSocketFrame msg)
-                   (.close handshaker ch (netty/acquire msg))
-
-                   :else
-                   (.fireChannelRead ctx msg)))))
-           (finally
-             (netty/release msg)))))]))
+         (let [ch (.channel ctx)]
+           (cond
+             (instance? TextWebSocketFrame msg)
```
gsnewmark: @kachayev you could use [`(condp instance? msg ...)`](https://clojuredocs.org/clojure.core/condp) to simplify code further.

shilder: That can be a little excessive :) 

My goal was to get rid of second `fireChannelRead`, and I don't think `condp` is used anywhere in aleph, so just for consistency i would prefer to stick with `cond`

kachayev: If my memory doesn't fail me... `condp` was used for handling websocket frames on the client side. I personally found it harder to read and reason about, but that might be just my personal preference :)

ztellman:
@kachayev also broken by the ping functionality

kachayev:
@ztellman Okay... merged this one. I'll do #422 after this one is merged, I think they will have more conflicts.

Issue #431, opened 2018-12-02
Race condition when cleaning up multipart decoder temps


kachayev:
The unfortunate fact about current `HttpPostMultipartRequestDecoder` API is that calling `destroy` [cleans up](https://github.com/netty/netty/blob/d05666ae2d2068da7ee031a8bfc1ca572dbcc3f8/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L937) all temps files that were not marked properly beforehand. In the most typical use case, I would probably do something like

```clojure
(let [parts (multipart/decode-request req)
      {:keys [file]} (stream/take! parts)]
  (io/copy file proper-destination)
```

Which may or may not work out depending on when resulting stream `on-close` callbacks are effectively [executed](https://github.com/ztellman/aleph/blob/ef331d14fabb629f94bc46b96af92b0e7d9ec37f/src/aleph/http/multipart.clj#L241-L248). The only one reasonable workaround here is to let the user to call `destroy` manually... but it would create an additional burden even for the simplest cases and will also make the entire process of request decoding more fragile. Maybe I will update the implementation with additional flags for switching off resources cleanup to make in manually whenever ready... which might also be quite cumbersome for users. Still thinking about options here. Would be glad to here any ideas and thoughts.

kachayev:
Closing the issue, implementation was merged into `1.0.0` branch

PR #432, opened 2018-12-03
Attempt to provide correct API for multipart resources management

kachayev:
The problem itself was described in #431.

Here is an implementation of another approach. Each chunk of decoded request now has `:release` closure that should be called by the end user **after** chunk is used. Looks somewhat clumsy to me but at least it's fair enough in terms of handling resources allocated earlier. I tried really hard to emphasize the idea in docs and cover with tests. Not sure if it's clear enough tho'. 

Decoder automatically takes care of those chunks that we not "used" which might be a problem too. Let's say I got a request with 2 files from the client but my handler actually does `(stream/take! chunks)` to deal with 1 of them. The underlying decoder will eventually read the entire input stream and parse that to `HttpData` that will be removed from cleaning procedure right after chunk was read from the chunks stream. Potentially it's a situation of another race: either `d/chain` callback or `release` wins. But I think it's better to try to release the same resource twice instead of not trying at all. Also not sure about performance penalties here, seems like a lot of juggling with callbacks is required here (see `read-attributes` implementation) @ztellman

### DIFF
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -38,7 +38,9 @@
     InterfaceHttpData
     InterfaceHttpData$HttpDataType]))
 
-(defn boundary []
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  boundary []
   (-> (ThreadLocalRandom/current) .nextLong Long/toHexString .toLowerCase))
 
 (defn mime-type-descriptor
@@ -48,7 +50,9 @@
    (when encoding
      (str "; charset=" encoding))))
 
-(defn populate-part
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  populate-part
   "Generates a part map of the appropriate format"
   [{:keys [part-name content mime-type charset transfer-encoding name]}]
   (let [file? (instance? File content)
@@ -80,7 +84,9 @@
 ;;
 ;; Note, that you can use transfer-encoding=nil or :binary to leave data "as is".
 ;; transfer-encoding=nil omits "Content-Transfer-Encoding" header.
-(defn part-headers [^String part-name ^String mime-type transfer-encoding name]
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  part-headers [^String part-name ^String mime-type transfer-encoding name]
   (let [cd (str "Content-Disposition: form-data; name=\"" part-name "\""
              (when name (str "; filename=\"" name "\""))
              "\r\n")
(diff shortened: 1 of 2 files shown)


kachayev:
I want to think about this API once again as I personally find it a bit clumsy... Resource cleanup for multipart is a mess in Netty by itself (e.g. some thoughts [here](https://github.com/netty/netty/issues/8655)). The necessity to call `release` closure manually will possibly introduce even more problems in applications code. Maybe we need to consider callback API, so the implementation here will be responsible for taking parts, passing them to callback and releasing after?

jameskleeh:
@kachayev Other than memory uploads/attributes not being completely cleaned up, I have a working solution implemented in Micronaut. We have the requirement of taking the data produced by netty and converting it into a method signature for example `Publisher<Part> file` so the user can process the data chunk by chunk and control the flow. The work to retrieve and release the uploads per chunk is mostly done here. https://github.com/micronaut-projects/micronaut-core/blob/master/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpDataReference.java

kachayev:
@jameskleeh Thanks a lot for the reference, I'll take a look shortly!

gsnewmark:
Among the first option variants I think `Closable` is the best, particularly because it's consistent with `http/start-server`. Do you have any particular reason to use vector as the return value instead of map?

kachayev:
> because it's consistent with http/start-server

Actually, **only** because it's consistent with the server. I don't see any other good reasons. We have `with-open` macro in a standard library, but it doesn't understand async context. Ideally, we need something like Python's `async with`. 

> Do you have any particular reason to use vector as the return value instead of map?

That's how it's done internally, e.g. for [websockets server handler](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/server.clj#L613). Map would be okay too.

kachayev:
Reimplemented public API using `io.netty.util.ReferenceCounted` and relying on `netty/release` for each consumed chunk. I played a bit with the version of API that exposes a single `.close` method for the entire decoder object and found that it lacks flexibility as I cannot deallocate memory from a specific chunk (potentially freeing a lot of space as soon as possible) while all other chunks are still in processing...  So, I think that releasing them one-by-one is the best option in terms of resources usage even tho' it might be a bit confusing from the first glance (well, nothing special if you have some experience with Netty/C/C++/Rust etc, but probably not something you're doing in Clojure on a daily basis). /cc @gsnewmark @ztellman

kachayev:

We might also consider to remove deprecated function for the sake of keeping 1.0 clean. More of a style question.

PR #436, opened 2018-12-08
Avoid reflection when calling java.net.URLEncoder/encode on JDK11

kachayev:
On JDK11

```clojure
Reflection warning, aleph/http/client_middleware.clj:137:5 - call to static method encode on java.net.URLEncoder can't be resolved (argument types: unknown, unknown).
nil
```

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -131,9 +131,9 @@
     (dissoc m k)))
 
 (defn url-encode
-  ([s]
+  ([^String s]
     (url-encode s "UTF-8"))
-  ([s encoding]
+  ([^String s ^String encoding]
     (URLEncoder/encode s encoding)))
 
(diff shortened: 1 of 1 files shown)


src/aleph/http/client_middleware.clj
```diff
@@ -131,9 +131,9 @@
     (dissoc m k)))
 
 (defn url-encode
-  ([s]
+  ([^String s]
```
rborer: As far as I know this type hint is not needed, only the one at line 136 is required.

kachayev: @rborer Yeah, you're right. Just a habit to ensure consistency in case if the first clause should be reimplemented in future (to keep user-facing API stable).

rborer: @kachayev Oh, good point about keeping the API stable :+1: .

kachayev: @rborer 👍

PR #437, opened 2018-12-09
Avoid releasing non reference counted request body

kachayev:
Request body might be on of the following:

* `nil`
* `java.io.ByteArrayInputStream`
* `manifold.stream.BufferedStream`

None of that implements `io.netty.util.ReferenceCounted` interface.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -181,7 +181,6 @@
         netty/wrap-future
         (fn [_]
           (netty/release req)
-          (netty/release body)
           (-> rsp
             (d/catch' error-response)
             (d/chain'


ztellman:
I think it might make sense to make `aleph.netty/release` use the `safeRelease` method, rather than simply not releasing the body.  Thoughts?

kachayev:
@ztellman No need too... `aleph.netty/release` is implemented with `ReferenceCountUtil` that bypasses everything that's not `ReferenceCounted`. So, it absolutely safe to release something that's not "releasable".

It just we don't need even try to do it right here.

ztellman:
Oh, sorry, I misunderstood the point you were making.  I agree, no need for this call here.

PR #440, opened 2018-12-10
Server should respond with 414 status code only in case URI is too long 

kachayev:
There're few other cases when HTTP request decoder might fail, i.e. header is too long or http version is invalid. Status code `414 URI Too Long` should not be used for those. New implementation carefully chooses a status code based on the actual error. Test cases are added to cover more cases (existing test case is fixed also as it didn't test anything).

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -212,6 +212,8 @@
 (defn uri-too-long? [^Throwable cause]
   (and
    (instance? TooLongFrameException cause)
+   ;; seems like there's no other way to distinguish too long
+   ;; URI from too long header or too large HTTP entity
    (str/starts-with? (.getMessage cause) "An HTTP line is larger than")))
 
 (defn reject-invalid-request [ctx ^HttpRequest req]
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -29,6 +29,7 @@
      ChannelHandler
      ChannelPipeline]
     [io.netty.handler.stream ChunkedWriteHandler]
+    [io.netty.handler.codec TooLongFrameException]
     [io.netty.handler.codec.http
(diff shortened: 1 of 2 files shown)


kachayev:
Well... dealing with `TooLongFrameException` is still not perfect, but it seems that Netty's current HTTP object decoder does not provide any other way. 

1. We need to read the exception message to test what actually happened, other projects depending on Netty's usually do the same thing, i.e. [Vert.x](https://github.com/eclipse-vertx/vert.x/blob/c7f27f370a91b1f38686b767bb79eeb61b1f44c6/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java#L530).

2. `TooLongFrameException` caught when reading the first line might be caused by either method, URI or version being too long (or the request is malformed). In all of those cases, we still return 414 status code. So this test effectively fails

```clojure
(deftest test-too-long-http-method
  (with-handler basic-handler
    (let [client @(tcp/client {:host "localhost" :port port})
          method (apply str (repeat 1e5 "x"))
          _ @(s/put! client (str method " / HTTP/1.1\r\n\r\n"))
          response (bs/to-string @(s/take! client))]
      (is (str/starts-with? response "HTTP/1.1 400"))
      (s/close! client))))
```

As we get "HTTP/1.1 414". But let's keep in mind. The solution presented here is not perfect, but that's what we can do using Netty. Will see what's possible to improve in Netty's decoder.

kachayev:
As far as the solution proposed here is still far from perfect, I'm going to close this PR. If there's a need to handle 414 response code properly, a better idea would be to change the behavior of `HttpDecoder`.

mping:
Hello @kachayev, did you ever get around this issue? Your PR looks legit, I wonder what the aleph team thinks of this cc @slipset

PR #442, opened 2018-12-14
Async Ring handlers wrapper 

kachayev:
Discussed [here](https://github.com/ztellman/aleph/issues/439).


kachayev:
@ikitommi Fixed, thanks!

PR #466, opened 2019-01-24
Honour -Dio.netty.eventLoopThreads property when creating a server

alexander-yakushev:
Right now, Aleph uses -Dio.netty.eventLoopThreads Java property when creating an EventLoop for a client, and ignores it for the server. Not sure if this is intended, and if it is – it could use a comment about the reasons; otherwise, it makes sense to fix this for consistency.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -924,8 +924,7 @@
    on-close
    ^SocketAddress socket-address
    epoll?]
-  (let [num-cores      (.availableProcessors (Runtime/getRuntime))
-        num-threads    (* 2 num-cores)
+  (let [num-threads    (get-default-event-loop-threads)
         thread-factory (enumerating-thread-factory "aleph-netty-server-event-pool" false)
         closed?        (atom false)
 


shilder:
You may want to have different threads count for client and server eventLoops.

I've added support for two system properties: `aleph.netty.server.eventLoopThreads` and `aleph.netty.client.eventLoopThreads`  to setup threads count separately in my local patches for aleph

alexander-yakushev:
@shilder This is another approach I agree with. However, AFAIU, Netty itself uses this single property to specify thread count for event loops regardless whether that's for server or for client. It would make sense for Aleph to mimic this behavior, and then to expose options to customize thread count explicitly (say, as a parameter to `start-server`).

kachayev:
I can see the logic behind having different settings for client & server as often times I have both intermixed in the same project. But

* Aleph doesn't use system properties except those are consumed by Netty, so... I think it would be more consistent to expose this as a `netty/start-server` param

* Client-level configuration is trickier as right now all clients use shared Nio* or Epoll* group instance

ztellman:
Yes, there's no clear way to configure the global, shared client event loop, hence using system properties, but each server has its own event loop, so it makes sense for that to be configured in `start-server`.  I'm going to close this issue, but I'm happy to consider a PR that expands the `start-server` options.

Issue #457, opened 2018-12-23
Should we close connection on RejectedExecutionException?


kachayev:
When rejecting an invalid request, we close the underlying connection [here](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/server.clj#L211-L219). But when rejecting it because of `RejectedExecutionException`, we [do not](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/server.clj#L160-L168). We probably do not need to do this, when `reject-handler` is defined by the user, but in case we return 503 status code to the client by ourselves.... it might be the case that request body was not consumed and the client misbehaved ignoring our premature response. To make sure this would not break the next request decoding, we probably need to clean everything up by closing the connection. It may lead to increased load on the server when rejected clients will try to reconnect... so I'm wondering about what is the best option here. /cc @ztellman

ztellman:
We should close the connection for all 4xx and 5xx responses, I have a
local change to this effect but it broke some tests and I haven’t tracked
down why yet.

PR #471, opened 2019-01-30
Catch FileNotFoundException when sending file body

### LINKED ISSUE #459: An attempt to send non-existing file as a server response hangs up the connection
Assume the following:

```clojure
(def f (java.io.File. "this-file-does-not-exists"))
(http/start-server (fn [_] {:status 200 :body f}) {:port 2018})
```

Any request to the server never completes (client will hang up either until own timeout or forever).  We need to catch exceptions thrown by `RandomAccessFile` constructor [here](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/core.clj#L348-L360) and respond with 500 status code.

kachayev:
Fixes #459.

In general, we need to be more careful with the cleanup procedure and any time `send-message` failed for whatever **uncaught** reason close the connection. This specific just PR covers one of the obvious failures that the user might face. /cc #457.

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -39,12 +39,19 @@
     [java.io
      File
      RandomAccessFile
-     Closeable]
+     Closeable
+     FileNotFoundException
+     StringWriter
+     PrintWriter]
     [java.util.concurrent
      ConcurrentHashMap
      TimeUnit]
     [java.util.concurrent.atomic
      AtomicBoolean]))
+
+(def ^CharSequence server-name (HttpHeaders/newEntity "Server"))
+(def ^CharSequence connection-name (HttpHeaders/newEntity "Connection"))
+(def ^CharSequence date-name (HttpHeaders/newEntity "Date"))
 
 (def non-standard-keys
   (let [ks ["Content-MD5"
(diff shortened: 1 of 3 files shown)


ztellman:
In general, I dislike the practice of the framework surfacing an exception to the client.  If the application developer wants to do that, fine, but elsewhere we provide a bare 500 response and log the issue.  I think I'd like to mirror that here.

kachayev:
@ztellman Do you mean how `error-response` is implemented with stack trace printer? I just copied that from the previous implementation, but yeah... sure, that might be reworked.

kachayev:
But let's finish with #485 first, these 2 have a lot of overlapping changes.

ztellman:
Sorry if I misattributed where that code came from, but even if I was the one that originally did it, I think it's a mistake.

kachayev:
@ztellman Oh, no problem with that. I just think that in such a case it makes sense to implement new behavior in a separated PR not to mess up the history of changes. WDYT?

kachayev:
@ztellman Merged with the latest master changes (notably `http-file` functionality). I also made internal server error to look just the same for any exception that happened. WDYT?

PR #472, opened 2019-01-31
Get rid of `:max-chunk-size` both for HTTP server and client

kachayev:
A reason for doing this was mentioned [here](https://github.com/ztellman/aleph/issues/443).


kachayev:
Let's finish with #495, I will resolve these conflicts after.

kachayev:
As of now, merging into `1.0.0` branch to accumulate pre-release changes there.

Issue #444, opened 2018-12-15
HTTP client to support body decompression out of the box


kachayev:
From `clj-http` documentation:

> The client transparently accepts and decompresses the gzip and deflate content encodings.

It also provides `:decompress-body?` option to switch off `Accept-Encoding` headers and body decompression on detected `Content-Encoding`. By default, this option is set to `true`. If we want to provide the same API, we need to cover those cases as well. But once again, we cannot just start including Accept-Encoding: gzip, deflate to any request as it's definitely would be a breaking change for a lot of users. But we can do the following: introduce `:decompress-body?` option set to `false` by default, when set to `true` explicitly using `:connection-options`, - add decompressor and appropriate header. I'll make a PR if there are no objections, @ztellman.

kachayev:
The implementation has been merged already.

PR #473, opened 2019-01-31
HTTP client to support body decompression

kachayev:
Covers #444, differences from `clj-http` API:

1. `:decompress-body?` is set for the **connection**, not for the **request**

2. `:decompress-body?` defaults to `false` when `clj-http` sets it to `true`. I will probably prefer the latest, but it would be a breaking change

3. `HttpContentDecompressor` removes "Content-Encoding" header when done. A simple way to get the information of the origin content encoding (`:orig-content-encoding` with `clj-http`) is to introduce another pipeline handler that copies the value to another header "X-Origin-Content-Encoding". I don't see any use cases except debug (well, it's used in tests too). So I introduced another option to switch this ON and not waste resources when the user doesn't need it: `:save-content-encoding?`.

4. It also automatically includes "Accept-Encoding" header as "gzip, deflate" if not set by the user manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ Aleph attempts to mimic the clj-http API and capabilities fully. It supports mul
 
 * `:cache` and `:cache-config` options are not supported as for now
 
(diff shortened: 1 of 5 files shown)


ztellman:
Approved for merging, pending fixing the conflicts.

kachayev:
@ztellman Merged!

kennyjwilli:
@kachayev @ztellman I'm quite interested in this feature. Is there anything more that needs to be done here? Happy to step in if so.

rschmukler:
Just pinging on this as well! I'd love to see this land in alpeh.

kachayev:

Issue #451, opened 2018-12-17
Use volatiles instead of atoms to store intermediate state in handlers


kachayev:
We use a bunch of atoms to store intermediate state to implement both `ring-handler` and `raw-ring-handler` (e.g. [here](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/server.clj#L224-L227)). It seems like the state is encapsulated in a way that access is sequential. We also don't use any fancy features of atoms, like watchers or compare-and-swap. So, I think, volatile variables would be sufficient to cover the functionality.

alexander-yakushev:
I've never seen anything atom-related showing up in the profile, so I don't think this would give any visible improvement.

kachayev:
@alexander-yakushev Well... most probably it would not, comparing to other parts of the handler. But it's the same story as we had in #378: when you can do less work, you most certainly should.

PR #474, opened 2019-02-01
Use volatile! instead of atoms for internal states in handlers

kachayev:
Mentioned in #451. It looks like all access to the internals of handlers should be sequential. If I'm not missing anything...

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -96,8 +96,8 @@
 
 (defn raw-client-handler
   [response-stream buffer-capacity]
-  (let [stream (atom nil)
-        complete (atom nil)
+  (let [stream (volatile! nil)
+        complete (volatile! nil)
 
         handle-response
         (fn [response complete body]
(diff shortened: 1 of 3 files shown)


src/aleph/http/client.clj
```diff
@@ -147,11 +147,11 @@
...
   [response-stream ^long buffer-capacity]
-  (let [response (atom nil)
-        buffer (atom [])
+  (let [response (volatile! nil)
+        buffer (volatile! [])
         buffer-size (AtomicInteger. 0)
-        stream (atom nil)
-        complete (atom nil)
+        stream (volatile! nil)
+        complete (volatile! nil)
```
alexander-yakushev: Correct me if I'm wrong but on this line https://github.com/ztellman/aleph/blob/master/src/aleph/http.clj#L334-L348 the HTTP client relies on `complete` being an atom so that it subscribe for the event of it becoming true and releasing the connection to the pool. You can't do that with volatiles, obviously.

kachayev: Well, actually it relies on anything that's "deferrable" (and `clojure.lang.IDeref` should work fine). You're right about access from different threads... which might not hurt in this particular case as we're only reading here. But overall it's a slippery slope I don't really want to play on, at least right now. Using non-coercible chains there also feels a bit suspicious, I need to dig more. Regarding this specific PR, I would close it to choose a better time for such changes in future.

alexander-yakushev: Oops, sorry, there are name clashes in the code. That `complete` that eventually becomes `:aleph/complete` is a deferred, not an atom. Sorry for the confusion.

kachayev: Oh, no problems. I still think that it's better to bring such changes when better tests suites will be ready to go, so we can be sure that nothing is broken in a non-obvious way.

PR #477, opened 2019-02-04
Use Epoll datagram channel for DNS resolver group when running on Epoll

kachayev:
Fixes #476.

It also extends DNS options with new `:epoll?` setting.

Issue #475, opened 2019-02-01
transport-native-epoll JAR should be loaded with OS classifier to support Epoll 


kachayev:
If I use Aleph with current `project.clj`, Epoll is not available even if kernel supports it. To support Epoll we need to download `io.netty/netty-transport-native-epoll` with "linux-x86_64" classifier. See screenshots

<img width="1133" alt="screen shot 2019-02-01 at 5 23 09 pm" src="https://user-images.githubusercontent.com/485647/52132189-4d203280-2647-11e9-8067-ed55a4b3c38f.png">

Set this manually to the list of dependencies as 

```[io.netty/netty-transport-native-epoll "4.1.33.Final" :classifier "linux-x86_64"]```

<img width="1267" alt="screen shot 2019-02-01 at 5 18 12 pm" src="https://user-images.githubusercontent.com/485647/52132230-5e693f00-2647-11e9-9336-9b39fa9adbe1.png">

Not sure what is the best way to make it less painful for the user. At least we should document this explicitly.

**Comment by alexander-yakushev** [comment]
Wow, **real** epoll is nice.
<img width="1251" alt="screen shot 2019-02-04 at 23 51 02" src="https://user-images.githubusercontent.com/468477/52239715-cfbf2100-28d7-11e9-8a4d-5ffb7ca8cc41.png">

13% usertime reduction with just one line. Thank you for discovering this!

**Comment by kachayev** [comment]
Closing the issue, changes were merged into `1.0.0` branch

**Comment by KaliszAd** [comment]
The verionsion of `netty-transport-native-epoll` should be the same as the Netty version installed as a dependency with aleph... I don't know if this can be handled automatically somehow, in the meantime, you can use `lein deps :tree` and look in the output:

```
[aleph "0.4.6"]
   [io.netty/netty-codec-http "4.1.25.Final"]
   [io.netty/netty-codec "4.1.25.Final"]
   [io.netty/netty-handler-proxy "4.1.25.Final"]
     [io.netty/netty-codec-socks "4.1.25.Final"]
   [io.netty/netty-handler "4.1.25.Final"]
   [io.netty/netty-resolver-dns "4.1.25.Final"]
     [io.netty/netty-codec-dns "4.1.25.Final"]
   [io.netty/netty-resolver "4.1.25.Final"]
   [io.netty/netty-transport-native-epoll "4.1.25.Final"]
   ```

PR #480, opened 2019-02-08
KQueue support

kachayev:
> Quick note. Marked  as WIP. Relies on #477 to be merged beforehand, so I can implement proper support for KQueue settings for DNS resolvers.

There're still a few things I want to discuss.

## KQueue

Overall it works the same way `:epoll?` works for TCP, HTTP, WebSockets, UDP and DNS by specifying `:kqueue?` option. `false` by default, to be consistent with `epoll?`. At first glance it feels a bit weird to carry two arguments `epoll?` and `kqueue?` here and there, but on the hand that gives me the ability to switch ON both native transports whenever possible (both will not be available on the same host).

`io.netty/netty-transport-native-epoll` dependency should also be specified with appropriate classifier ("osx-x86_64"). This is the same problem we have in #475. With the recent commits, I updated `project.clj` to include classifiers right away, both for epoll and kqueue JARs. Generally, this should work, apart from the fact, your JAR would be 180kb larger. It also may lead to some problems with transitive dependencies from other libraries, but I don't have a lot of experience with classpath resolution order when using classifiers. So, let's discuss it. The obvious advantage here: less confusion for users when working with native transports.

Netty `4.1.33.Final` has a weird [issue](https://github.com/netty/netty/issues/8849) with KQueue selector which I'm still struggling to catch. Hope the next version of Netty will fix this, so we can ship w/o concerns. Updated: [seems to be fixed](https://github.com/netty/netty/pull/8881) already.

## Unix Domain Socket

Unix domain socket support added for TCP & HTTP both for client and server. Unix socket works only with native transports (epoll or kqueue), Netty does not support it with NIO (yet?).

TCP API

```clojure
user=> (def s1 (tcp/start-server (fn [s _] (s/connect s s)) {:unix-socket "/tmp/tcp-server.sock"}))
#'user/s1
user=> (def c1 @(tcp/client {:unix-socket "/tmp/tcp-server.sock"}))
#'user/c1
user=> (s/put! c1 "hello")
<< … >>
user=> @(s/take! c1)
#object["[B" 0x2eb44ff0 "[B@2eb44ff0"]
user=> (.close s1)
nil
```

HTTP API

```clojure
nil
user=> (def s1 (http/start-server (fn [_] {:status 200 :body "OK"}) {:unix-socket "/tmp/server1.sock"}))
#'user/s1
user=> (def p (http/connection-pool {:connection-options {:unix-socket "/tmp/server1.sock"}}))
#'user/p
user=> (-> (http/get "http://domain/info" {:pool p}) deref :status)
200
```

There are a few points to mention here. When connecting to unix domain socket, we still require a valid parsable URL (host + scheme). We still need hostname to acquire a connection from the pool, setup SSL handler (improved, not trying to set SNI hostname if no `InetSocketAddress` available) and render "Host" header (this might be required by the server). `curl` has the same problem, so I think that's a reasonable trade-off. Quite a few tools use `http+unix://<socket-file>/<path>` to convey both unix socket filepath and "fake" domain, if that's interesting for users we can always implement this on top of the existing functionality.

From an implementation point of view, there're still some rough edges I'm not sure about. Notably, `netty/create-client` accepts `remote-address` option in the form of `InetSocketAddress`, we can relax that to `SocketAddress` and pass `DomainSocketAddress` right away... but that doesn't work in case we need to pass both inet (with host, port) and domain address (to connect to). `create-client` now takes both params, dynamically choose what address to connect to and rejects SSL handler setup if no `InetSocketAddress` provided (throwing exception in case `ssl-context` given with no host/port information). Overall, the number of positional arguments for `netty/create-client` looks awful right now, I'll probably reimplement that to accept a map of params (done).

Proxy options are rejected when connecting to unix domain socket, DNS resolver is set to `:noop` to bypass any attempts to resolve host before sending the request (we do just the same when sending requests through proxies).

@ztellman Waiting for your review!

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -21,7 +21,8 @@
      URI
      InetSocketAddress]
     [java.util.concurrent
-     TimeoutException]))
+     TimeoutException]
+    [io.netty.channel.unix DomainSocketAddress]))
 
 (defn start-server
   "Starts an HTTP server using the provided Ring `handler`.  Returns a server object which can be stopped
@@ -54,16 +55,26 @@
   "Returns a deferred that yields a function which, given an HTTP request, returns
    a deferred representing the HTTP response.  If the server disconnects, all responses
    will be errors, and a new connection must be created."
-  [^URI uri options middleware on-closed]
+  [^URI uri {:keys [unix-socket] :as options} middleware on-closed]
   (let [scheme (.getScheme uri)
-        ssl? (= "https" scheme)]
+        ssl? (= "https" scheme)
+        remote-address (cond
+                         (and (some? unix-socket)
+                              (instance? DomainSocketAddress unix-socket))
+                         unix-socket
+
+                         (some? unix-socket)
+                         (DomainSocketAddress. ^String unix-socket)
+
+                         :else
+                         (InetSocketAddress/createUnresolved
+                          (.getHost uri)
+                          (int
+                           (or
+                            (when (pos? (.getPort uri)) (.getPort uri))
+                            (if ssl? 443 80)))))]
     (-> (client/http-connection
-          (InetSocketAddress/createUnresolved
-            (.getHost uri)
-            (int
-              (or
-                (when (pos? (.getPort uri)) (.getPort uri))
-                (if ssl? 443 80))))
+          remote-address
           ssl?
           (if on-closed
             (assoc options :on-closed on-closed)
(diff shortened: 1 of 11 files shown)


src/aleph/http.clj
```diff
@@ -272,6 +294,23 @@
...
+           ;; "/images/json" is not a valid URL, no protocol, no host
+           ;; Which is a bad thing when I need to setup a connection to
+           ;; a remote address, but it's perfectly valid for unix socket
+           ;; What I can do here: take URL requested, try to parse it
+           ;; and if it fails just to put some dummy values to a host & port
+           ;; SSL will not work, but if don't pass any server name I should
+           ;; not expect it work at the first place. The only thing...
+           ;; I don't know the connection pool configuration here, meaning
+           ;; I have no idea if it's allowed to pass URI with no additional
+           ;; information
```
ztellman: Whatever is here will get passed through to `aleph.http.client/http-connection` by way of `aleph.http/create-connection`, which does assume the object is a `URI`, but does not assume any of the fields are populated.  I think it's okay to populate everything with dummy values.

kachayev:
Working on merging latest changes (there are quite a few conflicts because of changes to Netty module). After that is done, I'm going to revert changes about unix socket support keeping KQueue. Unix sockets implementation still has some rough edges I don't want to risk about. KQueue, on the other hand, seems to be a pretty straightforward addition.

kachayev:

Issue #478, opened 2019-02-06
Websocket client should let the server to reply to a closing frame


kachayev:
[Here](https://github.com/ztellman/aleph/blob/d35d5bbbb090c31f3f53997ee1e0d68fc7b03044/src/aleph/http/client.clj#L696-L697) we're closing the connection right after `CloseWebSocketFrame`was flushed. RFC says that client should let the server send a reply to avoid a problem with losing data w/o a proper closing "handshake". Ideally, we need to send a frame and then let the server do the job. To prevent any issues with that, we can apply a timeout to force close the connection after (if the server did not succeed).

I'll probably fix this one with #470 as they're tightly coupled.

kachayev:
Covered by #481.

Issue #470, opened 2019-01-29
Close websocket connection with custom code and reason text


kachayev:
Right now both server and client connection send `CloseWebSocketFrame` with default status code and empty reason text when the appropriate stream is drained. I think we can add `websocket-close!` (analog with `websocket-ping`) helper to take care of the closing handshake and cleanup stream & connection after that. So it will look like

```clojure
(def conn <...>)
;; returns deferred that yields when connection is closed
(http/websocket-close! conn 4009 "Idle timeout, sorry (send more pings next time)")
```

ztellman:
What does the `4009` represent here?

kachayev:
Application-level status code (range 4000-4999), just randomly picked up one. Here's the full list of codes from [rfc6455](https://tools.ietf.org/html/rfc6455#section-7.4)

kachayev:
I would say we need all 3 options:

```clojure
(websocket-close! conn)
(websocket-close! conn status-code)
(websocket-close! conn status-code reason)
```

Defaults are well defined, so it's clear what to use instead of missing params.

ztellman:
Okay, works for me.

kachayev:
@ztellman This functionality clashes with https://github.com/ztellman/aleph/pull/422 pretty heavily, so I'll make a new PR after that one is merged, so I can work around merge issue beforehand.

kachayev:
Covered by #481.

PR #481, opened 2019-02-09
Fine-grained websocket close handshake API

kachayev:
> Quick note. This functionality has a lot of merge conflicts with #422.

Covers #470 and #478. 

@ztellman The only one thing I have doubts about... I had to update `netty/sink` to "close" itself when receiving a specific message, [here](https://github.com/ztellman/aleph/compare/master...kachayev:ft-websocket-close?expand=1#diff-95720affa86696b46c5458a71ab3e8d7R372). Should I just use `.markClosed` or I need to call `(s/close! this)`? All attempts to reimplement this without "self-closing" led to potential race conditions, so I choose this approach as the most reasonable.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -249,15 +249,12 @@
    client waits for the connection to be closed by the server (no longer than close handshake
    timeout, see websocket connection configuration for more details)."
   ([conn]
-   (websocket-close! conn 1000 "" nil))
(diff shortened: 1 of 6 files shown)


kachayev:
Merged with the latest master to avoid conflicts. Part of this functionality is still not implemented tho' because it's intertwined with another PR.

Issue #452, opened 2018-12-18
How to limit max size of the request body?


kachayev:
Which is a typical safety requirement, something like `client_max_body_size` directive for Nginx.

Netty's `HttpObjectAggregator` [provides](https://github.com/netty/netty/blob/b6d6d98404f216602e0dec8efd83e140c5ff7e00/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L115-L117) this functionality out of the box. The problem with introducing the same kind of settings for our handlers is that the error most probably will be detected after the user's handler was invoked (when reading `:body` stream). Closing input stream from Aleph side would probably lead to unexpected behavior. Maybe we can provide some kind of a helper to "read body but not more than X bytes" that will throw an exception when the body is too large (it also would be nice to have the same helper to work correctly both with `InputStream` and manifold's stream for raw handlers)?

Another interesting question here: if I return `413 Entity Too Large` response from the handler, how can I force Aleph to close the connection **after** the response is sent? Regarding RFCs the client have to monitor server response even while transfering request body and have to stop sending it in case of the response arrived earlier. In practice, this is usually not handled well.. so mostly probably I would like to close the connection to make sure that if a client misbehaves - that is its problem, not mine.

kachayev:
I think we can add an additional setting, e.g. `:max-request-body-size`, that will effectively change the execution flow:

1. Aggregate body **before**  handler's invocation
2. Respond with 413 status code and close the connection in case a request body exceeds the limit given

This is not the best option when handling larger requests, but it might be super useful to cover and protect those use cases when I know in advance that my server is working with small (or relatively small) payloads, e.g. RPC server or forms w/o file uploading functionality etc. This might be done by introducing Netty's `HttpObjectAggregator` into the pipeline or manually dealing with buffering (at least `ring-handler` is already doing this to follow request buffer capacity setting.

dspearson:
This would be very nice to have. One of my projects is in a fairly early state of development, but this will be a necessary feature at some point; it's designed to be a standalone application, but exposed to a potentially-hostile network environment.

Thank you for your work so far. It is appreciated very much.

jhimanka:
https://github.com/ring-clojure/ring/issues/297 
Maybe relevant? This would be useful for us as well. Could be at Aleph, Ring or reverse proxy level.

arnaudgeiser:
> This might be done by introducing Netty's HttpObjectAggregator into the pipeline

[...]

PR #484, opened 2019-02-11
Add support for HttpObjectAggregator in server pipeline

alexander-yakushev:
Symmetrical to #393, this PR adds support for HttpObjectAggregator being added to the server pipeline. It also makes possible to solve #452 from the user code. Also, even if #452 is someday fixed by introducing `:max-request-body-size` option, this PR can be a vessel for it.

Regarding the implementation, notice that we don't release the content ByteBuf immediately unlike with other *HttpContent frames. I tried doing that, and Netty errors with refcount being negative. My understanding is that it is released [here](https://github.com/ztellman/aleph/blob/master/src/aleph/http/server.clj#L187) together with the whole request.

I also took the liberty of removing an unused import.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -38,6 +38,7 @@
      TooLongFrameException]
     [io.netty.handler.codec.http
      DefaultFullHttpResponse
+     FullHttpRequest
      HttpContent HttpHeaders HttpUtil
      HttpContentCompressor
      HttpRequest HttpResponse
@@ -272,6 +273,16 @@
               (handle-request ctx req s))
             (reset! request req)))
 
+        process-full-request
+        (fn [ctx ^FullHttpRequest req]
+          ;; HttpObjectAggregator disables chunked encoding, no need to check for it.
+          (let [content (.content req)
+                body (when (pos? (.readableBytes content))
+                       (netty/buf->array content))]
+            ;; Don't release content as it will happen automatically once whole
+            ;; request is released.
+            (handle-request ctx req body)))
+
         process-last-content
         (fn [ctx ^HttpContent msg]
           (let [content (.content msg)]
@@ -350,6 +361,12 @@
       ([_ ctx msg]
         (cond
 
+          ;; Happens when io.netty.handler.codec.http.HttpObjectAggregator is part of the pipeline.
+          (instance? FullHttpRequest msg)
+          (if (invalid-request? msg)
+            (reject-invalid-request ctx msg)
+            (process-full-request ctx msg))
+
           (instance? HttpRequest msg)
           (if (invalid-request? msg)
             (reject-invalid-request ctx msg)
@@ -397,6 +414,21 @@
       :channel-read
       ([_ ctx msg]
         (cond
+
+          ;; Happens when io.netty.handler.codec.http.HttpObjectAggregator is part of the pipeline.
+          (instance? FullHttpRequest msg)
+          (if (invalid-request? msg)
+            (reject-invalid-request ctx msg)
+            (let [^FullHttpRequest req msg
+                  content (.content req)
+                  ch (netty/channel ctx)
+                  s (netty/source ch)]
+              (when-not (zero? (.readableBytes content))
+                ;; Retain the content of FullHttpRequest one extra time to
+                ;; compensate for it being released together with the request.
+                (netty/put! ch s (netty/acquire content)))
+              (s/close! s)
+              (handle-request ctx req s)))
 
           (instance? HttpRequest msg)
           (if (invalid-request? msg)
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -15,7 +15,7 @@
   (:import
    (aleph.utils ConnectionTimeoutException RequestTimeoutException)
    (io.netty.channel ChannelHandlerContext ChannelOutboundHandlerAdapter ChannelPipeline ChannelPromise)
-   (io.netty.handler.codec.http HttpMessage)
+   (io.netty.handler.codec.http HttpMessage HttpObjectAggregator)
    (java.io File)
    (java.util.concurrent TimeoutException)
    (java.util.zip GZIPInputStream ZipException)))
@@ -570,3 +570,41 @@
                                       (.write ctx msg p)))))})
       (let [resp @(http-get (str "http://localhost:" port "/string"))]
         (is (= test-header-val (get (:headers resp) test-header-name)))))))
+
+(defn add-http-object-aggregator [^ChannelPipeline pipeline]
+  (let [max-content-length 5]
+    (.addBefore pipeline
+                "request-handler"
+                "http-object-aggregator"
+                (HttpObjectAggregator. max-content-length))))
+
+(deftest test-http-object-aggregator-support
+  (with-server (http/start-server
+                basic-handler
+                {:port port
+                 :pipeline-transform add-http-object-aggregator})
+    (let [rsp @(http-put (str "http://localhost:" port "/echo")
+                         {:body "hello"})]
+      (is (= "hello" (bs/to-string (:body rsp))))
+      (is (= 200 (:status rsp))))
+
+    (let [rsp @(http-put (str "http://localhost:" port "/echo")
+                         {:body "hello, world!"})]
+      (is (= 413 (:status rsp)))
+      (is (empty? (bs/to-string (:body rsp)))))))
+
+(deftest test-http-object-aggregator-raw-stream-support
+  (with-server (http/start-server
+                basic-handler
+                {:port port
+                 :raw-stream? true
+                 :pipeline-transform add-http-object-aggregator})
+    (let [rsp @(http-put (str "http://localhost:" port "/echo")
+                         {:body "hello"})]
+      (is (= "hello" (bs/to-string (:body rsp))))
+      (is (= 200 (:status rsp))))
+
+    (let [rsp @(http-put (str "http://localhost:" port "/echo")
+                         {:body "hello, world!"})]
+      (is (= 413 (:status rsp)))
+      (is (empty? (bs/to-string (:body rsp)))))))


kachayev:
I've asked [this](https://github.com/ztellman/aleph/pull/393#issuecomment-399250338) previously...  If we want to add full support for `HttpObjectAggregator`, maybe we should also add configuration to include `HttpObjectAggregator` with given settings? In some cases, this might be helpful (rejecting large requests, working with inlined execution instead of `response-executor` etc).

alexander-yakushev:
I agree, and it looks to me that such functionality could build upon such PR later.

kachayev:
Sure. I was going to work on #452 somewhat later. And #482 needs to be merged before introducing full-fledged support for aggregation, in fact, `HttpObjectAggregator` handles 100-continue separately.

kachayev:
Regarding releasing the content... You're right, it will happen [here](https://github.com/ztellman/aleph/blob/master/src/aleph/http/server.clj#L187) and it somewhat "accidental". The previous implementation relied on the fact, that it releases `HttpMessage` and the body was already released by the handler. Now we have `FullHttpResponse` which is `HttpMessage` + body, meaning that releasing actually frees both of them, so we should not release it earlier. Feels kinda fragile to me, but that's what you have for aggregated messages.

kachayev:
@alexander-yakushev I think for both #483 and #484 we need to implement the same functionality for raw-* handlers. Otherwise, it might break the contract when setting `:raw-stream? true`. Happy to update the implementation later, just need to set a reminder for myself.

alexander-yakushev:
I thought about it a little, and to me it doesn't make much sense to combine HttpObjectAggregator with `raw-streams?`. The latter is for cases when you want to consume the body asynchronously, and HttpObjectAggregator by definition disables asynchronous handling.

I don't mind having it there for "completeness" though, just want to be sure it doesn't add more confusion than value.

kachayev:
You still can use `:raw-stream?` for example to get access to Netty's `ByteBuf`s. I'm mostly worried about the confusion when switching from ring handler to a raw stream realizing that they process different sets of messages from your pipeline.

alexander-yakushev:
Looked at it again; this now looks more tricky than I thought. If I understand correctly, users are expected to release ByteBufs manually in `raw-stream?` mode. So, the content of FullHttpRequest shouldn't be released, yet it does get released together with the request in `handle-request`. We need somehow to release FullHttpRequest without the content. Perhaps, by `acquire`ing it an extra time? It all starts to look even more wobbly at this point.

kachayev:
@alexander-yakushev Well... we need to call `handle-request`, put a single message with the content to a stream and close it right after. In such a case we can probably retain `content` before putting on the stream. In fact, `FullHttpRequest` just proxies retain/release method calls to a `content`.

alexander-yakushev:
Added support for HttpObjectAggregator for the raw handler too. Tested only rudimentary, to the best of my ability:

```clj
(require '[aleph.http :as http]
         '[manifold.stream :as ms])

(def srv (http/start-server
          (fn [r] (println @(ms/reduce conj [] (:body r))))
          {:port 12345
           :raw-stream? true
           :pipeline-transform #(.addAfter % "http-server" "aggreg"
                                           (io.netty.handler.codec.http.HttpObjectAggregator. 1000))}))

...

;; curl -d '{}' 'localhost:12345'
[#object[io.netty.buffer.CompositeByteBuf 0x5f926fa4 CompositeByteBuf(ridx: 0, widx: 2, cap: 2, components=1)]]
```

src/aleph/http/server.clj
```diff
@@ -378,6 +394,21 @@
...
+          (if (invalid-request? msg)
+            (reject-invalid-request ctx msg)
+            (let [^FullHttpRequest req msg
+                  content (.content req)
+                  s (netty/source (netty/channel ctx))]
+              (when-not (zero? (.readableBytes content))
+                ;; Retain the content of FullHttpRequest one extra time to
+                ;; compensate for it being released together with the request.
+                (.retain content)
+                (netty/put! (.channel ctx) s content))
```
kachayev: nit: you took channel twice, using different APIs

kachayev: another nit: there's `netty/acquire` wrapper for `.retain` operations, you can do

```clojure
(netty/put! (.channel ctx) s (netty/acquire content))
```

alexander-yakushev: 👌

DerGuteMoritz:
FWIW, this patch looks good to me! I rebased it onto latest `master` and added a few test cases for it over here: https://github.com/bevuta/aleph/tree/server-aggregator

KingMob:
Thanks for tackling this, and thanks for the tests! It seems ok, but we need to double-check a lot of this old code.

- How does it interact with #483, as they mentioned above? 
- Also, is it releasing all the necessary bytebufs correctly? Check with `ResourceLeakDetector.setLevel()` or `-Dio.netty.leakDetectionLevel=PARANOID`
- What's the error-handling situation look like?

DerGuteMoritz:
> * How does it interact with [[client] Don't close connections when HttpObjectAggregator is on the pipeline #483](https://github.com/clj-commons/aleph/pull/483), as they mentioned above?

The discussion above where this PR is referenced only did so to consider the `:raw-stream?` case in that one, too. That case was subsequently addressed in this here PR by increasing the reference count before handing it over via the stream so that the contract that the individual `ByteBuf`s must be released by the consumer is upheld. This looks like a reasonable solution to me and appears to work as intended.

> * Also, is it releasing all the necessary bytebufs correctly? Check with `ResourceLeakDetector.setLevel()` or `-Dio.netty.leakDetectionLevel=PARANOID`

This should be covered by means of having the test cases, right? Note https://github.com/clj-commons/aleph/blob/master/test/aleph/http_test.clj#L29

> * What's the error-handling situation look like?

Should be fine as is. A `FullHttpRequest` is essentially a combination of a `HttpRequest` and a `LastHttpContent`, so the patch basically just does the same thing as for these individual pieces in one go (save for the extra reference count increment mentioned above). To be sure, I'll add a test case which intentionally throws an exception from the handler, though. Sound good?

alexander-yakushev:
Hello @KingMob!

1. #483 is independent of this because that PR fixes the HTTP client behavior, and this PR is about the aggregator in the server code.
2. I can't properly check it right now, but this code has been running in different heavy-loaded scenarios for almost 3 years, no issues detected.

src/aleph/http/server.clj
```diff
@@ -378,6 +394,21 @@
...
+          (if (invalid-request? msg)
+            (reject-invalid-request ctx msg)
+            (let [^FullHttpRequest req msg
+                  content (.content req)
+                  ch (netty/channel ctx)
+                  s (netty/source ch)]
+              (when-not (zero? (.readableBytes content))
+                ;; Retain the content of FullHttpRequest one extra time to
+                ;; compensate for it being released together with the request.
+                (netty/put! ch s (netty/acquire content)))
```
arnaudgeiser, line 408: The comment and the approach is correct there regarding the expected behavior of `raw-ring-handler`.

Without the `HttpObjectAggregator` on the pipeline, we are receiving first the `HttpRequest` (io.netty.buffer.ByteBuf) which doesn't contain the data of the HTTP request and we initialize the `manifold.stream`. The `HttpRequest` is released later by `handle-request`.
And then we are receiving (multiple) `HttpContent` (io.netty.buffer.ByteBuf) which will be put directly on the `manifold.stream` without any ref counting management.

When using the `HttpObjectAggregator`, we need to `retain` one more time to let the `handle-request` releases the `ByteBuf` and then the client of the stream to release the `ByteBuf` too.

An alternative would be to not perform the `retain` and to provide an additional parameter to `handle-request` to not perform the `release`. But it's an optimization that might not be needed at all but I wanted to mention it.

arnaudgeiser (approved):
Fine for me with a small comment.
As it's running fine on production (on Grammarly?), I assume it's safe to merge this one the way it is today.

src/aleph/http/server.clj
arnaudgeiser: (Netty 5 will remove the notion of refcounting, but this is another story...)

arnaudgeiser:
The only thing that bother me a bit is the fact we don't have any tests with a pipeline containing `HttpObjectAggregator`. (only for the websocket)

src/aleph/http/server.clj
DerGuteMoritz: > An alternative would be to not perform the retain and to provide an additional parameter to handle-request to not perform the release. But it's an optimization that might not be needed at all but I wanted to mention it.

Good thought! This would be a breaking change, though... I suspect that this wouldn't be worth it as bumping the refcount should be a pretty cheap operation.

DerGuteMoritz:
> This should be covered by means of having the test cases, right? Note https://github.com/clj-commons/aleph/blob/master/test/aleph/http_test.clj#L29

Welp, maybe not: I am not able to consistently trigger the leak detector even when I intentionally leak buffers in `:raw-stream? true`  handlers with paranoid mode enabled as per the cited line. Only about 1 in 20 runs ends up triggering the leak detector. Also tried adding `(System/gc)` and `(System/runFinalization)` to the end of the test but no dice. This is true even when `HttpObjectAggregator` is not involved. 

FTR: I noticed this when trying to add a test case with an intentionally throwing handler: AFAICT, in the `:raw-stream? true` case, such an unhandled exception would always lead to a leak since the default error handler doesn't care about releasing the bytebufs in the request body stream. This is of course an orthogonal problem which I will file a separate issue about. But not quite sure yet how to go about this with the leak detector being so unreliable :disappointed:  Any hints appreciated.

DerGuteMoritz:
> The only thing that bother me a bit is the fact we don't have any tests with a pipeline containing `HttpObjectAggregator`. (only for the websocket)

I added some in my rebased version of this branch: https://github.com/bevuta/aleph/commit/a4d03a2f736121ddccd0aaac5bca2f597d51366a 

@alexander-yakushev Feel free to update this PR to include my changes. If you'd rather not be bothered, I can also open a fresh one based on my branch!

arnaudgeiser:
> FTR: I noticed this when trying to add a test case with an intentionally throwing handler: AFAICT, in the :raw-stream? true case, such an unhandled exception would always lead to a leak since the default error handler doesn't care about releasing the bytebufs in the request body stream. This is of course an orthogonal problem which I will file a separate issue about. But not quite sure yet how to go about this with the leak detector being so unreliable disappointed Any hints appreciated.

This is unfortunately too late for Aleph to take care of this.
When using `:raw-stream?`, you have to be aware that some of your Netty ByteBuf on your body (`manifold.stream`)  will never reach your application code in case of Exception.
That's your responsibility to catch it, close the stream and drain it to apply `.release`.

This would be a breaking change with huge consequences.
As soon as you are using `io.netty.buffer.PooledByteBufAllocator` (and thus `io.netty.buffer.PooledByteBuf`) the last call to `.release` will bring back the ByteBuf to the pool which can be reused by something else.
On the worst case scenario, we might end up calling `.release` twice on the same `ByteBuf` where the second call could release something that does not belong to us anymore.

To be, the only thing we can do here is better documentation.

alexander-yakushev:
Rebased and cherry-picked tests commit off @DerGuteMoritz's branch.

arnaudgeiser (approved):

test/aleph/http_test.clj
```diff
@@ -570,3 +570,40 @@
...
+      (is (= 200 (:status rsp))))
+
+    (let [rsp @(http-put (str "http://localhost:" port "/echo")
+                         {:body "hello, world!"})]
+      (is (= 413 (:status rsp)))
+      (is (empty? (bs/to-string (:body rsp)))))))
+
+(deftest test-http-object-aggregator-raw-stream-support
+  (with-server (http/start-server
+                basic-handler
```
arnaudgeiser, line 597: After a second pass on this PR, I would consider adding a `.release` on
the handler regarding `raw-stream?` to validate this :

> ;; Retain the content of FullHttpRequest one extra time 
;; compensate for it being released together with the request.

DerGuteMoritz: Oh yeah, this is a bit sublte: For a while I was stumped by the fact that I couldn't trigger the leak detector because it seemed like I was leaking the buffer here. After digging into Netty a bit more, I learned why: Note how `echo-handler`  returns the `(:body request)` manifold stream verbatim as the response body. This leads to Netty itself releasing the buffers after writing them out to the wire. Makes sense when you think about it! Unfortunately, I was unable to find a reference document mentioning this but if it's anything to go by, the official user guide's ["Writing an Echo Server" section](https://netty.io/wiki/user-guide-for-4.x.html#wiki-h3-8) (which essentially does the same as we do here) mentions this in passing:

> It is because Netty releases it for you when it is written out to the wire.

So in fact, adding a `.release` call here would result in a double-free :grimacing: If you have an idea how to make this test rely on a little less context, I'm all ears, though :smile:

arnaudgeiser: TIL, thank you.

arnaudgeiser:
Whether @KingMob is fine with the state of this PR, I would merge it.

test/aleph/http_test.clj
```diff
@@ -570,3 +570,40 @@
                                       (.write ctx msg p)))))})
       (let [resp @(http-get (str "http://localhost:" port "/string"))]
         (is (= test-header-val (get (:headers resp) test-header-name)))))))
+
+(defn add-http-object-aggregator [^ChannelPipeline pipeline]
+  (.addBefore pipeline
+              "request-handler"
+              "http-object-aggregator"
+              (HttpObjectAggregator. 5)))
```
KingMob, line 578: Can we document that this means the max length is 5 bytes, either via docstring or replacing the magic num witha var?

KingMob (approved):
I have a minor suggestion about documenting the usage of HttpObjectAggregator, but otherwise LGTM

test/aleph/http_test.clj
arnaudgeiser: I guess it's good enough like this:

https://github.com/clj-commons/aleph/pull/484/commits/5d2689218eb603fec67bea1e08999a1c403608bd#diff-b7e83256bff42d68457008e51789109ff14c053e0d6de64c6ffa03d1d11b3dcfR575

arnaudgeiser:
Thank you @alexander-yakushev and @DerGuteMoritz for the hard work!
Let's say it's never too late :)

Issue #469, opened 2019-01-27
HTTP server to send a specific file region to deal with content range requests


kachayev:
Right now `aleph.http.core/send-file-region` internally creates an instance of `DefaultFileRegion` using `0..len` range (entire file). If I need to accept `Range` request header and reply with `HTTP/1.1 206 Partial Content`, I have to be able to specify custom offset and length params.

kachayev:
I think we can add this functionality by introducing the following API:

```clojure
(require '[aleph.http.core :as http])
(require '[clojure.java.io :as io])

;; manually create a record
;; we can also use simple Java class instead of record?
(defn handler [req]
  {:status 200
   :body (http/PartialFileContent. (io/file filepath) 1e3 1e4)})

;; helper to construct an instance of the same record?
(defn handler [req]
  {:status 200
   :body (http/partial-file (io/file filepath) 1e3 1e4)})
```

Changes to `send-file-region` and `send-chunked-file` are quite obvious, for `ssl?` set to `true` use case we also need to construct `manifold.stream` source carefully, but that's also doable. @ztellman What do you think about API?

ztellman:
I think I'd like to have a general `http/file` constructor that optionally takes range parameters (and defaults to the entire file), and a coercion that turns `java.io.File` (and maybe `java.nio.file.Path`?) into our internal file representation.

kachayev:
Closing the issue as the functionality was already merged.

PR #485, opened 2019-02-11
http/file API to send region of the file

kachayev:
> Quick note. This PR clashes with #471 almost entirely. So.. Either #471 should be merged first, so I can reimplement the current one, or this one should be merged first, so I can reimplement #471.

Covers #469. The key motivation factor is byte range requests.

Now I can send as a body:

* `java.io.File`
* `java.nio.file.Path`
* `(aleph.http/file path)`
* `(aleph.http/file path offset length)`

It also extends API so I can set custom chunk size.

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -340,7 +340,7 @@
   (HttpFile. fd 0 (.length fd) default-chunk-size))
 
 (defn send-chunked-file [ch ^HttpMessage msg ^HttpFile file]
-  (let [raf (RandomAccessFile. (.-fd file) "r")
+  (let [raf (RandomAccessFile. ^File (.-fd file) "r")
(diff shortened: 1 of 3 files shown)


src/aleph/http/core.clj
```diff
@@ -330,34 +338,128 @@
...
+
+(deftype HttpFile [^File fd ^long offset ^long length ^long chunk-size])
+
+(defmethod print-method HttpFile [file ^java.io.Writer w]
+  (.write w (format "HttpFile[fd:%s offset:%s length:%s]"
+                    (.-fd file)
+                    (.-offset file)
+                    (.-length file))))
+
+(defn new-http-file
```
ztellman: This should just be `http-file`.

kachayev: Done!

PR #487, opened 2019-02-16
SNI, enhanced SSL, static name resolver

kachayev:
> Quick note, this PR has some overlapped functionality with #480.

## SNI

* configuration options for `http/connection-pool` to disable SNI or provide peer host value manually (effectively, emulating `openssl client -servername <host>`)  

* server now supports SNI when `:sni` option is provided with mapping from domain names to appropriate `SslContext`

Netty's `SniHandler` (even `AbstractSniHandler` to be more specific) is implemented in quite liberal fashion: it requires default `SslContext` to be provided falling back to it when no matching hosts found, when `server_name` is not provided, when there's an issue with the `ClientHello` etc... Right now I do the same for Aleph configuration, just throwing `IllegalArgumentException` when no default context is provided. Ideally, there should be an option to implement strict validation, e.g. reject non-SNI clients.

P.S. It turned out that JDK does [the same](http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/3bb8137f36a8/src/share/classes/sun/security/ssl/ServerHandshaker.java#l363), silently ignoring SNI matchers when `ClientHello` has no `server_name` extension at all. 

## SSL

* extended version of `netty/ssl-client-context` and `netty/ssl-server-context` to manage `SslContext` objects w/o Java interop

* `ssl-context` now can be configured passing a map of options instead of calling a helper or juggling with Java interop, so you can do e.g.

```clojure
:ssl-context {:private-key <>
              :protocols ["TLSv1.1" "TLSv1.2"]}
```

* client catches SSL handshake exceptions and returns them to the caller instead of silently closing the connection after the timeout

* server does not log a full stack trace for failed SSL handshake, just a single line warning 

## Static Name Resolver

This was actually done mostly to test SNI support but it's still a useful standalone feature. `netty/static-resolver-group` creates an address resolver group to be used as a param for `http/connection-pool` that resolves the list of given hosts to given IP addresses, similar to what `curl --resolve <mapping>` does.


kachayev:
SNI seems to be quite a niche functionality in exchange of such a complexity. Closing PR in favor of #546.

PR #488, opened 2019-02-19
Custom DNS resolver and proxy support for WebSocket clients.

dspearson:


kachayev:
@dspearson thanks! I have a couple of concerns regarding this functionality:

1. Most probably we want to provide two different options: dns-options (to create a new DNS resolver) and name-resolver (to reuse the existing one). With the connection pool it’s more intuitive, as you spin and reuse the same resolver per each connection created within the pool. But here I’m forced to spin new resolver on each connection open (in most cases to resolve a single hostname). Which might be an overkill... 

2. I don’t have experience with proxies for WS and WSS, but it feels like we need to set “Proxy-Connection: Keep-Alive” to ensure that the proxy would not close connection to the peer right away. What do you think?

dspearson:
I haven't really looked at the WS spec in detail, nor looked at the methods for maintaining an open connection. However, I see that in the http-proxy-headers function, there is
`      (.set headers "Proxy-Connection" "Keep-Alive"))
`
and this appears to be called as part of the chain.

Regarding the first point, I have addressed it in a follow-up commit by simply removing dns-options, and allowing the passing of a pre-instantiated instance of the name resolver via name-resolver.

Cheers!

kachayev:
Yeah, there's a `:keep-alive?` setting that's set to `true` by default and it effectively adds an appropriate header. What we probably want to do here, is to force this setting to be `true` (just `assoc` it when creating the handler).

src/aleph/http.clj
```diff
@@ -205,7 +205,20 @@
...
+   | `heartbeats` | optional configuration to send Ping frames to the server periodically (if the connection is idle), configuration keys are `:send-after-idle` (in milliseconds), `:payload` (optional, empty frame by default) and `:timeout` (optional, to close the connection if Pong is not received after specified timeout).
+   | `name-resolver` | specify the mechanism to resolve the address of the unresolved named address. When not set or equals to `:default`, JDK's built-in domain name lookup mechanism is used (blocking). Set to`:noop` not to resolve addresses or pass an instance of `io.netty.resolver.AddressResolverGroup` you need.
+   | `proxy-options` | a map to specify proxy settings. HTTP, SOCKS4 and SOCKS5 proxies are supported. Note, that when using proxy `connections-per-host` configuration is still applied to the target host disregarding tunneling settings. If you need to limit number of connections to the proxy itself use `total-connections` setting.
+   Supported `proxy-options` are
```
kachayev: Please, add a single empty line before this one.

dspearson:
Erroneously hit close, sorry about that.

Addressed your comments. Thanks for your input.

src/aleph/http/client.clj
```diff
@@ -809,7 +811,7 @@
                           "websocket-deflater"
                           WebSocketClientCompressionHandler/INSTANCE)))
             (#(when (some? proxy-options)
-               (let [proxy (proxy-handler (assoc proxy-options :ssl? ssl?))]
+               (let [proxy (proxy-handler (assoc proxy-options :ssl? ssl? :keep-alive? keep-alive?))]
```
kachayev: Sorry, I was not clear enough... I meant just `assoc` here `true` to prevent self-shooting

```clojure
(assoc proxy-options :ssl? ssl? :keep-alive? true)
```

In such a case even if I passed `:keep-alive?`with my `proxy-options` (we never validate keys anyway), it still will be set to `true` when passing to the handler's builder. In such a case we don't need `:keep-alive?` being mentioned as an option.

kachayev:
> Erroneously hit close, sorry about that.

kachayev:
Thanks a lot!

dspearson:
I admit I still don't know my way around the code that well. I am a little dubious about the validity of associating :ssl? with the proxy-options there, having copied it from elsewhere, and looking at http-proxy-handler has me scratching my head. The value is given by looking at the WebSocket connection scheme, and appears unrelated to the protocol used by the proxy itself.

I'm not using the option myself, so haven't tested it, sorry.

kachayev:
We need `:ssl?` to be passed to proxy handler just to be sure if we need to open tunnel. It should work for Websockets the same way it works for any HTTP connection but overall HTTPS proxies are not that common.

ztellman:
@kachayev, is this ready to be merged?

kachayev:
@ztellman Yeah, this one is ready.

dspearson:
Let me know if I need to do any further changes here, since I note my branch is a little behind.

HaskellZhangSong:
Just drop by, sorry for my abrupt intrusion. 
I am using this fantastic package and this is one of the features what I am waiting for. Only one suggestion. Normally, in some traditional command line software and some packages, proxy will use `http_proxy` and `https_proxy` environment variables as proxy by default, see  python websocket package [https://github.com/websocket-client/websocket-client/blob/df275d351f9887fba2774e2e1aa79ff1e5a24bd1/websocket/_url.py#L152](python websocket) line 152.
and [https://github.com/jay/wget/blob/099d8ee3da3a6eea5635581ae517035165f400a5/src/retr.c#L1125](wget) line 1125
and [https://github.com/curl/curl/blob/c9c5304dd4747cbe75d2f24be85920d572fcb5b8/lib/url.c#L2257](curl) line 2257. 

All of them use these two environment variables as their proxy by default. I suppose that is a convention of linux/unix software/package. Would it be better if aleph also supports this?

dspearson:
This pull request is specifically only for adding the DNS resolver and proxy settings for WebSockets. Additional features for proxy selection from the environment should probably be in a different branch, if anything.

kachayev:
I'm going to rebase to 1.0.0 to merge with latest changes there. Substantial portion of the code is intertwined with websockets pipeline changes we've made so far.

kachayev:
As of now, merging into `1.0.0` branch (all conflicts are fixed).

PR #495, opened 2019-03-12
Use shared instance of LoggingHandler for the connection pool

kachayev:
`LoggingHandler` is shareable and we don't need to create a new instance per each connection. The new implementation uses a single logger for all connections within the same connections pool.

I've also moved a few helper functions to `netty` module in order to provide `log-activity` capabilities for a server (after other PRs would be merged, right now it will create a lot of merge conflicts).


kachayev:
@ztellman Merged with the master.

PR #502, opened 2019-04-06
Support raw streams for text websockets

scramjet:
Support the `raw-stream?` option for text frames so we send ByteBuf's rather than strings.

This avoids having Aleph call `TextWebSocket.text ()` which copies its ByteBuf to a string if we're only going to convert that to something else down the line (in my case from Transit to Clojure data).

### DIFF
--- a/test/aleph/websocket_test.clj
+++ b/test/aleph/websocket_test.clj
@@ -102,7 +102,7 @@
         (is @(s/put! c "text client hello"))
         (let [msg @(s/try-take! c 5e3)]
           (is (= "text client hello"
-                 (when msg (bs/to-string (netty/release-buf->array msg)))))))))
+                 (when msg (bs/to-string msg))))))))
 
   (testing "websocket server: raw-stream? with binary message"
     (with-handler raw-echo-handler
--- a/test/aleph/websocket_test.clj
+++ b/test/aleph/websocket_test.clj
@@ -88,12 +88,20 @@
       (is (= "YES" @(s/try-take! c 5e3))))))
 
 (deftest test-raw-echo-handler
-  (testing "websocket client: raw-stream?"
+  (testing "websocket client: raw-stream? with binary message"
     (with-handler echo-handler
       (let [c @(http/websocket-client "ws://localhost:8080" {:raw-stream? true})]
         (is @(s/put! c (.getBytes "raw client hello" "UTF-8")))
         (let [msg @(s/try-take! c 5e3)]
           (is (= "raw client hello"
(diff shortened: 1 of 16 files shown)


scramjet:
Sorry, should have run the tests. Will fix and update the PR.

scramjet:
Note that this changed the behaviour of clients that use `raw-stream?` with text frames, and thus the existing test case change.

This is consistent with the docs ("if `true`, the connection will emit raw `io.netty.buffer.ByteBuf` objects rather than strings or byte-arrays"), but obviously it's up to you to decide how controversial (if at all) this this.

kachayev:
@scramjet It's a breaking change but I think it's a reasonable step to do to have consistent expectations from `:raw-stream?`.

scramjet:
@kachayev Yes, it is a breaking change. It would only break handlers/clients using raw-stream? with both binary and text frames, who expect ByteBuf's for binary but still String's for text. I'd be surprised if there are any instances of that. And, technically, they're depending on undocumented behaviour ;)

kachayev:
@scramjet That's true.

test/aleph/websocket_test.clj
```diff
@@ -88,14 +88,22 @@
...
           (is (= "raw client hello"
                  (when msg (bs/to-string (netty/release-buf->array msg)))))))))
 
+  (testing "websocket client: raw-stream? with text message"
+    (with-handler echo-handler
+      (let [c @(http/websocket-client "ws://localhost:8080" {:raw-stream? true})]
+        (is @(s/put! c "text client hello"))
+        (let [msg @(s/try-take! c 5e3)]
+          (is (= "text client hello"
+                 (when msg (bs/to-string (netty/release-buf->array msg)))))))))
```
kachayev: I think we can do 

```clojure
(try (bs/to-string msg) (finally (netty/release msg)))
```

to avoid unnecessary conversion step.

scramjet: Strangely, this throws a reference count exception, indicating `msg` is has a zero ref count. Sure enough adding an acquire call in the client handler before putting it on the channel fixes that, but that's obviously not how it's supposed to work. 

The .content call `(.content ^TextWebSocketFrame msg)` should throw on zero ref count, so I can't see how it's getting changed?

scramjet: Aha, `bs/to-string` is calling release somewhere. Will update test.

jdf-id-au:
I hacked together a [TextWebsocketFrame fragmenter](https://gist.github.com/jdf-id-au/335744a20d78b2e61703ecf9e49c7511) for while yada's stuck on old aleph.

PR #503, opened 2019-04-06
Support WebSocketChunkedInput in websockets

scramjet:
Allow the client to send `WebSocketChunkedInput`'s on the channel. Assumes that client has added a `ChunkedWriteHandler` in the customise-pipleline logic.

It would probably make sense for Aleph to support this directly for websockets in future like it does for HTTP, but for now it makes sense to pass though chunked inputs if we're doing it manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Aleph exposes data from the network as a [Manifold](https://github.com/ztellman/
 
 ### HTTP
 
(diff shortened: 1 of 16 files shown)


kachayev:
@scramjet Should we allow this only when `ChunkedWriteHandler` was added into the pipeline before?

scramjet:
@kachayev A warning would be nice, but add some overhead checking the filter stack. I'd expect anyone knowledgable enough about Netty to send `WebSocketChunkedInput`'s down the line to know they also need a `ChunkedWriteHandler` in there too (and to not want the overhead of checking and warning).

Like I said, it would be nice if there were an API in Aleph to make this possibly without messing directly with Netty, but this seems like a pragmatic middle ground for now.

scramjet:
I've amended this to be more general: pass though anything implementing `ChunkedInput`.

PR #510, opened 2019-05-06
[WIP] Draft support for MQTT client

kachayev:
Publishing this earlier to get feedback.

Right now implementation covers MQTT3.1 client publish (fire-and-forget and at-most-once flows). Subscriptions are still in progress.

Flow control is implemented on top of Manifold's stream put/take operations. It makes code easier to read/maintain, but probably not the most performant approach. Potentially we can get better performance implementing flows as Netty handlers. To make the final decision I still need to implement both and run performance tests to compare both approaches.

### DIFF
--- a/src/aleph/mqtt.clj
+++ b/src/aleph/mqtt.clj
@@ -1,0 +1,307 @@
+(ns aleph.mqtt
+  (:require [aleph.netty :as netty]
+            [manifold.deferred :as d]
+            [manifold.stream :as s]
+            [manifold.bus :as bus])
+  (:import
+   [java.io Writer]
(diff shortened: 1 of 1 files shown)


kachayev:
Overall, it would be nice to have a Clojure-friendly wrapper for MQTT functionality (raw Netty version of MQTT client/server are somewhat hard to grasp). Unfortunately, I won't have time to push implementation to production-level state. Whoever has capacity and willing to push development of MQTT wrapper forward, - please, feel free to use this PR as a foundation. As of now, I'm going to close it.

P.S. Also, will be more than happy to answer questions or help with suggestions.

PR #511, opened 2019-05-21
Attach idle handlers before other handlers

czan:
Having the idle handlers after the other handlers results in them timing out the connection after the timeout without any consideration of actual traffic. This can lead to a socket being closed while it is still actively being used. Attaching the idle handlers first means that a connection will only be treated as idle if no traffic has been sent through the socket for a given period of time.

This is leading a problem in production for us, where Aleph would terminate connections while a request was in-flight (~20ms after receiving the data) because the connection was older than the request. We have disabled the idle-timeout to get things working for now, but this change will fix things properly.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -489,6 +489,7 @@
                                                    ssl?))]
 
       (doto pipeline
+        (http/attach-idle-handlers idle-timeout)
         (.addLast "http-server"
           (HttpServerCodec.
             max-initial-line-length
@@ -501,7 +502,6 @@
             (let [compressor (HttpContentCompressor. (or compression-level 6))]
               (.addAfter ^ChannelPipeline %1 "http-server" "deflater" compressor))
             (.addAfter ^ChannelPipeline %1 "deflater" "streamer" (ChunkedWriteHandler.))))
-        (http/attach-idle-handlers idle-timeout)
         pipeline-transform))))
 
 ;;;


kachayev:
@czan It generally makes sense to keep idle handlers closer to network part to be more precise tracking timeout... It might be problematic in terms of performance when streaming response, e.g. using HTTP chunks writer (but I need to run detailed tests to double check on this). Note, that current implementation of idle detector in Netty is still far from being perfect and the connection still might be closed if the client consumes or produces data slowly.

czan:
In our case, we were seeing the connecting being closed at _exactly_ the idle timeout, every time. We set the idle timeout to 65 seconds, then opened a telnet connection and constantly sent requests and received responses. After 65 seconds the connection was closed, no matter what.

We're fine with imprecise tracking of the idle timeout, as long as there's some tracking of actual traffic.

KingMob:
Hi @czan is this still a problem?

This seems like a straightforward enough change, so I'm inclined to merge, but I'd like a test to go with it, if you have the time.

czan:
Sorry, I don't work on the project that was using Aleph any more, so I'm not sufficiently motivated to write a test for this.

KingMob:
I understand. Do you know someone else who's still on it, who would be interested?

arnaudgeiser:
Hey @KingMob!

We just spotted today we have the exact same issue in one on our product.
I might have the opportunity to work on this during my working hours.

Coming back to you ASAP!

@bcachet, @mping

arnaudgeiser:
Closed in favour of this : https://github.com/clj-commons/aleph/pull/601

arnaudgeiser:
Closed in favour of this : https://github.com/clj-commons/aleph/pull/601

PR #514, opened 2019-07-02
tcp: allow exception handler to be overriden

pyr:
Some pipeline elements rely on exception for control flow.
This patch proposes a pluggable exception-handler which is fed
the current stream as well as the exception.

### DIFF
--- a/src/aleph/tcp.clj
+++ b/src/aleph/tcp.clj
@@ -24,14 +24,16 @@
 (alter-meta! #'->TcpConnection assoc :private true)
 
 (defn- ^ChannelHandler server-channel-handler
-  [handler {:keys [raw-stream?] :as options}]
+  [handler {:keys [raw-stream? exception-handler] :as options}]
   (let [in (atom nil)]
     (netty/channel-inbound-handler
 
       :exception-caught
       ([_ ctx ex]
-        (when-not (instance? IOException ex)
-          (log/warn ex "error in TCP server")))
+       (if (some? exception-handler)
+         (exception-handler @in ex)
+         (when-not (instance? IOException ex)
+           (log/warn ex "error in TCP server"))))
 
       :channel-inactive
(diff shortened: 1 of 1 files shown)


vigneshsarma:
This pr might help with #529 we are facing.

src/aleph/tcp.clj
```diff
@@ -24,14 +24,16 @@
...
     (netty/channel-inbound-handler
 
       :exception-caught
       ([_ ctx ex]
-        (when-not (instance? IOException ex)
-          (log/warn ex "error in TCP server")))
+       (if (some? exception-handler)
+         (exception-handler @in ex)
+         (when-not (instance? IOException ex)
+           (log/warn ex "error in TCP server"))))
```
vigneshsarma: Why only log if its an IOException? 
Also, wouldnt it be better if there was a default handler? 
Something like, 
```clojure
(or exception-handler #(log/warn ex "error in TCP server"))))
```

pyr: @vigneshsarma it was the previous behavior which I would rather leave untouched. It actually logs when the exception is not an IO one

kachayev:
"Netty-friendly" approach would be to fire exception down the `ChannelPipeline`... In this case, `pipeline-transform` could be used to setup additional `ChannelHandler` to deal with errors. Though I understand it might be quite challenging API to adopt, I'm still curious why do we swallow unhandled exceptions without closing the connection in the first place. Seems like this might lead to a lot of problematic situations. Let me think more about this.

Having explicit `exceptions-handler` seems like a nice approach (similar to e.g. how Vert.x exposes it). Not sure if it should **only** TCP server feature... what about HTTP server? Or other servers? Maybe we need to setup `exceptions-handler` in `netty/start-server` (as a new handler in the pipeline)? 🤔

PR #521, opened 2019-08-29
Customizable error handler

prepor:
Hey.

In case of any unhandled exception, the current behavior is print whole stracktrace into response's body. It's not what we want and there is no way to alter this behavior currently.

This PR adds new configuration option `:error-handler` to http server, this handler is called for unhandled exceptions inside HTTP machinery. By default, it's old `error-response`.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -117,13 +117,13 @@
       [server-value keep-alive-value close-value]
       (map #(HttpHeaders/newEntity %) ["Aleph/0.5.0" "Keep-Alive" "Close"])]
   (defn send-response
-    [^ChannelHandlerContext ctx keep-alive? ssl? rsp]
+    [^ChannelHandlerContext ctx keep-alive? ssl? error-handler rsp]
     (let [[^HttpResponse rsp body]
           (try
             [(http/ring-response->netty-response rsp)
              (get rsp :body)]
             (catch Throwable e
-              (let [rsp (error-response e)]
+              (let [rsp (error-handler e)]
                 [(http/ring-response->netty-response rsp)
                  (get rsp :body)])))]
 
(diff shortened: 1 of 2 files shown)


kachayev:
Hey! Some additional error handling changes were implemented here. The idea of having fully customizable error handler seems like a good addition but there are a few questions to ask first:

* should `error-handler` provide only sync interface? or it might be also async?
* should it run on it's own thread pool? should there be a way to configure executor (e.g. as it's done for `continue-handler`)?
* there probably should be a wrapper to catch errors when `error-handler` is applied (at least to close connection to avoid resources from leaking)

Let me think about this a bit more.

arnaudgeiser:
This PR has been requested multiple times to be merged on Aleph and tests prove the functionality
is working as expected.

Regarding Oleksii's comments:

> should error-handler provide only sync interface? or it might be also async?

We can provide a sync interface for now and adapt afterwards if requested.

> should it run on it's own thread pool? should there be a way to configure executor (e.g. as it's done for continue-handler)?

The `continue-executor` is a second parameter, we should be able to provide a `error-executor` if requested.

> there probably should be a wrapper to catch errors when error-handler is applied (at least to close connection to avoid resources from leaking)

Yes, probably!
But I think it can be handled afterwards as it's not the case currently.

arnaudgeiser (approved):

arnaudgeiser:
Rebased, fixed and approved on my side.

src/aleph/http/server.clj
```diff
@@ -168,13 +168,13 @@
               ;; handle request on a separate thread
               (try
                 (d/future-with executor
-                  (handler req'))
+                               (handler req'))
```
KingMob, line 171: NB: as a body macro, this should be indented only 2 spaces, for future reference

KingMob (approved):

PR #528, opened 2019-10-18
Handle 308 Permanent Redirect (same behaviour than for 307)

rborer:
Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308:

> The HyperText Transfer Protocol (HTTP) 308 Permanent Redirect redirect status response code indicates that the resource requested has been definitively moved to the URL given by the Location headers. A browser redirects to this page and search engines update their links to the resource (in 'SEO-speak', it is said that the 'link-juice' is sent to the new URL).

> The request method and the body will not be altered, whereas 301 may incorrectly sometimes be changed to a GET method.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -201,7 +201,7 @@
 
 ;; Statuses for which clj-http will not throw an exception
 (def unexceptional-status?
-  #{200 201 202 203 204 205 206 207 300 301 302 303 304 307})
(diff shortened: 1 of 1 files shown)


kachayev:
Good catch, thanks! The client code here follows `clj_http` implementation, which (as far as I see right now) does not handle 308 status code as well (source code [here](https://github.com/dakrone/clj-http/blob/4b22f0f0476174af788e3c385beb7d0ad10eee1c/src/clj_http/client.clj#L337)). So, I'm going to open similar issue there.

PR #533, opened 2020-02-07
Improve raw-stream example

rafd:
Minor tweak to `:raw-stream?` example to prevent future readers from following the example directly and ending up with a program that creates a connection-pool on every request (...eventually consuming all threads on the OS).


kachayev:
Good catch @rafd, thanks! `def` also would not work for applications, as you need something like `components` to manage resources properly. Tho' I agree it's still less harmful compared to keeping connection pool within `let`.

kachayev:
There's another connection pool (in the same file) created the same way. Let me update the PR.

PR #541, opened 2020-05-21
Support -Dio.netty.eventLoopThreads for start-server

rymndhng:
Uses `get-default-event-loop-threads` which is the same logic as the clients.


kachayev:
It seems to me using `-D` configuration properties suites better applications and less so libraries. Moving away from implicit parameters in favor of explicit configuration was a part of Netty 5 roadmap as well. I see advantages of keeping this behavior consistent with client implementation, but we should consider explicit parameter to give flexibility to override JVM properties. Let me merge this into `1.0.0` branch (this change might effectively change the behavior of existing servers) and implement additional configuration key for the server.

kachayev:
As of now, merging into `1.0.0` branch

PR #547, opened 2020-08-16
remove Java exception classes

ertugrulcetin:


KingMob:
Hi @ertugrulcetin , and thanks for contributing to Aleph. Might I ask what the motivation here is? There's no description in the PR.

If the benefit is to gain typed ExceptionInfo exceptions, that's nice, but unfortunately, this will break any existing code that attempts to catch the parent class. Anything looking for a TimeoutException will break, and there's a lot of calls to `(d/catch' TimeoutException ...)` in Aleph that would have to be added to this PR. 

If these exceptions were used purely internally, that would be one thing, but since they bubble up to user code, this would be a breaking change for a lot of people.

ertugrulcetin:
Hi @KingMob, thanks for the response! To be honest, I forgot that I created this PR :) You're right, I agree with you. I don't think these changes not that required.

KingMob:
Sorry for the long delay. The original creator, Zach Tellman, moved on a couple years ago, and recently handed off the reins to the clj-commons community, so we're trying to go through and clean up old issues and PRs.

PR #549, opened 2020-09-10
Add ring middleware helpers

lambdam:


KingMob:
This partly exists already as `aleph.http/wrap-ring-async-handler`. 

For the rest of it, Aleph is built on Manifold streams and Netty channels (which are closer to interceptor chains), and not middleware. Best way to incorporate async ring middleware is to wrap your async ring handler, and then at the very end, take the resulting middleware+handler and use `wrap-ring-async-handler` on it.

Closing for now.

KingMob:
This partly exists already as `aleph.http/wrap-ring-async-handler`. 

For the rest of it, Aleph is built on Manifold streams and Netty channels (which are closer to interceptor chains), and not middleware. Best way to incorporate async ring middleware is to wrap your async ring handler, and then at the very end, take the resulting middleware+handler and use `wrap-ring-async-handler` on it.

Closing for now.

PR #553, opened 2020-10-06
Error handling, revisited

kachayev:
There are quite a few changes introduced here. Would be more than happy to get comments before merging into `1.0.0` branch 😀 

Notably:
* both servers and clients close connection (context, to be more precise) in case of exceptions
* `error-logger` parameter is added for HTTP client/server, Websocket client/server to customize exceptions logging for those exceptions that could not be visible to user's code in any other way (e.g. something that happens within HTTP server pipeline)
* `TimeoutException` handlers are fixed in HTTP client (to avoid double disposal of the connections)
* HTTP decoder errors are now properly handled to deal with bad traffic, error code statuses are fixed
* HTTP client propagates throwable to the caller whenever possible (rather than just relying on connection being closed later)
* `:aleph/interrupted?` flag is introduce as a part of ring request object for HTTP server, making it's possible to track if streaming body was closed because of an error (e.g. decoder exception, premature close, etc)
* a lot of unit tests to check failure scenarios

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -54,7 +54,8 @@
    | `idle-timeout` | when set, forces keep-alive connections to be closed after an idle time, in milliseconds
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
-   | `num-event-loop-threads` | optional, defaults to double number of available processors."
+   | `num-event-loop-threads` | optional, defaults to double number of available processors.
+   | `error-logger` | optional, function to be invoked on each exception propagated through the pipeline up to `request-handler`. Supposed to be used only for logging, crash reporting, metrics, etc rather than error recovery."
   [handler options]
   (when (contains? options :max-chunk-size)
     (log/warn "Ignoring :max-chunk-size option as it was deprecated"))
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -270,7 +270,8 @@
    | `max-frame-payload` | maximum allowable frame payload length, in bytes, defaults to `65536`.
    | `max-frame-size` | maximum aggregate message size, in bytes, defaults to `1048576`.
    | `allow-extensions?` | if true, allows extensions to the WebSocket protocol, defaults to `false`.
-   | `heartbeats` | optional configuration to send Ping frames to the client periodically (if the connection is idle), configuration keys are `:send-after-idle` (in milliseconds), `:payload` (optional, empty uses empty frame by default) and `:timeout` (optional, to close the connection if Pong is not received after specified timeout)."
+   | `heartbeats` | optional configuration to send Ping frames to the client periodically (if the connection is idle), configuration keys are `:send-after-idle` (in milliseconds), `:payload` (optional, empty uses empty frame by default) and `:timeout` (optional, to close the connection if Pong is not received after specified timeout).
+   | `error-logger` | optional, function to be invoked on each exception propagated through the pipeline up to `request-handler`. Supposed to be used only for logging, crash reporting, metrics, etc rather than error recovery."
   ([req]
     (websocket-connection req nil))
   ([req options]
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -354,19 +354,14 @@
 
                      (maybe-timeout! connection-timeout)
 
-                     ;; connection timeout triggered, dispose of the connetion
-                     (d/catch' TimeoutException
-                       (fn [^Throwable e]
-                         (flow/dispose pool k conn)
-                         (d/error-deferred (ConnectionTimeoutException. e))))
-
-                     ;; xxx(errors-handling): seems like in this case we process
-                     ;; exception twice
                      ;; connection failed, bail out
                      (d/catch'
-                       (fn [e]
+                       (fn [^Throwable e]
                          (flow/dispose pool k conn)
-                         (d/error-deferred e)))
+                         (d/error-deferred
+                           (if (instance? TimeoutException e)
+                             (ConnectionTimeoutException. e)
+                             e))))
 
                      ;; actually make the request now
                      (d/chain'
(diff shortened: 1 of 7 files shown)


src/aleph/http/client.clj
```diff
@@ -453,9 +482,12 @@
      decompress-body? false
      save-content-encoding? false}}]
   (fn [^ChannelPipeline pipeline]
-    (let [handler (if raw-stream?
-                    (raw-client-handler response-stream response-buffer-size)
-                    (client-handler response-stream response-buffer-size))
+    (let [error-logger' (or error-logger
+                            (fn [^Throwable ex]
+                              (log/warn "error in HTTP client" (.getMessage ex))))
```
gsnewmark, line 487: Do you ignore stacktrace for the exception on purpose? What I mean is that `(log/warn ex "error in HTTP client")` would print both the exception message and its trace, but your code will only print the message itself. It's easy to get it by using the custom logger, but I believe having traces by default will actually be useful for more people.

src/aleph/http/client.clj
```diff
@@ -110,34 +110,32 @@
...
-    ;; could happens when io.netty.handler.codec.http.HttpObjectAggregator
-    ;; is part of the pipeline
-    (instance? TooLongFrameException ex)
+(defn handle-exception [^ChannelHandlerContext ctx ^Throwable ex response-stream error-logger]
+  ;; Typical non-IO exception would be TooLongFrameException, SSLHandshakeError etc
+  (when-not (not (instance? IOException ex))
+    (try
+      (error-logger ex)
+      (catch Throwable e
+        (log/warn "error in error logger" e)))
```
gsnewmark, line 119: Do you ignore stacktrace for the exception on purpose? What I mean is that `(log/warn e "error in error logger")` would print both the exception message and its trace, but your code will only print the message itself.

src/aleph/http/client.clj
```diff
@@ -725,7 +776,10 @@
        :exception-caught
        ([_ ctx ex]
         (when-not (d/error! d ex)
-          (log/warn ex "error in websocket client"))
+          (try
+            (error-logger ex)
+            (catch Throwable e
+              (log/warn "error in error logger" e))))
```
gsnewmark, line 782: Do you ignore stacktrace for the exception on purpose? What I mean is that `(log/warn e "error in error logger")` would print both the exception message and its trace, but your code will only print the message itself.

src/aleph/http/core.clj
```diff
@@ -480,112 +514,148 @@
...
+                 (netty/close ch))))
+            (d/catch' (fn [_]))))
+
+      handle-exception
+      (fn [error-logger body e keep-alive? ^Channel ch]
+        (if (nil? error-logger)
+          (let [class-name (.getName (class body))]
+            (log/errorf "error sending body of type %s: %s"
+                        class-name
+                        (.getMessage ^Throwable e)))
```
gsnewmark, line 595: Do you ignore the stacktrace for this exception on purpose?

gsnewmark: ok, the exception is rethrowed later in the handler, so it makes sense in this particular case.

src/aleph/http/server.clj
```diff
@@ -196,42 +199,73 @@
...
-    netty/wrap-future
-    (fn [_] (netty/close ctx))))
+                      (invalid-value-response req rsp error-logger))
+                    error-logger))))))))))
+
+(defn handle-exception [error-logger ctx ex]
+  (try
+    (error-logger ex)
+    (catch Throwable e
+      (log/warn "exception in error logger" e)))
```
gsnewmark, line 209: Do you ignore the stacktrace for this exception on purpose? What I mean is that `(log/warn e "error in error logger")` would print both the exception message and its trace, but your code will only print the message itself.

src/aleph/http/server.clj
```diff
@@ -634,7 +709,12 @@
          :exception-caught
          ([_ ctx ex]
           (when-not (instance? IOException ex)
-            (log/warn ex "error in websocket handler"))
+            (if (some? error-logger)
+              (try
+                (error-logger ex)
+                (catch Throwable e
+                  (log/warn "exception in error logger" e)))
```
gsnewmark, line 716: Do you ignore the stacktrace for this exception on purpose? What I mean is that `(log/warn e "error in error logger")` would print both the exception message and its trace, but your code will only print the message itself.

gsnewmark (commented):
Thanks for still working on Aleph and keeping it in a good shape :tada: I've left a handful of comments (mostly about the logging :sweat_smile: ), but only briefly skimmed the server changes (particularly the new `:aleph/interrupted?` flag): in case nobody else will be able to look into it, I can try to do it later this week.

kachayev:
@gsnewmark Thanks a lot for the review! Let me address obviously wrong things and get back with more comments 😀

src/aleph/http/client.clj
```diff
@@ -110,34 +110,32 @@
...
-;; by the user and putting a new one would simply mean we're overriding
-;; response to the next request over the same connection)
-(defn exception-handler [ctx ex response-stream]
-  (cond
-    ;; could happens when io.netty.handler.codec.http.HttpObjectAggregator
-    ;; is part of the pipeline
-    (instance? TooLongFrameException ex)
+(defn handle-exception [^ChannelHandlerContext ctx ^Throwable ex response-stream error-logger]
+  ;; Typical non-IO exception would be TooLongFrameException, SSLHandshakeError etc
+  (when-not (not (instance? IOException ex))
```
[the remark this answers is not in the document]
kachayev: Oh, that's actually completely wrong 🤦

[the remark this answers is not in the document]
kachayev: Fixing this. We close connection on non-IO exception, I/O exception in the pipeline would stand for the fact that the channel was closed previously.

src/aleph/http/client.clj
kachayev: I honestly don't know if the entire trace is necessary here. If it feels like we need to log it when logging fails... I would be totally down to switch implementation.

src/aleph/http.clj
```diff
@@ -54,7 +54,8 @@
    | `idle-timeout` | when set, forces keep-alive connections to be closed after an idle time, in milliseconds
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
-   | `num-event-loop-threads` | optional, defaults to double number of available processors."
+   | `num-event-loop-threads` | optional, defaults to double number of available processors.
+   | `error-logger` | optional, function to be invoked on each exception propagated through the pipeline up to `request-handler`. Supposed to be used only for logging, crash reporting, metrics, etc rather than error recovery."
```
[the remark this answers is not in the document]
kachayev: Monitoring was exactly my first thought. But... more I think about how this would be useful in production more doubt I have about it. I can definitely see logging or any kind of error reporting system in here, but it seems unlikely someone needs a metric around something as unpredictable as "uncaught by any other means exceptions".

arnaudgeiser:
Quick update

I spent a bit of time already trying to rebase this branch on master.
It's not that easy but the number of tests that have been added are really valuable.

arnaudgeiser:
Superseded by #658 

Not all the changes from this PR have been included as some were significant in terms of impact.
However, the unit tests have been included and some fixes have been made.

PR #562, opened 2021-01-13
Fix websocket-upgrade-request? for Firefox WebSocket upgrade requests.

### LINKED ISSUE #561: websocket-upgrade-request? doesn't detect connection upgrade when there are other tokens
Firefox sometimes (always?) sends `keep-alive` as well as `Upgrade` in the `Connection` request header when it tries a WebSocket upgrade request, so `websocket-upgrade-request?` as written does not return true for Firefox WebSocket upgrade requests.

xorgy:
Fixes #561

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -708,7 +708,10 @@
   (let [headers (:headers req)
         conn (get headers :connection)
         upgrade (get headers :upgrade)]
-    (and (= "upgrade" (when (some? conn) (str/lower-case conn)))
+    (and (contains? (when (some? conn)
+                      (set (map str/trim
+                                (-> (str/lower-case conn) (str/split #",")))))
+                    "upgrade")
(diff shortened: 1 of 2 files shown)


slipset:
Thanks a lot xorgy. Could you please also provide a test which shows that this doesn't break existing functionality, ie that it works when only sending "Upgrade" in the connection header, as well as when sending "keep-alive,Upgrade"?

xorgy:
> Thanks a lot xorgy. Could you please also provide a test which shows that this doesn't break existing functionality, ie that it works when only sending "Upgrade" in the connection header, as well as when sending "keep-alive,Upgrade"?

Sure, I'll write a test for it.

xorgy:
Let me know if the test I've added is satisfactory, I can squash it in.

slipset (approved):
LGTM

No need to squash, I'll do a squash merge here. Thanks!

xorgy:
Thank you as well.

PR #564, opened 2021-03-04
[dns] Extract dns-resolver-group builder code into a separate function

alexander-yakushev:
Motivation: in certain cases, I want to use RoundRobinDnsAddressResolverGroup for DNS resolution. Some services provide a way of load balancing in the form of multiple DNS entries per URL, e.g. that's how AWS Load Balancer "balances" its own instances. The default DnsAddressResolverGroup caches just the first entry of the DNS server response and uses it throughout the cache invalidation period, leaving other returned entries unused. RR resolver caches the entire set of returned entries and goes over them on each request to the resolver.

In the first cut, I implemented the customizability via reflection which may seem clunky. The only other option I have is to pass a "constructor function" that accepts the Builder object and returns the resolver group. Please, tell me if that method would be preferred.

CC @kachayev.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -896,35 +896,33 @@ initialize an DnsAddressResolverGroup instance.
   (let [^Class
         channel-type (if (and epoll? (epoll-available?))
                        EpollDatagramChannel
-                       NioDatagramChannel)
+                       NioDatagramChannel)]
+    (cond-> (doto (DnsNameResolverBuilder.)
+              (.channelType channel-type)
+              (.maxPayloadSize max-payload-size)
+              (.maxQueriesPerResolve max-queries-per-resolve)
+              (.queryTimeoutMillis query-timeout)
+              (.ttl min-ttl max-ttl)
+              (.traceEnabled trace-enabled?)
+              (.optResourceEnabled opt-resources-enabled?)
+              (.ndots ndots)
+              (.decodeIdn decode-idn?)
+              (.recursionDesired recursion-desired?))
 
-        b (cond-> (doto (DnsNameResolverBuilder.)
-                    (.channelType channel-type)
-                    (.maxPayloadSize max-payload-size)
-                    (.maxQueriesPerResolve max-queries-per-resolve)
-                    (.queryTimeoutMillis query-timeout)
-                    (.ttl min-ttl max-ttl)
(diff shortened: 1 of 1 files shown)


kachayev:
@alexander-yakushev Hey, thanks a lot for the PR! As of now, custom `dns-resolver-group` could be provided directly as a configuration parameter into `connection-options` (as a `:name-resolver`). So, technically it's possible to create any resolver that suits best. I assume, that the use case here is more of "how to reuse fancy logic from the helper but only change one specific parameter?", right? Let me think a little bit about this.

alexander-yakushev:
I wasn't aware that the resolver object can be directly provided! Should have checked that before :). But yes, the point still stands whether it is worth trying to extend the Clojury helper further.

kachayev:
> I wasn't aware that the resolver object can be directly provided

Let's be honest, the documentation for custom DNS is barely existing :)

Here's what we can do: split `dns-resolver-group` into 2 functions by separating `dns-resolver-group-builder`. So, it would be something like

```
(def dns-resolver-group [dns-options]
  (DnsAddressResolverGroup. (dns-resolver-group-buider b))
```

In you case you would just call `dns-resolver-group-buider`, re-attach whatever property you need utilizing Java object setters, and than pass `:name-resolver` built from it.

What do you think?

alexander-yakushev:
That is a reasonable solution that allows reusing the builder code without going crazy on customizing the group class. Thanks for the suggestion! I've updated the PR accordingly.

src/aleph/netty.clj
```diff
@@ -923,7 +924,13 @@
               (not (empty? name-servers)))
             (.nameServerProvider ^DnsServerAddressStreamProvider
               (dns-name-servers-provider name-servers)))]
-    (DnsAddressResolverGroup. b)))
+    b))
```
kachayev, line 927: nitpick: we could avoid creating temp variable `b` within `let` just to return it immediately

alexander-yakushev: I wanted a slim diff and thought some future methods of the builder might not flow with `->` (however that's unlikely). But I'm OK to change it.

kachayev:
@alexander-yakushev Cool, thanks!

alexander-yakushev:
Thanks for handling this!

PR #579, opened 2022-01-10
Remove test dependency on /usr/share/dict/words

DerGuteMoritz:
This file is not present by default on many systems and in fact, on
systems like NixOS, it's not trivial to install at that path. Also,
there is an unchecked reliance on that file to be at least 1e7 bytes
long. If it is in fact shorter, the tests relying on it will just
silently run with shorter inputs.

Instead, just randomly generate printable ASCII characters. This seems
to be of comparable performance and doesn't have the above problems.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,6 @@ jobs:
 
     steps:
       - checkout
-      - run: sudo apt-get install wbritish
       # Download and cache dependencies
       - restore_cache:
           keys:
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -234,7 +234,9 @@
             @(http-get (str "https://localhost:" port "/" path)))))
        (str path "path failed")))))
 
-(def words (slurp "/usr/share/dict/words"))
+(def characters
+  (let [charset (conj (mapv char (range 32 127)) \newline)]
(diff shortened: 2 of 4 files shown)


KingMob:
@DerGuteMoritz I like the idea of this, but is your new version equivalent in behavior? `slurp` returns a single string, but yours returns an infinite seq of chars, iiuc.

KingMob:
I took a closer look, and it appears `words` being a string isn't an issue, since it's always treated as a seq of chars.

However, the char range you use excludes newlines, which means `test-line-echo` isn't being exercised properly any more. Can you add in newlines to the infinite sequence? And can you rename it from `words` to `characters`, since they're no longer words?

I think that's all that needs to change to make it merge-ready.

DerGuteMoritz:
> I took a closer look, and it appears `words` no longer being a string isn't an issue, since it's always treated as a seq of chars.

Indeed :+1:

> However, the char range you use excludes newlines, which means `test-line-echo` isn't being exercised properly any more.

Ouch, very good catch!

> Can you add in newlines to the infinite sequence? And can you rename it from `words` to `characters`, since they're no longer words?

Both done! Also made the character range a bit more explicit so that it's easier to follow.

KingMob (approved):
LGTM

Issue #193, opened 2015-11-28
http client sometimes hangs when receiving a body in excess of response-buffer-size


pbwolf:
The enclosed small web client uses aleph.http ("0.4.1-beta3") to retrieve a byte array with `aleph.http/get`.  Sometimes, the web response (the Deferred) never materializes!  The program sits, seemingly forever, after the "Sending..." message and before the "Received".

I have only seen the problem arise when the byte array would be bigger than response-buffer-size and the client and server are running on the same computer.  Even then, it is hit-or-miss.

Here is the web client program:

```
(ns giblet.podunk
  (:require [aleph.http :as http]))
(defn -main
  [url]
  (let [pool (http/connection-pool
              {:connections-per-host 1
               :connection-options {:keep-alive? true}})]
      (dotimes [n 42000000]
        (println "Sending request" n)
        (let [hr (deref
                  (http/get url
                            {:pool pool
                             :as :byte-array
                             :query-params {"n" (str n)}}))]
          (println "Received" n (alength (:body hr)))))))
```

Adding the connection-option `{:response-buffer-size bigger-number}`, to accommodate the biggest-imaginable response body, seems to work around the problem.  But doing so probably shouldn't be necessary, since that option is documented as "This does *not* represent the maximum size response that the client can handle (which is unbounded), and is only a means of maximizing performance".

ztellman:
If you don't consume the body, the response won't be considered complete, and subsequent requests won't be able to be made.  By increasing the size of the response buffer, you're working around this by making the consumption implicit, but I would consider this behavior "expected".  I'm going to mark this closed, but let me know if you think this is incorrect.

pbwolf:
Aleph dispenses the body as a byte array, so I think it's consumed.  Moreover, when the program runs distantly from the server, it does not seem to run into the problem.

ztellman:
Oh, I missed the `:as` clause.  This may be an issue where the response comes fast enough that the `bs/to-byte-array` call occurs on the Netty thread, rather than the response handling pool, which is meant to prevent these sorts of deadlocks.

Reopening, I'll take a closer look.  Sorry for the confusion on my end.

ztellman:
[...]

PR #583, opened 2022-03-19
Fix all linter warnings and errors in test suite

DerGuteMoritz:
Decided to start small since I already encountered a few issues that I wanted to coordinate on before continuing. I will add comments on the diff to discuss these!

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,1 +1,5 @@
-{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]}
+{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]
+ :linters      {:unsorted-required-namespaces {:level :warning}}
+ :lint-as      {dynamic-redef.core/with-dynamic-redefs clojure.core/with-redefs
+                manifold.deferred/let-flow clojure.core/let
+                manifold.deferred/loop clojure.core/let}}
--- a/test/aleph/classloader_test.clj
+++ b/test/aleph/classloader_test.clj
@@ -1,14 +1,16 @@
 (ns aleph.classloader-test
-  (:require  [clojure.test       :refer [deftest testing is]]
-             [aleph.http         :as http]
-             [aleph.netty        :as netty]
-             [manifold.deferred  :as d]
-             [manifold.utils     :refer [when-class]]
-             [signal.handler     :refer [on-signal]]
-             [dynamic-redef.core :refer [with-dynamic-redefs]])
-  (:import [io.netty.util.concurrent Future]
-           [java.lang.management ManagementFactory]
-           [java.util.concurrent CompletableFuture]))
+  (:require
+   [aleph.http :as http]
+   [aleph.netty :as netty]
+   [clojure.test :refer [deftest is testing]]
+   [dynamic-redef.core :refer [with-dynamic-redefs]]
+   [manifold.deferred :as d]
+   [manifold.utils :refer [when-class]]
+   [signal.handler :refer [on-signal]])
+  (:import
+   (io.netty.util.concurrent Future)
+   (java.lang.management ManagementFactory)
+   (java.util.concurrent CompletableFuture)))
 
 (defn- operation-complete
   "Stubs for `GenericFutureListener/operationComplete` which
--- a/test/aleph/http/core_test.clj
+++ b/test/aleph/http/core_test.clj
@@ -1,11 +1,9 @@
 (ns aleph.http.core-test
-  (:use
-    [clojure test])
   (:require
-    [aleph.http.core :as core])
+   [aleph.http.core :as core]
+   [clojure.test :refer [deftest is]])
   (:import
-    [io.netty.handler.codec.http
-     DefaultHttpRequest]))
+   (io.netty.handler.codec.http DefaultHttpRequest)))
 
 (deftest test-HeaderMap-keys
   (let [^DefaultHttpRequest req (core/ring-request->netty-request
--- a/test/aleph/http/multipart_test.clj
+++ b/test/aleph/http/multipart_test.clj
@@ -1,26 +1,23 @@
 (ns aleph.http.multipart-test
   (:require
-    [aleph.http :as http]
-    [aleph.http.core :as core]
-    [aleph.http.multipart :as mp]
-    [clj-commons.byte-streams :as bs]
-    [clojure.edn :as edn]
-    [clojure.test :refer [deftest testing is]]
-    [manifold.deferred :as d]
-    [manifold.stream :as s]
-    [clojure.string :as str])
+   [aleph.http :as http]
+   [aleph.http.core :as core]
+   [aleph.http.multipart :as mp]
+   [clj-commons.byte-streams :as bs]
+   [clojure.edn :as edn]
+   [clojure.string :as str]
+   [clojure.test :refer [deftest is testing]]
+   [manifold.deferred :as d]
+   [manifold.stream :as s])
   (:import
-    (io.netty.buffer
-      ByteBufAllocator)
-    (io.netty.handler.codec.http
-      HttpContent)
-    (io.netty.handler.stream
-      ChunkedInput)
-    (java.io
-      File)))
+   (io.netty.buffer ByteBufAllocator)
+   (io.netty.handler.codec.http HttpContent)
+   (io.netty.handler.stream ChunkedInput)
+   (java.io File)))
 
 (def file-to-send (File. (str (System/getProperty "user.dir") "/test/file.txt")))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest test-multipart-builder
   (let [body (mp/encode-body [{:part-name "part1"
                                :content "content1"
@@ -58,12 +55,14 @@
     ;; filename header
     (is (.contains body-str "filename=\"content5.pdf\""))))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest test-custom-boundary
   (let [b (mp/boundary)
         body (mp/encode-body b [{:part-name "part1" :content "content1"}])
         body-str (bs/to-string body)]
     (is (.endsWith body-str (str b "--")))))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest test-base64-content-transfer-encoding
   (let [body (mp/encode-body [{:part-name "part1"
                                :content "content1"
@@ -72,6 +71,7 @@
     (is (.contains body-str "base64"))
     (is (.contains body-str "Y29udGVudDE="))))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest test-binary-content-transfer-encoding
   (testing "specify 'binary' in headers"
     (let [body (mp/encode-body [{:part-name "part1"
@@ -88,12 +88,14 @@
       (is (.contains body-str "content2"))
       (is (false? (.contains body-str "Content-Transfer-Encoding"))))))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest reject-unknown-transfer-encoding
   (is (thrown? IllegalArgumentException
       (mp/encode-body [{:part-name "part1"
                         :content "content1"
                         :transfer-encoding :uknown-transfer-encoding}]))))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
 (deftest test-content-as-file
   (let [parts [{:part-name "part1"
                 :content file-to-send}
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -1,36 +1,24 @@
 (ns aleph.http-test
   (:require
-    [clojure.java.io :as io]
-    [clojure.string :as str]
-    [clojure.test :refer [deftest testing is]]
-    [aleph
-     [http :as http]
-     [netty :as netty]
-     [flow :as flow]
-     [ssl :as ssl]
-     [tcp :as tcp]]
-    [clj-commons.byte-streams :as bs]
-    [manifold.deferred :as d]
-    [manifold.stream :as s]
-    [manifold.time :as t])
+   [aleph.flow :as flow]
+   [aleph.http :as http]
+   [aleph.netty :as netty]
+   [aleph.ssl :as ssl]
+   [aleph.tcp :as tcp]
+   [clj-commons.byte-streams :as bs]
+   [clojure.java.io :as io]
+   [clojure.string :as str]
+   [clojure.test :refer [deftest is testing]]
+   [manifold.deferred :as d]
+   [manifold.stream :as s]
+   [manifold.time :as t])
   (:import
-    (java.io
-      File)
-    (java.util.zip
-      GZIPInputStream
-      ZipException)
-    (java.util.concurrent
-      TimeoutException)
-    (aleph.utils
-      ConnectionTimeoutException
-      RequestTimeoutException)
-    (io.netty.channel
-      ChannelHandlerContext
-      ChannelOutboundHandlerAdapter
-      ChannelPipeline
-      ChannelPromise)
-    (io.netty.handler.codec.http
-      HttpMessage)))
+   (aleph.utils ConnectionTimeoutException RequestTimeoutException)
+   (io.netty.channel ChannelHandlerContext ChannelOutboundHandlerAdapter ChannelPipeline ChannelPromise)
+   (io.netty.handler.codec.http HttpMessage)
+   (java.io File)
+   (java.util.concurrent TimeoutException)
+   (java.util.zip GZIPInputStream ZipException)))
 
 ;;;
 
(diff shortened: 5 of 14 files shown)


.clj-kondo/config.edn
```diff
@@ -1 +1,3 @@
-{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]}
+{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]
+ :linters      {:unsorted-required-namespaces {:level :warning}}
```
DerGuteMoritz, line 2: This linter isn't enabled by default but I think it's a nice one to have. And since we're likely going to have to touch pretty much all `ns` forms anyway for cleanup ... but feel free to disagree, of course!

.clj-kondo/config.edn
```diff
@@ -1 +1,3 @@
...
+ :lint-as      {manifold.deferred/let-flow clojure.core/let}}
```
DerGuteMoritz, line 3: Ideally this would be provided by `manifold`, of course. There are a few more like this coming up. Right now I tend to collect all of them here and once everything is green, extract them out and file PRs against the other projects. Sound good?

test/aleph/http/core_test.clj
```diff
@@ -1,11 +1,9 @@
 (ns aleph.http.core-test
-  (:use
-    [clojure test])
   (:require
-    [aleph.http.core :as core])
+   [aleph.http.core :as core]
+   [clojure.test :refer [deftest is]])
```
DerGuteMoritz, line 4: As you can see, I changed the indentation of these forms. That's because I used https://github.com/clojure-emacs/refactor-nrepl's `clean-ns` refactoring which produces this indentation by default. I did a quick research on how to change this but couldn't come up with anything so far. Let me know if you mind, otherwise I'll make it work *somehow* :smile:

test/aleph/http/multipart_test.clj
```diff
@@ -1,20 +1,18 @@
...
-   [clojure.edn :as edn])
+   [manifold.stream :as s])
   (:import
-   [java.io
-    File]))
+   (java.io File)))
 
 (def file-to-send (File. (str (System/getProperty "user.dir") "/test/file.txt")))
 
+#_{:clj-kondo/ignore [:deprecated-var]}
```
DerGuteMoritz, line 15: Almost all tests in this namespace are tagged with this since they all test `mp/encode-body` rather than its apparent successor `mp/encode-request`. Maybe this is an oversight that should be corrected separately?

test/aleph/http_test.clj
```diff
@@ -420,14 +403,14 @@
 
 (deftest test-large-responses
   (with-handler basic-handler
-    (let [pool (http/connection-pool {:connection-options {:response-buffer-size 16}})]
-      (dotimes [i 1 #_1e6]
+    (let [_pool (http/connection-pool {:connection-options {:response-buffer-size 16}})]
```
DerGuteMoritz, line 406: At first I thought to remove the whole `let` form but then I figured that perhaps it's in here as a convenience for tweaking the test somehwo?

test/aleph/http/core_test.clj
DerGuteMoritz: FTR: I just figured out how to adapt to it! Turns out that this is actually not `refactor-nrepl`'s job bur rather `clojure-mode`'s. So yeah, let me know if you'd like to preserve this!

test/aleph/http_test.clj
```diff
@@ -102,14 +90,14 @@
...
+(defn hello-handler [_request]
   {:status 200
    :body "hello"})
 
-(defn big-handler [request]
+(defn big-handler [_request]
   {:status 200
    :body (->> (s/periodically 0.1 #(byte-array 1024))
-           (s/transform (take 1e3)))})
+              (s/transform (take (long 1e3))))})
```
DerGuteMoritz, line 100: Ah, right, thread macros are another instance of indentation divergence. I only changed it when I had to touch the form to fix any linter warnings, though. Note that the style I changed it to is the one recommended by [The Clojure Style Guide](https://guide.clojure.style/#threading-macros-alignment) and happens to be the default of [`clojure-mode`](https://github.com/clojure-emacs/clojure-mode). But now I know how to adjust this, so if you'd prefer, I'm happy to adapt.

.clj-kondo/config.edn
KingMob: Agreed. I prefer sorted namespaces, and they're especially cluttered in aleph/manifold.

Sorting should be safe, but in general, my one major concern is that we never take linter warnings as gospel, because there's some really funky stuff in there. Like, namespaces might be required for `extend-type` side effects, but never refer to any vars. And in one case I know of, Zach temporarily switched namespaces in the middle of a file, which plays havoc with modern editors/linters.

.clj-kondo/config.edn
KingMob: That's fine.

`let-flow'` should be linted the same way.

test/aleph/http/core_test.clj
KingMob: The new 1-space indentation is more standard, according to https://guide.clojure.style/#line-break-ns-declaration.

We should probably make a single reformat-only PR. Otherwise the indentation either never gets fixed, or it gets fixed bit-by-bit and clutters up PR diffs with formatting changes.

test/aleph/http/multipart_test.clj
KingMob: Hmmm. Let's leave this for now.

test/aleph/http_test.clj
KingMob: Base on git blame, this was a partial attempt at deciphering clj-commons/aleph#193. I don't think it works as intended, so maybe the whole thing can be commented-out for now.

test/aleph/http_test.clj
KingMob: Nono, I prefer the general style guide, thanks for this.

KingMob:
@DerGuteMoritz What's the status of this? I was looking forward to it

KingMob:
No worries, take your time.

KingMob:
@DerGuteMoritz Any updates?

.clj-kondo/config.edn
DerGuteMoritz: > Agreed. I prefer sorted namespaces, and they're especially cluttered in aleph/manifold.

Alright, thanks for the feedback!

> Sorting should be safe, but in general, my one major concern is that we never take linter warnings as gospel, because there's some really funky stuff in there.

Sure thing! I will judiciously apply `#_:clj-kondo/ignore` comments :smile: 

> Like, namespaces might be required for extend-type side effects, but never refer to any vars.

`clj-kondo` has a convention for dealing with this - only gotta be careful to apply it where needed :smile: Thanks for the headsup!

test/aleph/http/core_test.clj
DerGuteMoritz: Alright, 1 space it is. If you don't mind, I would already apply this for the test suite in this PR. The next one will then do the same for the main code.

test/aleph/http_test.clj
DerGuteMoritz: Thanks for checking! I dug some more and found that this `let` was likely just forgotten to be removed with c031a64cd5110efa5585c7700e8f7de5648ac429. I'll just do that in a separate commit for clarity!

test/aleph/classloader_test.clj
```diff
@@ -39,7 +41,7 @@
 (deftest test-classloader
   (testing "classloader: ensure the class loader is always a DynamicClassLoader"
     (let [result (CompletableFuture.)]
-      (with-dynamic-redefs [netty/operation-complete (partial operation-complete result)]
+      (with-dynamic-redefs [#_{:clj-kondo/ignore [:private-call]} netty/operation-complete (partial operation-complete result)]
```
DerGuteMoritz, line 44: @KingMob Had to add this one after the rebase - not sure if the var being private is intended here?

DerGuteMoritz:
@KingMob Sorry for the delay :bow:  Back on track now. I've rebased and adjusted for the latest `master` changes and incorporated your feedback from above (see threads there for some further details). I want to look into adding a circle-ci job for `clj-kondo` and then this could be merged!

DerGuteMoritz:
Alright, just added a `lint` job to the Circle CI config. Don't know how to test it, though - it doesn't seem to be picked up automatically for PR checks, at least. Hope you know what to do :pray:

test/aleph/classloader_test.clj
arnaudgeiser: That's not something we want to expose externally, Aleph is already exposing way to many public `vars` in my opinion.
Having this `:private-call` is fine for me.

arnaudgeiser (approved):
Fine for me,
A follow-up for this PR would be to format using `cljfmt`.
Using any `parinfer/paredit` is really complicated on the state of the codebase.

Regarding the Github action, I suppose @KingMob will fix it afterwards if needed. :)

test/aleph/classloader_test.clj
KingMob: Can that be set at another level? I’d prefer not to have kondo annotations deep in code if possible.

KingMob:
I need to check with borkdude on using kondo for circleci; I think Eastwood is preferred for non-interactive CI linting because it more closely mimics how the real compiler sees code. Let me get back to you on that.

KingMob:
> Alright, just added a `lint` job to the Circle CI config. Don't know how to test it, though - it doesn't seem to be picked up automatically for PR checks, at least. Hope you know what to do 🙏

@DerGuteMoritz I think the reason it's not picked up is the indentation level makes it a separate job, but we probably want to place it under the existing `build` job as another step (preferably right before the testing step). In general, clj-commons has used just two circleci jobs, one for building and one for deploying (deployment jobs don't yet exist for aleph and manifold for historical reasons, since their deployment isn't yet under the clj-commons maven group).

KingMob:
I see plenty of people using kondo for CI, so that's fine. It might be wiser to make a separate ci config file, as described in https://github.com/clj-kondo/clj-kondo/blob/master/doc/ci-integration.md

The goal should be to get it in place now, with nothing blocking merges. Aleph and Manifold do some funky stuff sometimes, and kondo should not block current development just because of that. As we modernize the codebase, we can revisit what warnings we want to turn back on.

DerGuteMoritz:
> A follow-up for this PR would be to format using `cljfmt`.
> Using any `parinfer/paredit` is really complicated on the state of the codebase.

Indeed, the formatting is quite idiosyncratic in places, mainly due to all these macros I think. Running `cljfmt` on it sounds like a nice follow-up indeed. I'll see if I can tackle that, too, once we've got the linting in :smile_cat: 


> @DerGuteMoritz I think the reason it's not picked up is the indentation level makes it a separate job, but we probably want to place it under the existing `build` job as another step (preferably right before the testing step).

I broke it out into a separate job for two reasons:

1. This way, the lint job will show up as a separate check on a PR, giving immediate indication of what kind of failure there is.
2. A different Docker image is required for running `clj-kondo`. AFAIK there is no way to combine two existing images into one so to achieve that one would have to either build a custom image or install `clj-kondo` from within the job which is pretty messy.

> In general, clj-commons has used just two circleci jobs, one for building and one for deploying (deployment jobs don't yet exist for aleph and manifold for historical reasons, since their deployment isn't yet under the clj-commons maven group).

And is there a technical rationale for doing this or just tradition / convention? :smile: 


> I see plenty of people using kondo for CI, so that's fine. 

Yup, the way it operates is pretty much intended for non-interactive use. Also, without an automated job like this, it's pretty much guaranteed that the code will evolve into a non-linted state pretty quickly unless reviewers diligently run the linter before merging. And since that's so trivially automated ... :smile:

> It might be wiser to make a separate ci config file, as described in https://github.com/clj-kondo/clj-kondo/blob/master/doc/ci-integration.md

This would only be needed if we want CI-specific config which currently doesn't seem necessary nor particularly desirable - why would you want to have different linting rules for development vs. CI? :thinking: 

That being said, I'm happy to adjust any of the above to your liking, just gimme the word :+1:

DerGuteMoritz:
Sorry, missed on bit:

> The goal should be to get it in place now, with nothing blocking merges. Aleph and Manifold do some funky stuff sometimes, and kondo should not block current development just because of that. As we modernize the codebase, we can revisit what warnings we want to turn back on.

If by that you mean that lint failures should not block PRs from getting merged, that would be another case in point for having it be a separate CI job so that it can be marked as non-blocking.

KingMob:
We can try adding automatic cljfmt, see how it goes. In my limited experience with it, it requires a bunch of config for good results, especially around macros.

I'm fine with treating linting as a separate job, as long as we can ensure dependencies. E.g., if either linting or building/testing fail, a deploy job should not be runnable (see other clj-commons repos for examples). Regardless, people should always be able to click into CircleCI for details, so making it directly visible in GH isn't a big priority to me.

> And is there a technical rationale for doing this or just tradition / convention?

I don't know if there's a good reason the build job can't be split up, but I do know the other clj-commons deployment jobs run on different git tags, and the branch/tag criteria is set at the job level. So...maybe?

> A different Docker image is required for running clj-kondo. 

I didn't notice you used a different docker image. Potentially dumb question: does this work correctly in circleci? I know in other clj-commons Circle setups, we download a Babashka script for deployment. We could do something similar to download kondo if necessary. (We should also upgrade to the [newer circleci clojure img](https://circleci.com/developer/images/image/cimg/clojure)...)

> If by that you mean that lint failures should not block PRs from getting merged, that would be another case in point for having it be a separate CI job so that it can be marked as non-blocking.

But some rules SHOULD be hard failures (e.g., arity errors, unresolved symbols), just not necessarily all. Here's an example of why we might want separate CI and dev configs: https://github.com/clj-commons/manifold/blob/dbdd600c5a48df314e3ad06fe2986e05e5af21c1/src/manifold/time.clj#L126-L128. Zach does some crazy ns stuff here. At the developer level, I want to be reminded to fix this, but I don't want to have this block CI, because it may take some time to fix. 

Validating correctness is more important, but as you point out, maintaining good style requires regular linting, too. We could make all the changes at once and then enable them as blockers in the same PR, but I'm afraid this is turning out to be a lot of work already. I prefer smaller, more focused PRs; they're easier to finish, easier to review, and less risky to merge. 

If you don't want to have separate dev/CI configs (which is fair), let's still try to split up this work into the following PRs: 

1. A kondo and kondo-based fix PR, no circleCI
2. A cljfmt merge that does nothing but update the indentation based on the Clojure style guide and specify the indentation rules
3. A circleCI PR that adds cljfmt and kondo jobs or steps

DerGuteMoritz:
> We can try adding automatic cljfmt, see how it goes. In my limited experience with it, it requires a bunch of config for good results, especially around macros.

Yeah, this will likely be some work - same with `clj-kondo` actually :grimacing:

> I'm fine with treating linting as a separate job, as long as we can ensure dependencies. E.g., if either linting or building/testing fail, a deploy job should not be runnable (see other clj-commons repos for examples).

Ah, right, looks like we would have to set up a workflow for this, similar to what's done e.g. here: https://github.com/clj-commons/byte-streams/blob/master/.circleci/config.yml#L7-L22 - looks like it should be possible to just add the lint job as another `requires` dependency to the deploy job.

> I didn't notice you used a different docker image. Potentially dumb question: does this work correctly in circleci? 

Fairly sure! [Docs say this about `image`](https://circleci.com/docs/configuration-reference#docker): "The name of a custom docker image to use. The first image listed under a job defines the job’s own primary container image where all steps will run."


> I know in other clj-commons Circle setups, we download a Babashka script for deployment. We could do something similar to download kondo if necessary.

Right, that's the approach I referred to as "messy" above :smile:  

> But some rules SHOULD be hard failures (e.g., arity errors, unresolved symbols), just not necessarily all. Here's an example of why we might want separate CI and dev configs: https://github.com/clj-commons/manifold/blob/dbdd600c5a48df314e3ad06fe2986e05e5af21c1/src/manifold/time.clj#L126-L128. Zach does some crazy ns stuff here. At the developer level, I want to be reminded to fix this, but I don't want to have this block CI, because it may take some time to fix.

OK makes sense! But I think I would rather have the default developer config be the same as the one used by CI so that one doesn't get all kinds of warnings during development which end up being irrelevant. So far I didn't have to disable any linting rules globally but once we do, we could provide a separate config which one could use instead for "full" linting when desired.


> We could make all the changes at once and then enable them as blockers in the same PR, but I'm afraid this is turning out to be a lot of work already. I prefer smaller, more focused PRs; they're easier to finish, easier to review, and less risky to merge.

Aye, agree with you here. That's why I decided to make this first PR about the test suite only so that we can settle these discussions before having invested a ton of work on the main code base :sweat_smile: 

> If you don't want to have separate dev/CI configs (which is fair), let's still try to split up this work into the following PRs:
> 
> 1. A kondo and kondo-based fix PR, no circleCI
> 2. A cljfmt merge that does nothing but update the indentation based on the Clojure style guide and specify the indentation rules
> 3. A circleCI PR that adds cljfmt and kondo jobs or steps

Sounds like a plan :+1: I will drop the CI job from this PR then. Shall we go ahead with merging it as-is, i.e. test suite only or would you prefer a big one for the whole code base now that we've mostly settled the nitty-gritty? FWIW I think I would prefer to merge this one as-is because it already provides some value and it will probably take a while for me to fix up the main code.

.clj-kondo/config.edn
```diff
@@ -1 +1,5 @@
-{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]}
+{:config-paths ["../resources/clj-kondo.exports/clj-commons/aleph"]
+ :linters      {:unsorted-required-namespaces {:level :warning}}
+ :lint-as      {dynamic-redef.core/with-dynamic-redefs clojure.core/with-redefs
+                manifold.deferred/let-flow clojure.core/let
+                manifold.deferred/loop clojure.core/let}}
```
KingMob, line 5: Minor: this is fine for now, but there's a way to export kondo linting notes as part of a lib, so we should really add that to manifold instead. That way everything using Manifold will get the correct linting.

test/aleph/classloader_test.clj
KingMob: @DerGuteMoritz Was there a way to set this elsewhere?

KingMob (approved):
Approved, but could you look at my last few comments before we merge?

.clj-kondo/config.edn
DerGuteMoritz: Yep, already on my radar: https://github.com/clj-commons/aleph/pull/583#discussion_r830513450 :+1:

test/aleph/classloader_test.clj
DerGuteMoritz: Even better: Looks like adding https://github.com/clj-commons/aleph/pull/583/files#diff-d8bfac7f9aa64d7a63a2c6810095e91e716527266288a210b3f31a2c141ce3f2R3 made it completely unnecessary :smile:

DerGuteMoritz:
@KingMob There you go!

KingMob (approved):

PR #584, opened 2022-03-24
Add server option for user defined header key normalization function

cnautze:
### Description

Since some of our response header keys are case sensitive we would like to handle these differently than the vanilla capitalization currently used. The suggested change adds a server option `:header-key-transform`  to pass in a function that is used for the non-standard header keys. This function expects a header key string and returns a normalized string. Looking forward to your feedback.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -51,7 +51,9 @@
    | `compression-level` | optional compression level, `1` yields the fastest compression and `9` yields the best compression, defaults to `6`. When set, enables http content compression regardless of the `compression?` flag value
    | `idle-timeout` | when set, forces keep-alive connections to be closed after an idle time, in milliseconds
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
-   | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread."
+   | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
+   | `header-key-transform` | optional function that takes a String representation of a header key and returns a
+   normalized version used in the server response."
   [handler options]
   (server/start-server handler options))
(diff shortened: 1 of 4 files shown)


KingMob:
Well, with the caveat that [HTTP headers are defined to be case-insensitive](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2), I have no fundamental problem incorporating this functionality.

KingMob:
Hmm, upon further thought, I think I have a different solution.

First, this need is really niche. HTTP/1.1 is supposed to be case-insensitive, and most things support that. HTTP/2 [requires lowercase headers](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2), so it should be even less of a concern going forward.

Many of the intermediate Aleph fns have too many params already, but unfortunately, they're public, so we can't just collect them into maps without breakage, or I'd try that route to support you. (Though I am considering 2-arity versions that would do that at the cost of speed.)

Instead, I'd like to suggest you use the `pipeline-transform` parameter to insert a new channel handler in the Netty pipeline that alters headers after the main handler is done, but before anything gets written to the wire. It would look something like this:

```clojure
(http/start-server
    your-main-handler
    {;; other parameters go here...
     :pipeline-transform
     (fn [^io.netty.channel.ChannelPipeline pipeline]
       ;; yes it's .addBefore, on the way out, Netty handlers go in reverse, sort of like interceptors
       (.addBefore pipeline
                   "request-handler"                        ; the Aleph name for the main handler
                   "server-header-transformer"              ; the name of your handler
                   (proxy [io.netty.channel.ChannelOutboundHandlerAdapter] []
                     (write [^io.netty.channel.ChannelHandlerContext ctx
                             ^Object msg
                             ^io.netty.channel.ChannelPromise p]
                       (when (instance? io.netty.handler.codec.http.HttpMessage msg)
                         (let [^io.netty.handler.codec.http.HttpMessage http-msg msg
                               headers (.headers http-msg)
                               server-val (.get headers "Server")]
                           (.set headers "SERVER" server-val))) ; make the "server" header uppercase
                       (.write ctx msg p)))))})
```

NB: if you're testing this with the Aleph client, you should know it _also_ lower-cases headers it receives, via the NettyResponse and HeaderMap classes, so it may be hard to see the results if you're transforming another way. But I verified this technique works via Wireshark.

cnautze:
Thank you for your feedback! I understand your objections. After an internal discussion we decided to go forward with our current solution (not the suggested changes) which is perhaps not the most beautiful but works for us.

KingMob:
@cnautze I think the problem is, there's _no_ good way to handle this, given the way Aleph currently works. Please don't hesitate to contribute again, though.

PR #585, opened 2022-03-26
Better handling of REQUEST_URI_TOO_LONG/REQUEST_HEADER_FIELDS_TOO_LARGE

arnaudgeiser:
## Description

In case the request is rejected (because the decoder failed), it always ends up on `414 - URI Too Long`
while it could be for other reasons.
Taking the same approach as @kachayev closed PR #440 and how this behaviour is currently implemented on another
Netty based library (Vertx) [1], I propose the following changes on Aleph:

When the Exception is of type `TooLongFrameException` and contains:
- An HTTP line is larger than => `414 - URI Too Long`
- HTTP header is larger than => `431 - Request Header Fields Too Large`

In any other case => `400 - Bad Request`

[1] https://github.com/eclipse-vertx/vert.x/blob/de2edcdead9e91f6e47970449bde4f0cf7254a8b/src/main/java/io/vertx/core/http/HttpServerRequest.java#L63-L79

## Impacts

Users relying on the constant `414` status code will be impacted.
However, I make the assumption that no user consider `414` as a good default
when the decoder failed for other reasons than the URI being too long. `400` would have been more
correct in any case.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -7,7 +7,7 @@
     [clojure.tools.logging :as log]
     [clojure.string :as str]
     [manifold.deferred :as d]
-    [manifold.stream :as s])
+    [manifold.stream :as s]) 
   (:import
     [java.util
      EnumSet
@@ -217,7 +217,7 @@
 
 (defn- ^HttpResponseStatus cause->status [^Throwable cause]
   (if (instance? TooLongFrameException cause)
-    (let [message ^String (.getMessage cause)]
+    (let [message (.getMessage cause)]
       (cond
         (.startsWith message "An HTTP line is larger than")
         HttpResponseStatus/REQUEST_URI_TOO_LONG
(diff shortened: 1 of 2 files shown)


KingMob (approved):
@arnaudgeiser It looks pretty good.

I made a few minor changes. I thought I'd try out the maintainer PR editing capability, but that didn't work so well. I couldn't push local changes to the PR branch, so I had to do it via the GH web interface, which is pretty... meh. Perhaps because the PR was from your master branch? Dunno.

Anyway, thanks for contributing!

arnaudgeiser:
> I made a few minor changes. I thought I'd try out the maintainer PR editing capability, but that didn't work so well. I couldn't push local changes to the PR branch, so I had to do it via the GH web interface, which is pretty... meh. Perhaps because the PR was from your master branch? Dunno.

I didn't fork from `clj-commons/aleph` initially, that might explain why you wasn't able to edit my branch.
For the next PRs, I will fix this.

Thanks a lot for taking over Aleph!

Issue #575, opened 2021-11-20
Understanding of `max-queue-size` of connection-pool


FiV0:
When running 
```clj
(require '[aleph.http :as aleph]
         '[manifold.deferred :as d])

(dotimes [_ 10000]
  (d/on-realized (aleph/get "http://localhost:8080" {:throw-exceptions false})
                 (fn [r] (println "Status: " (:status r)))
                 (fn [e] (println "Error type: " (type e)))))
```
I am getting some `java.util.concurrent.RejectedExecutionException` exceptions not even in my handler but on `manifold.deferred` level.

My understanding from looking at the docs of [connection-pool](https://github.com/clj-commons/aleph/blob/83850fb50e8d160eb089a98a38bbe42c892fcb5c/src/aleph/http.clj#L94) and further [flow/instrumented-pool](https://github.com/clj-commons/aleph/blob/83850fb50e8d160eb089a98a38bbe42c892fcb5c/src/aleph/flow.clj#L28) was that one could do at least 65536 requests (even per domain) before this should happen. 

Is this assumption wrong? Is pending requests not to be equated with pending acquires?

KingMob:
@FiV0 There's definitely something going on, but I haven't figured out what yet. I'll take a closer look later when I get a chance. What's weird is the two major Manifold executor pools are utilization-executors, which default to a queue length of 0, and are supposed to be unbounded. 

I haven't examined that part of the code quite as closely since becoming maintainer, and I don't know Aleph super-well. PRs welcome!

FiV0:
@KingMob Ok. I might take a look when I find the time.

arnaudgeiser:
Hello @FiV0,

In your case, the `java.util.concurrent.RejectedExecutionException` is not raised by the `flow/instrumented-pool`.
As you mentioned, the default value is __65536__, so you are not running out-of-capacity here. [1]

Any `java.util.concurrent.Executor` can raise this Exception. [2]
In your case, it's raised by the (undocumented) `response-executor`, a `manifold.executor/utilization-executor` which defaults to 256 threads with a queue-length of 0. [3]

There is fortunately an escape hatch, you can pass your own Executor as an option of any HTTP request.

Here is a simple example where is fails:

```clojure
(require '[aleph.http :as aleph]
         '[aleph.netty :as netty]
         '[manifold.deferred :as d])

(let [executor (flow/utilization-executor 0.9 1
[...]

PR #587, opened 2022-03-26
Document response-executor for http/request

arnaudgeiser:
## Description

This PR is a follow-up to the #575 issue.
It documents the response-executor on `aleph.http/request` (which is actually a request-executor...).
I took the opportunity to remove `follow-redirects?` which was not used.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -280,6 +280,7 @@
      | `pool` | a custom connection pool
      | `middleware` | custom client middleware for the request
      | `pool-timeout` | timeout in milliseconds for the pool to generate a connection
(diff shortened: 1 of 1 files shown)


KingMob:
Thanks for tackling this, @arnaudgeiser. Off the top of my head this looks good, but it's late here, so I'm going to look at it tomorrow.

It should probably be renamed to `request-executor` to be less confusing. (It was originally copied from elsewhere and the name wasn't changed.) We can use an `or` to support both keys, but keep `response-executor` undocumented.

KingMob (approved):
Ehhh, on further thought, `response-executor` makes sense too, since it's crafting the _response_ to the request, so... 😄 

This is good, thanks!

Issue #559, opened 2020-12-30
HTTP request hangs when sending empty file


sergeyminaev:
```clojure
  @(aleph.http/put "https://localhost:3333"
                   {:body (clojure.java.io/file "empty.txt")
                    :pool (aleph.http/connection-pool {:connection-options {:ssl-context ctx}})
                    :request-timeout 1000})
```

This code throws TimeoutException because of converting empty file to Manifold source inside `aleph.http.core` fails with NullPointerException.

kachayev:
@sergeyminaev Thanks for reporting! 1.0.0 branch has a few improvements for errors handling, a few more are introduces in [this](https://github.com/clj-commons/aleph/pull/553) PR. I  believe that this specific issue were addressed there (and there's a unit test to cover it).

KingMob:
Fixed by #590

PR #589, opened 2022-03-28
client: Accept empty file and ensure response on error

arnaudgeiser:
## Description

In case of TLS usage (`ssl-context`) and an empty file, a `NullPointerException` was raised on the following code because `bs/to-byte-buffers` was returning `nil` and a `manifold.stream/->source` cannot be build from it.

```clojure
(send-streaming-body ch msg
      (-> file
        (bs/to-byte-buffers {:chunk-size (.-chunk-size file)})
        s/->source))
```

This issue has been reported here : https://github.com/clj-commons/aleph/issues/559

The following PR considers an empty file as a valid request when using TLS (it's already the case when TLS is not used).

Two other changes are included:
- We always log when there is a failure on `aleph.http.core/send-message`. [1]
- We always provide a response to the client when an Exception is raised - related to the issue [2]

Both of approaches have also been considered by @kachayev (#553) but I don't think it will be ever merged...

[1] https://github.com/clj-commons/aleph/pull/553/files#diff-f3c6a88143a34bab27cfd4dde73563c71de85a57da8133def509631b893d05ddR648
[2] https://github.com/clj-commons/aleph/pull/553/files#diff-2d6587452350a3603e05f940d19be2b98e6377dfc7efc3dc1eb099ff332e4b3bR654

### DIFF
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -1,12 +1,15 @@
 (ns aleph.http-test
   (:require
+    [clojure.java.io :as io]
     [clojure.string :as str]
     [clojure.test :refer [deftest testing is]]
     [aleph
      [http :as http]
      [netty :as netty]
      [flow :as flow]
+     [ssl :as ssl]
      [tcp :as tcp]]
+    [aleph.http.core :as core]
     [byte-streams :as bs]
     [manifold.deferred :as d]
     [manifold.stream :as s])
@@ -192,9 +195,12 @@
      (with-server (http/start-server ~handler {:port port :compression-level 3})
        ~@body)))
 
-(defmacro with-ssl-handler [handler & body]
-  `(with-server (http/start-server ~handler {:port port, :ssl-context (netty/self-signed-ssl-context)})
-     ~@body))
+(def default-ssl-options {:port port, :ssl-context (netty/self-signed-ssl-context)})
+
+(defmacro with-handler-options
+   [handler options & body]
+   `(with-server (http/start-server ~handler ~options)
+      ~@body))
 
 (defmacro with-raw-handler [handler & body]
   `(with-server (http/start-server ~handler {:port port, :raw-stream? true})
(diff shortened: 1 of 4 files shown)


test/aleph/http_test.clj
```diff
@@ -192,9 +195,12 @@
...
        ~@body)))
 
-(defmacro with-ssl-handler [handler & body]
-  `(with-server (http/start-server ~handler {:port port, :ssl-context (netty/self-signed-ssl-context)})
-     ~@body))
+(def default-ssl-options {:port port, :ssl-context (netty/self-signed-ssl-context)})
+
+(defmacro with-ssl-handler
+   [handler options & body]
+   `(with-server (http/start-server ~handler ~options)
```
KingMob, line 202: Shouldn't `with-ssl-handler` include the SSL options? Otherwise, it's the same as `with-server`

src/aleph/http/client.clj
```diff
@@ -554,8 +554,11 @@
                     ;; might be different in case we use :multipart
                     (reset! save-body body))
 
-                  (netty/safe-execute ch
-                    (http/send-message ch true ssl? req' body))))
+                  (d/catch'
+                   (netty/safe-execute ch
+                                       (http/send-message ch true ssl? req' body))
+                   (fn [^Throwable e]
+                     (s/put! responses (d/error-deferred e))))))
```
KingMob, line 561: Better to thread this with `->` so it more closely looks like standard try/catch

KingMob:
Thanks for working on this. Can you split it up into smaller pieces? It's easier to review when multiple concerns aren't bundled together. I'd prefer to keep the empty file fix and the additional error logging separate. 

In particular, there's a lot of old code and notes from Kachayev about errors, and I'd like to look them over.

For the empty file fix, can you add non-TLS tests as well?

test/aleph/http_test.clj
```diff
@@ -238,6 +245,28 @@
...
+            (bs/to-string
+             (:body @(http-put client-url
+                               {:body (io/file "test/file.txt")
+                                :pool client-pool})))))
+     (is (thrown-with-msg? Exception
+                             #"BOOM!"
+                             (with-redefs [core/send-file-body (fn [_ _ _ _] (throw (Exception. "BOOM!")))]
+                                    @(http-put client-url
+                                     {:body (io/file "test/file.txt")
+                                      :pool client-pool})))))))
```
KingMob, line 268: This test is too fragile for the value it provides, I think. I'd rather not incorporate it.

arnaudgeiser:
Thank you @KingMob for the review, I will split this PR in two and add a test regarding non-TLS one.
=> PR turned into draft

PR #590, opened 2022-03-29
client: Accept empty file

arnaudgeiser:
## Description

In case of TLS usage (`ssl-context`) and an empty file, a `NullPointerException` was raised on the following code because `bs/to-byte-buffers` was returning `nil` and a `manifold.stream/->source` cannot be build from it.

```clojure
(send-streaming-body ch msg
      (-> file
        (bs/to-byte-buffers {:chunk-size (.-chunk-size file)})
        s/->source))
```

This issue has been reported here : https://github.com/clj-commons/aleph/issues/559

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -459,13 +459,16 @@
     (netty/write ch fr)
     (netty/write-and-flush ch empty-last-content)))
 
+(defn- file->stream [^HttpFile file]
+  (-> file
+      (bs/to-byte-buffers {:chunk-size (.-chunk-size file)})
+      s/->source))
+
 (defn send-file-body [ch ssl? ^HttpMessage msg ^HttpFile file]
   (cond
     ssl?
-    (send-streaming-body ch msg
-      (-> file
-        (bs/to-byte-buffers {:chunk-size (.-chunk-size file)})
-        s/->source))
+    (let [body (when (pos-int? (.length file)) (file->stream file))]
+      (send-streaming-body ch msg body))
 
     (chunked-writer-enabled? ch)
     (send-chunked-file ch msg file)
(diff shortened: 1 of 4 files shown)


src/aleph/http/core.clj
```diff
@@ -459,13 +459,16 @@
...
+
 (defn send-file-body [ch ssl? ^HttpMessage msg ^HttpFile file]
   (cond
     ssl?
-    (send-streaming-body ch msg
-      (-> file
-        (bs/to-byte-buffers {:chunk-size (.-chunk-size file)})
-        s/->source))
+    (let [body (when (pos-int? (.length file)) (file->stream file))]
+      (send-streaming-body ch msg body))
```
KingMob, line 471: I'm trying to think of a way this all might still fail if body becomes nil. According to the [javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#length()), 0 could also be returned if  it was a device or pipe. I don't think we'd ever support that, though.

KingMob: Wait, if we pass nil into send-streaming-body, won't `(netty/to-byte-buf-stream body' 8192)` throw an exception?

arnaudgeiser: The `if-let` is misleading, we are reaching that point:

```clojure
(netty/write-and-flush ch empty-last-content)
```
https://github.com/arnaudgeiser/aleph/blob/559/src/aleph/http/core.clj#L344

arnaudgeiser: The aim of this PR is to ensure the client never hangs (waiting for the `TimeoutException`) when the file is empty.
The situation is a little bit better than previously, however, I don't think we are handling every edge case.

KingMob: Ahh, right, right. This code should really be broken up into smaller fns, so to make it easier to follow.

KingMob (approved):
OK, looks good

Issue #542, opened 2020-06-23
Can't catch IllegalArgumentException on http methods


metametadata:
```clj
(require '[aleph.http :as http])

(try
  @(http/put "https://example.com"
            {:body            123
             :read-timeout    1000
             :request-timeout 1000})

  (catch Exception e
    (prn :exception! e)))
```

## Actual

`IllegalArgumentException` is logged while returned deferred contains an error about timeout:

```
error sending body of type  java.lang.Long                                                                     aleph.http.core                      ERROR 23.06.20 21:08:45 [aleph-netty-client-event-pool-1-18] 
java.lang.IllegalArgumentException: Don't know how to convert class java.lang.Long into (stream-of io.netty.buffer.ByteBuf)
	at byte_streams$convert.invokeStatic(byte_streams.clj:196)
	at byte_streams$convert.invoke(byte_streams.clj:162)
	at aleph.netty$to_byte_buf_stream.invokeStatic(netty.clj:185)
	at aleph.netty$to_byte_buf_stream.invoke(netty.clj:184)
	at aleph.http.core$send_streaming_body.invokeStatic(core.clj:294)
	at aleph.http.core$send_streaming_body.invoke(core.clj:249)
	at aleph.http.core$eval35149$send_message__35156$fn__35157.invoke(core.clj:404)
	at aleph.http.core$eval35149$send_message__35156.invoke(core.clj:403)
	at aleph.http.client$http_connection$fn__35515$fn__35516$f__34548__auto____35517.invoke(client.clj:466)
	at aleph.http.client$http_connection$fn__35515$fn__35516$fn__35519.invoke(client.clj:465)
	at clojure.lang.AFn.run(AFn.java:22)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:465)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:884)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)
:exception! #error {
 :cause "timed out after 1000 milliseconds"
 :via
 [{:type aleph.utils.RequestTimeoutException
[...]

Issue #532, opened 2020-01-13
Aleph/manifold thread occasionally hangs when dereferencing http requests


sbocq:
We're are using `Aleph 0.4.6` and we ran several times into the issue where threads get stuck when dereferencing `http` requests.

Today, it occurred in these 2 functions:

```clojure
(defn intervention-details-select [{:keys [leo-url]} user intervention-id]
  (->>
   (-> @(http/get (str leo-url "/sqa/interventions/"
                       intervention-id "/formValue")
                  {:query-params
                   {:form "INTERVENTIONS/Intervention_Details"
                    :user user
                    :mode "list"}})
       :body
       bs/to-string
       (json/decode keyword)
       (try (catch com.fasterxml.jackson.core.JsonParseException e
              (throw
               (ex-info "Resource not found"
                        {::anomalies/category ::anomalies/not-found
                         :error/parameters {:parameter :object}}))))
       (group-intervention-details intervention-id))
   (mapcat tabelize-intervention-detail-item)))

(defn leo-xml-call
  [xml-body xml-namespaces & [{:keys [leo-url api-extra-attrs] :as options}]]
  (let [req (xml-emit-string (leo-xml-request xml-body xml-namespaces options))
        resp (try (-> @(http/post (str leo-url "/request")
                                  {:headers {"Content-Type" "text/plain; charset=utf-8"}
                                   :query-params {:RC "csIntrfc"
                                                  :AD "KEY:/cf"}
                                   :body req}))
                  (catch Exception e e))
        resp-string (when (:body resp) (bs/to-string (:body resp)))
        resp-body (when resp-string (xml-parse resp-string))
      resp-body))
```

And here are 2 consecutive thread dumps, at about 10 minutes interval, that show threads stuck while dereferencing these queries.

First thread dump:
```
--
Jan 13 15:56:54 sp-sqa01 SQA_alpha[24105]: "manifold-pool-2-556" #1026 daemon prio=5 os_prio=0 tid=0x00007f58b4002000 nid=0x157 waiting on condition [0x00007f5893cb3000]
Jan 13 15:56:54 sp-sqa01 SQA_alpha[24105]:    java.lang.Thread.State: WAITING (parking)
Jan 13 15:56:54 sp-sqa01 SQA_alpha[24105]:         at sun.misc.Unsafe.park(Native Method)
[...]

PR #592, opened 2022-04-01
client: ensure response in case of Exception

arnaudgeiser:
## Context

This PR ensures a response is always given to the client even
when an Exception is thrown.

It has been reported by the following issue : #542

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -555,14 +555,15 @@
                             (reset! save-body body))
 
                           (-> (netty/safe-execute ch
-                                                  (http/send-message ch true ssl? req' body))
+                                (http/send-message ch true ssl? req' body))
                               (d/catch' (fn [e]
                                           (s/put! responses (d/error-deferred e))
                                           (netty/close ch))))))
 
                       ;; this will usually happen because of a malformed request
                       (catch Throwable e
-                        (s/put! responses (d/error-deferred e)))))
+                        (s/put! responses (d/error-deferred e))
+                        (netty/close ch))))
                   requests)
 
                 (s/on-closed responses
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -554,8 +554,11 @@
                             ;; might be different in case we use :multipart
                             (reset! save-body body))
 
-                          (netty/safe-execute ch
-                            (http/send-message ch true ssl? req' body))))
+                          (-> (netty/safe-execute ch
+                                                  (http/send-message ch true ssl? req' body))
+                              (d/catch' (fn [e]
+                                          (s/put! responses e)
+                                          (netty/close ch))))))
 
                       ;; this will usually happen because of a malformed request
                       (catch Throwable e
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -557,7 +557,7 @@
                           (-> (netty/safe-execute ch
                                                   (http/send-message ch true ssl? req' body))
                               (d/catch' (fn [e]
-                                          (s/put! responses e)
+                                          (s/put! responses (d/error-deferred e))
                                           (netty/close ch))))))
 
                       ;; this will usually happen because of a malformed request
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -539,7 +539,8 @@
                 (try
                   (send-streaming-body ch msg body)
                   (catch Throwable e
-                    (log/error e "error sending body of type " class-name)))))]
+                    (log/error e "error sending body of type " class-name)
+                    (throw e)))))]
 
       (when-not keep-alive?
         (handle-cleanup ch f))
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -509,6 +509,12 @@
             (d/catch' (fn [_]))))]
 
   (defn send-message
+    "Write an HttpMessage and body to a Netty channel.
+
+     Accepts Strings, ByteBuffers, ByteBufs, byte[], ChunkedInputs,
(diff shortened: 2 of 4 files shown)


src/aleph/http/client.clj
```diff
@@ -554,8 +554,11 @@
                             ;; might be different in case we use :multipart
                             (reset! save-body body))
 
-                          (netty/safe-execute ch
-                            (http/send-message ch true ssl? req' body))))
+                          (-> (netty/safe-execute ch
+                                                  (http/send-message ch true ssl? req' body))
+                              (d/catch' (fn [e]
+                                          (s/put! responses e)
```
KingMob, line 560: @arnaudgeiser 

@kachayev 's code in c9a6739db61948d64ef4c98e3df1d31753987fc3 wraps the throwable in an error-deferred, but this is raw. Which is correct?

KingMob:
@arnaudgeiser 

I've been reading Oleksii's old code today in the error-handling branch-1.0.0 branch, and unfortunately, there's a **lot** of diffs. On the one hand, he wrote many test cases, made commits _already_ for some of the issues that we're now addressing, and was (honestly) more familiar with the whole Aleph code base than either of us (well, me, for sure). On the other hand, the branch was never merged, so I don't know which ideas he meant to keep working on and which he considered done.

I think to the extent that he wrote tests, and his code passes those tests, we should defer to him in cases where his code diverges from ours, and think very hard about when we shouldn't use his code. I personally do not have the time to master Aleph's code base now and work on my own job. If you do, and can write up why we should or shouldn't use his code in this case, go for it, and let me know.

One possibility I'm mulling is to work through his 1.0.0 branch, commit-by-commit, cherry-picking/merging into the master branch and using each as an opportunity to understand a bit more. Most of the commits seem independent, with a few exceptions.

src/aleph/http/client.clj
arnaudgeiser: Both are correct actually and end up with the exact same result.
On my PR, I leverage the fact you can put a `Throwable` on `responses` which will be eventually coerced into
a manifold deferred : https://github.com/clj-commons/aleph/blob/master/src/aleph/http/client.clj#L581-L582

I did the minimum to solve the #542 issue. But I'm pretty sure we also need to wrap everything in a `try` like it's done here:
https://github.com/clj-commons/aleph/commit/c9a6739db61948d64ef4c98e3df1d31753987fc3#diff-f3c6a88143a34bab27cfd4dde73563c71de85a57da8133def509631b893d05ddR549-R550
I haven't done that modification since I haven't any test cases.

arnaudgeiser:
> I've been reading Oleksii's old code today in the error-handling branch-1.0.0 branch, and unfortunately, there's a lot of diffs. On the one hand, he wrote many test cases, made commits already for some of the issues that we're now addressing, and was (honestly) more familiar with the whole Aleph code base than either of us (well, me, for sure). On the other hand, the branch was never merged, so I don't know which ideas he meant to keep working on and which he considered done.

Maybe it's not the right place to have this discussion, but my personal feeling around this branch is it will never be merged. As you mentioned, there are too many changes we are not comfortable enough with. Even if we decide to merge it, we will never be sure that the open issues have been resolved, which is even more annoying.
On the current state, only Oleksii and Zach can "quickly" determine if merging the branch is safe or not. It's been two years from now, not sure everything is still fresh on their head. 
On my side, I consider the work done on the `1.0.0` (and the error-handling one) branch as a source of inspiration.

> I think to the extent that he wrote tests, and his code passes those tests, we should defer to him in cases where his code diverges from ours, and think very hard about when we shouldn't use his code. I personally do not have the time to master Aleph's code base now and work on my own job. If you do, and can write up why we should or shouldn't use his code in this case, go for it, and let me know.

All for it. However, the commit you pointed has only one test while you would expect two (this line probably has no test: https://github.com/clj-commons/aleph/commit/c9a6739db61948d64ef4c98e3df1d31753987fc3#diff-2d6587452350a3603e05f940d19be2b98e6377dfc7efc3dc1eb099ff332e4b3bR631). Furthermore, I don't know if the test he added cover the exact same issue as mine as it depends on another commit (the fact everything is wrapped into a `try`).

For the current PR, I can switch `d/error-deferred ex` instead of `ex` though.

> One possibility I'm mulling is to work through his 1.0.0 branch, commit-by-commit, cherry-picking/merging into the master branch and using each as an opportunity to understand a bit more. Most of the commits seem independent, with a few exceptions.

:+1: 
It would be great whether we could correlate the commits with the open issues. It's already a lot of work I know, but people using `Aleph` client knows it contains some really annoying bugs, in particular the following one : https://github.com/clj-commons/aleph/issues/532

Some of us are using Aleph server which is perfectly fine in the current state, but ditch the client one because of the locks in favor of something else (clj-http, hato, telex etc...)

KingMob:
Yeah, sorry if I wasn't clear, but I agree the 1.0.0 branch will never be merged as is. In general, as PRs get larger, the review time and failure risk go up _exponentially_.

I get that this is a pain point for you (and others), let me take a fresh look and try to make some headway.

src/aleph/http/client.clj
KingMob: Thanks for returning a deferred, I feel like that'll be safer/more future-proof.

What `try` is missing? You mean in the server-side code? Yeah, we probably need to add them to the stacks containing `send-message`

arnaudgeiser: `send-message` is used both on the client and on the server.
But I guess we are talking about the same thing though.

KingMob: Or audit the code to verify they're caught and handled already.

arnaudgeiser:
If I can be of any help, please do not hesitate.

KingMob:
Fixes #542

PR #593, opened 2022-04-11
Update byte-streams and use clj-commons namespace prefixes

skynet-gh:
Update byte-streams to 0.3.0 and use `clj-commons` prefix for `byte-streams` and `primitive-math`.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,7 @@
 (def other-dependencies
   '[[org.clojure/tools.logging "1.1.0" :exclusions [org.clojure/clojure]]
     [manifold "0.1.9"]
-    [org.clj-commons/byte-streams "0.2.10"]
+    [org.clj-commons/byte-streams "0.3.0"]
     [potemkin "0.4.5"]])
 
 (defproject aleph "0.4.7"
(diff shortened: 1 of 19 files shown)


project.clj
```diff
@@ -13,7 +13,7 @@
 (def other-dependencies
   '[[org.clojure/tools.logging "1.1.0" :exclusions [org.clojure/clojure]]
     [manifold "0.1.9"]
-    [org.clj-commons/byte-streams "0.2.10"]
+    [org.clj-commons/byte-streams "0.3.0"]
```
KingMob, line 16: Two things:

1. We directly require primitive-math in aleph, but apparently it's only a transitive dependency. This should be rectified. Can you add primitive-math as a dep here?
2. There's now a conflict in manifold versions thanks to the new byte-streams. Can you update manifold to 0.2.3?

If you run into issues, let me know.

KingMob (changes requested):
Looks good, but I have a few requests. See comment. Also, can you update the byte-streams requires under `examples`?

Thanks!

project.clj
skynet-gh: @KingMob updated to address those, thanks.

I've actually seen issues in building GraalVM `native-image` with manifold 0.2.3 for some reason, but I haven't found the source. I think I can pin the version in my project to 0.1.9 for now, and I will investigate that further.

KingMob: Hmmm. Well, we're slowly moving up the dep chain, so we can take a look at that in Manifold next

PR #425, opened 2018-11-10
Fix ExceptionInInitializerError in reified GenericFutureListener 

kachayev:
The issue was mentioned [here](https://github.com/ztellman/aleph/issues/365#issuecomment-378416967) and [here](https://github.com/ztellman/aleph/pull/406#issuecomment-434446433).

**The root of the problem:**

[Here](https://github.com/ztellman/manifold/blob/d67a8c1b9f1268c094895d70dbbf146521f5774b/src/manifold/deferred.clj#L281-L293) we have a `try-catch` block that compiles to a separate class `Deferred$fn__**` to be lately instanciated (and invoked) in `Deferred.success`. The class contains a static variable with `RT.readString` of the `ns` declaration, I assume that comes from a `clojure.tools.logging` toolchain. The problem is that our first attempt to use this class happens on a Netty's thread (`globalEventExecutor-**` created by [this](https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java#L58-L59) factory to be more specific) and... there's no appropriate class loader there. `RT.baseLoader` function [here](https://github.com/clojure/clojure/blob/a1722cf9a15d917ee78a117732cf5c5e01fe3557/src/jvm/clojure/lang/RT.java#L2176-L2177) falls back to a current thread context loader (because of the default value of `*use-context-classloader*` set to `true`) which is set to `NULL` on Netty's threads.

**The solution:**

The simplest thing we can do here is to carry class loader in a closure (either current loader from a thread local Clojure frame or newly created). I'm also going to submit a PR to `manifold` to set an appropriate class loader to each new thread created by [this](https://github.com/ztellman/manifold/blob/master/src/manifold/executor.clj#L43) factory. It will reduce the number of calls to `RT.makeClassLoader` as we need to do this only once per thread. 

I've been also thinking if it's worth the effort to try to submit an issue to Clojure JIRA with the problem definition but now I doubt if it's necessary. Being pessimistically careful here would introduce some performance penalties in a lot of places when in fact the code that was invoked from a non-Clojure thread is not that wide-spread pattern. And it's hard for the compiler to distinguish such cases in advanced. There're still a lot of improvements might be done at least to eliminate unnecessary calls to a reader.. but that's for sure a completely separate topic.

P.S. Submitting this a separate PR not to mess up with tons of changes from #406.

PR #595, opened 2022-04-26
Use FastThreadLocalThread

arnaudgeiser:
## Description

~This PR makes use of the `io.netty.util.concurrent.DefaultThreadFactory` from Netty instead of using the implementation of `manifold.executor`.
The reason for this is that `DefaultThreadFactory` is creating `FastThreadLocalThread` instead of JDK `Thread` [1].
As demonstrated on the following code, it's supposed to be faster (`fastGet`) [2].~

This PR ensures we are using `io.netty.util.concurrent.FastThreadLocalThread` when creating the Aleph executor.

Here are some (interesting?) links on  Netty:
- https://github.com/netty/netty/pull/11999
- https://github.com/netty/netty/issues/4401
- https://github.com/netty/netty/issues/4402

[1] : https://github.com/netty/netty/blob/f17b5fe3cd7155a0281d081edabe223e131ba332/common/src/main/java/io/netty/util/concurrent/DefaultThreadFactory.java#L121
[2] : https://github.com/netty/netty/blob/69e1d36cb19c034dcdcaaa1085c0387435c06975/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java#L100-L124

### Exercised code

```clojure
(require '[clj-async-profiler.core :as prof])

(deftest test-compressed-response
  (prof/profile (with-compressed-handler basic-handler
                  (doseq [[index [path result]] (map-indexed vector expected-results)
                          :let [resp @(http-get (str "http://localhost:" port "/" path)
                                                {:headers {:accept-encoding "gzip"}})
                                unzipped (try
                                           (bs/to-string (GZIPInputStream. (:body resp)))
                                           (catch ZipException _ nil))]]
                    (is (= "gzip" (get-in resp [:headers :content-encoding])) 'content-encoding-header-is-set)
                    (is (some? unzipped) 'should-be-valid-gzip)
                    (is (= result unzipped) 'decompressed-body-is-correct)))))


```

### Before
02-cpu-flamegraph-2022-04-26-23-52-57

### After
01-cpu-flamegraph-2022-04-26-23-49-01

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,7 @@
 
 (def other-dependencies
   '[[org.clojure/tools.logging "1.1.0" :exclusions [org.clojure/clojure]]
-    [manifold "0.2.3"]
+    [manifold "0.2.4"]
     [org.clj-commons/byte-streams "0.3.0"]
     [org.clj-commons/primitive-math "1.0.0"]
     [potemkin "0.4.5"]])
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -4,13 +4,9 @@
     [clj-commons.byte-streams :as bs]
     [clojure.tools.logging :as log]
     [manifold.deferred :as d]
-    [manifold.executor :as e]
     [manifold.stream :as s]
     [manifold.stream.core :as manifold]
     [clj-commons.primitive-math :as p]
-    [clojure
-     [string :as str]
-     [set :as set]]
     [potemkin :as potemkin :refer [doit doary]])
   (:import
     [java.io IOException]
@@ -29,7 +25,6 @@
      EpollSocketChannel
      EpollDatagramChannel]
     [io.netty.util Attribute AttributeKey]
-    [io.netty.handler.codec Headers]
     [io.netty.channel.nio NioEventLoopGroup]
     [io.netty.channel.socket ServerSocketChannel]
     [io.netty.channel.socket.nio
@@ -56,7 +51,7 @@
      ResourceLeakDetector$Level]
     [java.net URI SocketAddress InetSocketAddress]
     [io.netty.util.concurrent
-     GenericFutureListener Future]
+     DefaultThreadFactory GenericFutureListener Future]
     [java.io InputStream File]
     [java.nio ByteBuffer]
     [io.netty.util.internal SystemPropertyUtil]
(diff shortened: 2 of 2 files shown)


KingMob:
Hmm, I'm concerned about breaking things. 

`enumerating-thread-factory` doesn't promise to be anything more than a ThreadFactory based on its type hint, so I don't think swapping concrete classes will be the issue. The thread names will not be exactly the same (DefaultThreadFactory adds pool IDs), but hopefully nobody's got mission-critical logging regexes on thread names.

No, the only thing I'm actually concerned about is losing the classloader behavior @kachayev added to `m.executor/thread-factory` in https://github.com/clj-commons/manifold/commit/26038e03642ff8b0e125a7e35b333153b295c46f. See https://github.com/clj-commons/aleph/pull/425 and https://github.com/clj-commons/manifold/pull/161 for more background, but IIUC, Netty threads won't have the Clojure classloader set, so if a Clojure class has to be loaded for the first time on a Netty thread, it fails.

The fix would then be one of:
1. Copy `m.executor/thread-factory` to Aleph and replace `Thread.` with `FastThreadLocalThread.`
2. Add a `thread-class` param to `m.executor/thread-factory` that allows us to override Thread with FastThreadLocalThread
3. ...proxy DefaultThreadFactory in Aleph?

I like option 2 the most. I hoped there might be a way to globally tell Netty to use the Clojure classloader for its threads, but given how much it messes with classloaders itself, I think that's risky. 

Thoughts?

arnaudgeiser:
> The thread names will not be exactly the same (DefaultThreadFactory adds pool IDs), but hopefully nobody's got mission-critical logging regexes on thread names.

Good catch, I didn't notice the pool-id on the name

> No, the only thing I'm actually concerned about is losing the classloader behavior @kachayev added to m.executor/thread-factory in https://github.com/clj-commons/manifold/commit/26038e03642ff8b0e125a7e35b333153b295c46f. See https://github.com/clj-commons/aleph/pull/425 and https://github.com/clj-commons/manifold/pull/161 for more background, but IIUC, Netty threads won't have the Clojure classloader set, so if a Clojure class has to be loaded for the first time on a Netty thread, it fails.

I was not able to reproduce the issue with my changes while it's reproducible with Aleph 0.4.6.

```clojure
(defn handler [req]
  {:status 200
   :headers {"content-type" "text/plain"}
   :body "hello!"})

(def foo (http/start-server handler {:port 8080}))

;; after a sec

(.close foo)
``` 

Anyway, you are right, I think @kachayev poured a lot of thought on this change.

> I like option 2 the most. I hoped there might be a way to globally tell Netty to use the Clojure classloader for its threads, but given how much it messes with classloaders itself, I think that's risky.

All for it. I will work on a PR on `manifold` to be able to pass a `classname` as the last argument of `manifold.executor/thread-factory`.

KingMob:
At some point, we should consider destructuring maps, possibly with some fast destructuring lib. A lot of Aleph/Manifold fns have _way_ too many params already.

KingMob (approved):

Issue #567, opened 2021-04-29
How do multipart requests work?


valerauko:
I couldn't figure out how to handle incoming multipart requests on the server using the `decode-request` utility.

It seems that the stream gets closed as soon as the upload is done (which can be quite quick), which then calls `.destroy` on Netty's decoder, which in turn (I guess?) cleans up all the temp files, meaning none of the uploads are actually accessible in the handler.

Am I doing something wrong?

valerauko:
I see. So for now using maybe the Apache multipart decoder (the ring-core multipart-params middleware) is the recommended way of dealing with uploads?

arnaudgeiser:
For now, I would say there are three ways to handle multipart requests:

1. __Use the multipart-params middleware__
 
It's working as long as your body is an `java.io.InputStream`.
By default, every file is stored on the filesystem during 1 hour (configurable) [1]

An example:

```clojure
(require '[ring.middleware.multipart-params :refer [wrap-multipart-params]])

(defn handler [request]
  (let [params    (-> request :params)
        files     (vals params)
        tempfiles (map :tempfile files)]
    ;; do something with the `tempfiles`
    {:body "done!"}))

(aleph.http/start-server
 (wrap-multipart-params handler)
 {:port 8080})

@(http/post "http://localhost:8080" {:multipart [{:part-name "a file"
                                                  :content (io/file "test/file.txt")}]})

```

[1] : https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/multipart_params/temp_file.clj#L50-L51 


2. __Ensure the `decode-request` never use files__

This can be done by setting an infinite value for `memory-limit`.

```clojure
(require '[aleph.http.multipart :as multipart])

(defn handler [request]
  (let [chunks (multipart/decode-request request {:memory-limit Integer/MAX_VALUE})]
    (-> (d/loop []
          (d/chain' (s/take! chunks)
                    (fn [{:keys [content] :as chunk}]
                      (when (some? chunk)
                        ;; do something with the `content`
                        (d/recur)))))
        (d/chain'
         (fn [_]
           {:body "done!"})))))

(aleph.http/start-server handler {:port 8080})

@(http/post "http://localhost:8080" {:multipart [{:part-name "a file"
                                                  :content (io/file "test/file.txt")}]})
```

3. __Implement your own `decode-request`__

[...]

PR #596, opened 2022-05-03
Attempt to provide correct API for multipart resources management

arnaudgeiser:
## Description

This PR takes the commits from the `1.0.0` branch related to multipart requests and add
some tests around it : [Attempt to provide correct API for multipart resources management](https://github.com/clj-commons/aleph/pull/432)

Unfortunately, the tests have demonstrated the proposed implementation was still
not working as expected. There is absolutely no guarantee we'll able to read the content
of the stream (the files) before it has been closed and thus the `decoder` destroyed, removing
all the files before we were able to copy them.

~~The proposed fix is actually a __breaking change__ (I don't expect to get it merge as it is but I 
want to raise the discussion).~~
~~- Instead of returning a `manifold.stream`, we return a vector containing the `manifold.stream` 
and a function that can be called to destroy the decoder. That `destroy-decoder-fn` needs to be 
called on a `d/finally` to ensure all the resources have been cleaned up.~~

Here is the related issue : https://github.com/clj-commons/aleph/issues/567

~~As we don't want breaking changes, I guess we'll have to deprecate the `decode-request` function.
I see absolutely no-way to fix the current implementation... It's either we break or we deprecate at that point.~~

I tried to split the commits in three parts:
- The original commit from @kachayev (https://github.com/clj-commons/aleph/pull/596/commits/e67f920acc68fccde25e7448af66118731900e2d)
- Some tests to prove it's not working in some cases (https://github.com/clj-commons/aleph/pull/596/commits/5e20b5d041dccaadbdaaee0ebb44748834c8e6bc)
- ~~Introduce a configurable delay before destroying the decoder :  (https://github.com/clj-commons/aleph/pull/596/commits/0588cd58691ed90b9c3154c7ecd1c71fb37b0ebc)~~
- Ensure all files are removed before calling destroy : https://github.com/clj-commons/aleph/commit/16c9ed3136242a1ef137c6ad91150c3004531c8b

~~What are your thoughts about that?~~

### DIFF
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -26,11 +26,13 @@
     DefaultHttpContent
     DefaultHttpRequest
     FullHttpRequest
+    LastHttpContent
     HttpConstants]
    [io.netty.handler.codec.http.multipart
     Attribute
     MemoryAttribute
     FileUpload
+    HttpData
     HttpDataFactory
     DefaultHttpDataFactory
     HttpPostRequestDecoder
@@ -325,11 +327,27 @@
       (fn []
         (when (compare-and-set! destroyed? false true)
           (try
-            ;; we're removing each received http data chunk
-            ;; from cleanup queue before pushing to `parts` stream,
-            ;; meaning that this `destroy` call would only cleanup
-            ;; those chunck that were not consumed for some reasons
-            ;; (i.e. the user closes the stream given earlier)
+            ;; The `decoder` expects to receive a `LastHttpContent` to make use
+            ;; of `getBodyHttpDatas` and to actually complete the decoding
+            ;; operation.
+            ;; https://github.com/netty/netty/blob/839eda075531ec4bbdb247c6db2cf286e34a130b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L266
+            (.offer decoder (LastHttpContent/EMPTY_LAST_CONTENT))
+
+            ;; The call to `destroy` on the `decoder` will also delete some
+            ;; files that are supposed to be stored on memory while actually
+            ;; on the filesystem. Those are removed from the clean up process.
+            ;; https://github.com/netty/netty/blob/ee9bb63aa5382bf9930eb0df4b690286173f126a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L969-L970
+            (let [iter (.iterator (.getBodyHttpDatas decoder))]
+              (while (.hasNext iter)
+                (let [http-data ^HttpData (.next iter)]
+                  (when-not (.isInMemory http-data)
+                    (.remove iter)))))
+
+            ;; Since all the files have been removed from the automatic clean up
+            ;; process (through the lines just above and with `removeHttpDataFromClean`),
+            ;; we can finally call `destroy` on the `decoder`.
+            ;; We rely on the user to actually delete the files while handled
+            ;; by calling `netty/release` on each part.
             (.destroy decoder)
             (catch Exception e
               (log/warn e "exception when cleaning up multipart decoder"))))))
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -19,6 +19,7 @@
     Charset]
    [java.net
     URLConnection]
+   [io.netty.util ReferenceCounted]
    [io.netty.util.internal
     ThreadLocalRandom]
    [io.netty.handler.codec.http
@@ -37,7 +38,9 @@
     InterfaceHttpData
     InterfaceHttpData$HttpDataType]))
 
-(defn boundary []
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  boundary []
   (-> (ThreadLocalRandom/current) .nextLong Long/toHexString .toLowerCase))
 
 (defn mime-type-descriptor
@@ -47,7 +50,9 @@
    (when encoding
      (str "; charset=" encoding))))
 
-(defn populate-part
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  populate-part
   "Generates a part map of the appropriate format"
   [{:keys [part-name content mime-type charset transfer-encoding name]}]
   (let [file? (instance? File content)
@@ -79,7 +84,9 @@
 ;;
 ;; Note, that you can use transfer-encoding=nil or :binary to leave data "as is".
 ;; transfer-encoding=nil omits "Content-Transfer-Encoding" header.
-(defn part-headers [^String part-name ^String mime-type transfer-encoding name]
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  part-headers [^String part-name ^String mime-type transfer-encoding name]
   (let [cd (str "Content-Disposition: form-data; name=\"" part-name "\""
              (when name (str "; filename=\"" name "\""))
              "\r\n")
@@ -89,7 +96,9 @@
               (str "Content-Transfer-Encoding: " (cc/name transfer-encoding) "\r\n"))]
     (bs/to-byte-buffer (str cd ct cte "\r\n"))))
 
-(defn encode-part
+(defn
+  ^{:deprecated "use aleph.http.multipart/encode-request instead"}
+  encode-part
   "Generates the byte representation of a part for the bytebuffer"
   [{:keys [part-name content mime-type charset transfer-encoding name] :as part}]
   (let [headers (part-headers part-name mime-type transfer-encoding name)
@@ -162,6 +171,33 @@
     (let [req' (.finalizeRequest encoder)]
       [req' (when (.isChunked encoder) encoder)])))
 
+(defrecord MultipartChunk [part-name
+                           content
+                           name
+                           charset
+                           mime-type
+                           transfer-encoding
+                           memory?
+                           file?
+                           file
+                           size
+                           ^ReferenceCounted raw-http-data]
+  ReferenceCounted
+  (refCnt [_]
+    (.refCnt raw-http-data))
+  (retain [_]
+    (.retain raw-http-data))
+  (retain [_ increment]
+    (.retain raw-http-data increment))
+  (^ReferenceCounted touch [_]
+    (.touch raw-http-data))
+  (^ReferenceCounted touch [_ ^Object hint]
+    (.touch raw-http-data hint))
+  (release [_]
+    (.release raw-http-data))
+  (release [_ decrement]
+    (.release raw-http-data decrement)))
+
 (defmulti http-data->map
   (fn [^InterfaceHttpData data]
     (.getHttpDataType data)))
@@ -169,35 +205,53 @@
 (defmethod http-data->map InterfaceHttpData$HttpDataType/Attribute
   [^Attribute attr]
   (let [content (.getValue attr)]
-    {:part-name (.getName attr)
-     :content content
-     :name nil
-     :charset (-> attr .getCharset .toString)
-     :mime-type nil
-     :transfer-encoding nil
-     :memory? (.isInMemory attr)
-     :file? false
-     :file nil
-     :size (count content)}))
+    (map->MultipartChunk
+     {:part-name (.getName attr)
+      :content content
+      :name nil
+      :charset (-> attr .getCharset .toString)
+      :mime-type nil
+      :transfer-encoding nil
+      :memory? (.isInMemory attr)
+      :file? false
+      :file nil
+      :size (count content)
+      :raw-http-data attr})))
 
 (defmethod http-data->map InterfaceHttpData$HttpDataType/FileUpload
   [^FileUpload data]
   (let [memory? (.isInMemory data)]
-    {:part-name (.getName data)
-     :content (when memory?
-                (bs/to-input-stream (netty/acquire (.content data))))
-     :name (.getFilename data)
-     :charset (-> data .getCharset .toString)
-     :mime-type (.getContentType data)
-     :transfer-encoding (.getContentTransferEncoding data)
-     :memory? memory?
-     :file? true
-     :file (when-not memory? (.getFile data))
-     :size (.length data)}))
+    (map->MultipartChunk
+     {:part-name (.getName data)
+      :content (when memory?
+                 (bs/to-input-stream (netty/acquire (.content data))))
+      :name (.getFilename data)
+      :charset (-> data .getCharset .toString)
+      :mime-type (.getContentType data)
+      :transfer-encoding (.getContentTransferEncoding data)
+      :memory? memory?
+      :file? true
+      :file (when-not memory? (.getFile data))
+      :size (.length data)
+      :raw-http-data data})))
 
 (defn- read-attributes [^HttpPostRequestDecoder decoder parts]
-  (while (.hasNext decoder)
-    (s/put! parts (http-data->map (.next decoder)))))
+  (d/loop []
+    (if-not (.hasNext decoder)
+      (d/success-deferred true) ;; go for another chunk of body
+      (let [^InterfaceHttpData data (.next decoder)]
+        (if (nil? data)
+          ;; this probably could happen only in case of
+          ;; simultaneous access to the decoder object...
+          (d/success-deferred true)
+          (d/chain'
+           (s/put! parts (http-data->map data))
+           (fn [succeed?]
+             (if succeed?
+               (do
+                 (.removeHttpDataFromClean decoder data)
+                 (d/recur))
+               (d/success-deferred false)))))))))
 
 (defn decode-request
   "Takes a ring request and returns a manifold stream which yields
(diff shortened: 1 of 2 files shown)


KingMob:
I'll have to take a deeper look later. I take it there's no further commentary from Oleksii on this?

arnaudgeiser:
@kachayev did a nice work at summarizing the different paths we could take for the cleanup API. https://github.com/clj-commons/aleph/pull/432#issuecomment-449185986
He proposed an implementation that didn't introduce any breaking change. 
However, unless I'm mistaken, `s/on-closed` might not always happen on the same thread that consumes 
the stream.

```clojure
(s/on-closed
      parts
      (fn []
        (when (compare-and-set! destroyed? false true)
          (try
            (.destroy decoder)
            (catch Exception e
              (log/warn e "exception when cleaning up multipart decoder"))))))
```

Now that I think about it, maybe we can include an additional `opts` to the `decode-request` 
function and do not perform `s/on-closed` when used. That way, we'll be able to keep the current
behavior and introduce a new one without any breaking change.

Something like `clean-on-closed?` which defaults to true.
What do you think?

KingMob:
How do the proposed changed play with upcoming Netty changes, in particular: https://github.com/netty/netty-incubator-buffer-api/issues/48 ?

arnaudgeiser:
To be honest, I'm close to think we should just deprecate the feature on Aleph (and mark it as unstable) and just provide examples about how it can be done using Aleph/Netty.
In the current situation, we will always run to have something that works more or less. It will either be too low/high level to be usable so I wouldn't do any commitment about a leaky Aleph abstraction based on a leaky Netty abstraction that might be revisited on the future.

KingMob:
That works for me, especially if Netty itself has issues with it. 

1. Want to add docs/examples demoing it with that Apache middleware you mentioned?
2. Is there something we can link to in Netty to track their internal status/progress?

arnaudgeiser:
> 1. Want to add docs/examples demoing it with that Apache middleware you mentioned?

Yes, I will add the content of this PR as an example + one that uses multipart upload.

> 2. Is there something we can link to in Netty to track their internal status/progress?

I don't know.

KingMob (approved):

arnaudgeiser:
After a second thought about this.
I think a good API around this would be to not return a `manifold.stream` at all but provide
something like `multipart/transduce` that will return a deferred but will close the resources (ByteBuf + decoder)
under the hood.

I will try to cook something...

arnaudgeiser:
Quick update...
No progress for now on that topic, I expect to spend some time on the following days (it matures on my mind, not on the PR for now).

arnaudgeiser:
"Good news", I have been able to land a fix on top of Oleksii's PR.
The idea is to introduce a `destroy-delay` before calling `.destroy` on the decoder
to let some time to the client to consume the stream. [1]

It seems a bit hacky but that's the only way I found to not break or deprecate the current
API which returns a `manifold.stream`.

I still think having a higher level API on Aleph might be valuable, like `transduce-request`.
I have a branch for this, but as long as there is no demand from Aleph users, I won't spend
more time create a true PR (I will keep this dormant for a while) [2].

[1] : https://github.com/clj-commons/aleph/commit/3f6bb240f70fb8db7908668a8f6491092cda78f4
[2] : https://github.com/clj-commons/aleph/commit/9217079d2efc2720b71fd0bccb3bcc25833dded5#diff-272a90e0af5e55488c08269c3ab128f28f2d63dcad30be761aca8c59cd0769c1R348-R362

src/aleph/http/multipart.clj
```diff
@@ -162,42 +170,87 @@
...
+      :raw-http-data data})))
 
 (defn- read-attributes [^HttpPostRequestDecoder decoder parts]
-  (while (.hasNext decoder)
-    (s/put! parts (http-data->map (.next decoder)))))
+  (d/loop []
+    (if-not (.hasNext decoder)
+      (d/success-deferred true) ;; go for another chunk of body
+      (let [^InterfaceHttpData data (.next decoder)]
+        (if (nil? data)
```
KingMob, line 242: Minor: consider `if-some`

src/aleph/http/multipart.clj
```diff
@@ -207,15 +260,48 @@
...
+   (defn file-upload-handler [req]
+     (let [chunks (multipart/decode-request req)]
+       (d/chain'
+         (stream/take! chunks)
+         (fn [{:keys [file] :as chunk}]
+           (io/copy file writer)
+           (netty/release chunk)
+           (stream/close! chunks)
+           {:status 200 :body \"Succesfull!\"}))))
+   ```
```
KingMob, line 287: Maybe add this to `examples/`?

test/aleph/http/multipart_test.clj
```diff
@@ -229,10 +264,27 @@
     ;; charset
     (is (= "ISO-8859-1" (get-in chunks [5 :charset])))
 
-    (.close ^java.io.Closeable s)))
+    (.close ^java.io.Closeable s))))
 
 (deftest test-mutlipart-request-decode-with-ring-handler
```
KingMob, line 269: Trivial: typo in "multipart"

test/aleph/http/multipart_test.clj
```diff
@@ -188,39 +189,73 @@
...
+                     ;; when `chunks` stream has been closed and there is no more
+                     ;; `chunk` left.
+                     body
+                     (let [packed-chunk (pack-chunk chunk)]
+                       ;; we should be able to read file before removal (if any)
+                       (when (some? file)
+                         (is (some? (slurp (.getAbsolutePath ^java.io.File file)))))
+                       (is (netty/release chunk))
+                       ;; temp file is now removed (if any)
+                       (when (some? file)
```
KingMob, line 226: `when-some`

src/aleph/http/multipart.clj
```diff
@@ -162,42 +170,87 @@
...
+      :charset (-> data .getCharset .toString)
+      :mime-type (.getContentType data)
+      :transfer-encoding (.getContentTransferEncoding data)
+      :memory? memory?
+      :file? true
+      :file (when-not memory? (.getFile data))
+      :size (.length data)
+      :raw-http-data data})))
 
 (defn- read-attributes [^HttpPostRequestDecoder decoder parts]
```
KingMob, line 237: Can we rename this private fn? Nothing about it suggests it "reads attributes" to me.

src/aleph/http/multipart.clj
```diff
@@ -162,42 +170,87 @@
...
+        (if (nil? data)
+          ;; this probably could happen only in case of
+          ;; simultaneous access to the decoder object...
+          (d/success-deferred true)
+          (d/chain'
+           (s/put! parts (http-data->map data))
+           (fn [succeed?]
+             (if succeed?
+               (do
+                 (.removeHttpDataFromClean decoder data)
```
KingMob, line 251: Why are we telling it not to clean up the file again?

src/aleph/http/multipart.clj
```diff
@@ -207,15 +260,48 @@
...
+   (require '[aleph.http.multipart :as multipart])
+   (require '[aleph.netty :as netty])
+   (require '[manifold.stream :as stream])
+   (require '[clojure.java.io :as io])
+
+   (defn file-upload-handler [req]
+     (let [chunks (multipart/decode-request req)]
+       (d/chain'
+         (stream/take! chunks)
+         (fn [{:keys [file] :as chunk}]
```
KingMob, line 282: These are _parts_, not _chunks_, correct? We should be precise. IIUC, chunks may be incomplete pieces of a part, and a part corresponds to everything between the multipart boundaries (minus headers, etc), e.g., a whole file. If that's right, we should indicate that `decode-request` returns a stream of parts, not chunks.

test/aleph/http/multipart_test.clj
```diff
@@ -188,39 +189,73 @@
...
+                   (bs/to-string content))]
+    (assoc chunk :content content')))
+
+(defn- pack-chunk [chunk]
+  (-> chunk
+      (coerce-chunk-content)
+      (dissoc :file :raw-http-data)))
+
+(defn- decode-handler [req options]
+  (let [chunks (mp/decode-request req options)]
```
KingMob, line 210: As above, isn't this a stream of parts, not chunks?

test/aleph/http/multipart_test.clj
```diff
@@ -188,39 +189,73 @@
...
+    (->
+     (d/loop [body []]
+       (d/chain' (s/take! chunks)
+                 (fn [{:keys [file] :as chunk}]
+                   (Thread/sleep 20)
```
KingMob, line 215: Why sleep? Will user handlers have to do this, too?

src/aleph/http/multipart.clj
arnaudgeiser: It's named that way because it reads `io.netty.handler.codec.http.multipart.MixedAttribute` [1] or `io.netty.handler.codec.http.multipart.MixedFileUpload` [2].

I will look for a better name.


[1] : https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/multipart/MixedAttribute.java#L27
[2] : https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/multipart/MixedFileUpload.java

KingMob (commented):
I'm not generally comfortable with a time-based clean-up approach. Nothing prevents the user from sticking the files in a list to batch transfer overnight, after the files was deleted. OTOH, I don't see any great alternatives. We could:

1. Force the user to manually clean up, just like with raw-streams. I think Netty does this.
2. Stick files in a temp directory, and let the OS clean up when it needs space. (Is that safe?)
3. Clean up automatically if the user follows a certain pattern, namely, a) being totally done with the parts when their handler exits, and b) no handing off references to other threads.
4. Delete after a delay when the stream closes, like this PR.

What about something like option 3, where we delete automatically, but users can indicate they want to keep a file around (in which case, it becomes their responsibility to delete)?

Even if we stick with 4, I'd rather bump up the default delay a lot, like several hours, maybe a whole day.

src/aleph/http/multipart.clj
arnaudgeiser: Here is the rationale behind the idea. [1]

> Reimplemented public API using io.netty.util.ReferenceCounted and relying on netty/release for each consumed chunk. I played a bit with the version of API that exposes a single .close method for the entire decoder object and found that it lacks flexibility as I cannot deallocate memory from a specific chunk (potentially freeing a lot of space as soon as possible) while all other chunks are still in processing... So, I think that releasing them one-by-one is the best option in terms of resources usage even tho' it might be a bit confusing from the first glance (well, nothing special if you have some experience with Netty/C/C++/Rust etc, but probably not something you're doing in Clojure on a daily basis). /cc @gsnewmark @ztellman

Basically, it transfers the responsibility of cleaning up the resources from `HttpPostRequestDecoder#destroy` to the user so each part/chunk can be removed one per one.   

> But if so, shouldn't we only call this when .isInMemory is true?

The same applies when it's a file. 
When `.release` (and thus `.deallocate`) [2] is called on a `AbstractMixedHttpData`, it cleans the associated file [3].

[1] : https://github.com/clj-commons/aleph/pull/432#issuecomment-457755670
[2] : https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/multipart/AbstractMixedHttpData.java#L101
[3] : https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/multipart/AbstractDiskHttpData.java#L281

src/aleph/http/multipart.clj
arnaudgeiser: Agreed.

test/aleph/http/multipart_test.clj
arnaudgeiser: It's basically how I demonstrated the proposed implementation wasn't working as expected.

`stream/on-closed` is called when the data has been fully read from the `body` and it has been closed.
However, nothing guarantees the data has been taken on the consumer side. Some parts might deleted
before the client has the time to consume them.

Without that `Thread/sleep` which simulates time spent handling a part, all the tests pass.

Oleksii also gives a bit of context around it [1]

> At all. chunks stream will be eventually closed by the implementation as soon as decoded parts are exhausted. on-close is implemented there to release all resources that we didn't consume from the decoder yet. But the stream is only a delivery mechanism, so when it's closed we know that values were conveyed to a consumer, but we don't know if they are still in use. You can pass them to any async processing pipeline you want, and the closing of the stream means nothing for us.

[1] : https://github.com/clj-commons/aleph/pull/432#issuecomment-449376344

arnaudgeiser:
> 1. Force the user to manually clean up, just like with raw-streams. I think Netty does this.

Yes, that could work whether it was a new API.
But the current implementation rely on the fact that the clean-up is performed automatically
when the decoder is destroyed.

This would require an additional parameter, it might not be a bad idea though.

> 2. Stick files in a temp directory, and let the OS clean up when it needs space. (Is that safe?)

This is a time-bomb for me and even if the OS cleans up the space at some point, most platforms
are monitored for disk-usage. I'm afraid it will create even more issues on our side.

> 3. Clean up automatically if the user follows a certain pattern, namely, a) being totally done with the parts when their handler exits, and b) no handing off references to other threads.

While I have a rough idea how `a` could be implemented, I'm totally armless for `b`.

> 4. Delete after a delay when the stream closes, like this PR.

The `multipart_params` middleware on ring seems to use the same approach and defaults to one hour. [1]

[1] : https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/multipart_params/temp_file.clj#L50-L51

So for me:

4 > 1 > 2 > 3

But I will follow your call on this.

test/aleph/http/multipart_test.clj
KingMob: OK

KingMob:
"armless"? I haven't heard that one before.

Regardless, for option 3, I wasn't suggesting we implement a and b. I meant, a and b are a contract that the user will have to promise to follow, there's no enforcement, unfortunately. In that sense, what we would do is wrap the user handler, dispose of resources when the handler exits, and it's up to the user to not screw up. Since we'd delete by default, if the user needed the files to hang around, we'd add an API for them to unmark for deletion. Does that change your ranking at all?

If clj-http does an hour, I'm fine with that. There's no perfect solution, and at least 1 hour has parity going for it...

arnaudgeiser:
Regarding this comment:

> I'm not generally comfortable with a time-based clean-up approach. Nothing prevents the user from sticking the files in a list to batch transfer overnight, after the files was deleted.

That's why on our example we recommend to perform an `io/copy` so the temporary
file can be removed afterwards when calling `.release`. [1]

[1] : https://github.com/clj-commons/aleph/pull/596/files#diff-272a90e0af5e55488c08269c3ab128f28f2d63dcad30be761aca8c59cd0769c1R283

src/aleph/http/multipart.clj
arnaudgeiser: Actually, it has been called `chunk` because we introduced the `MultipartChunk` record...
Renaming `chunk` to `part` makes sense on one hand but not on the other.

arnaudgeiser:
What we can do also...

5. Match `multipart_params` middleware and (optionally) delete __THE FILES__ after a delay when the stream closes. (~like 4)

And we revisit/ditch the rest of this PR:
- it's still the responsibility of the decoder to release/delete the files (and not the user)
- it's still the responsibility of the `decode-handler` [1] to release the memory (no more `.acquire` [2]) and not the user
- no more evolved read-attributes [3]
- MultipartChunk can be removed
- we keep the broken behavior when files are involved (and add an optional `destroy-delay` parameter)
- we recommend users to use an "infinite" memory-limit by default
- we recommend users to use a "destroy-delay" (of 1 hour) when files will be involved

And we have something that works with Aleph without introducing
any breaking changes, risky fixes, bad defaults and that we can easily justify 
regarding the rest of the Clojure ecosystem.

WDYT?..

(:cry: : bringing 1.0.0 features to main branch always raise so many questions)

[1] : https://github.com/clj-commons/aleph/pull/596/files#diff-272a90e0af5e55488c08269c3ab128f28f2d63dcad30be761aca8c59cd0769c1R325
[2] : https://github.com/clj-commons/aleph/pull/596/files#diff-272a90e0af5e55488c08269c3ab128f28f2d63dcad30be761aca8c59cd0769c1R226
[3] : https://github.com/clj-commons/aleph/pull/596/files#diff-272a90e0af5e55488c08269c3ab128f28f2d63dcad30be761aca8c59cd0769c1R238-R251

src/aleph/http/multipart.clj
KingMob: Ehh, this is confusing. How can we bring clarity?

KingMob:
> That's why on our example we recommend to perform an `io/copy` so the temporary file can be removed afterwards when calling `.release`. 

Ahh, can we add a comment explaining the "why" of that call for readers?

KingMob:
I think option 5 sounds good. Fundamentally, I think it's up to the user to not leak resources, since there's no solution we can do that's 100% correct.

I'm not sure about recommending an infinite memory-limit, though. Very large, yes, but infinite?

Which do you think is the best option at this point?

arnaudgeiser:
> I'm not sure about recommending an infinite memory-limit, though. Very large, yes, but infinite?

Yes, infinite might not be a good recommendation.
I would rather formulate as :

When using `decode-request` (without `destroy-delay`), set a `memory-limit` to ensure your data won't be stored in temporary files and enforce the presence of the `content-length` header up to this maximum value. Whether the `content-length` is greater than your `memory-limit` or absent, return a bad request.

Something along those lines...

> Which do you think is the best option at this point?

Another round on this PR to apply __5__. :dancers:

arnaudgeiser:
Okay, I decided to start over again and revisit all my assumptions.
I removed all my modifications since the beginning and did a deep dive onto Netty
regarding multipart uploads. 

Oleksii's approach was good.
Unfortunately, calling `removeHttpDataFromClean` was not enough as some files
were borrowed inside `bodyListHttpData` [1].

I end up having to do only those modifications [2]
Which means:

- No new `destroy-delay` parameter
- Cleanup of the files is left to the user (with netty/release)

Before I get more involved in this (cleanup, addressing all the remarks you did above), can you give me an opinion on this approach @KingMob ?

[1] : https://github.com/netty/netty/blob/839eda075531ec4bbdb247c6db2cf286e34a130b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L970
[2] : https://github.com/clj-commons/aleph/pull/596/commits/16c9ed3136242a1ef137c6ad91150c3004531c8b

KingMob:
@arnaudgeiser I think your new approach is sound. There's no perfect option, so let's just do the best we can, document it, and let the user decide what to do with the files.

arnaudgeiser:
Pfff... Actually, the proposed solution is a breaking change even in its initial form as you will have to call `netty/release` on each file manually.
Let's close this one in favor of this : #656 

The findings were interesting though if one day we need a fine granularity clean up.

PR #598, opened 2022-05-12
Redirects broken with :method key

### LINKED ISSUE #597: :follow-redirects? does not follow through
```
› cat deps.edn 
{:deps {aleph/aleph {:mvn/version "0.4.7"}}}

› clj
Clojure 1.10.3
nil
user=> (-> (http/request {:method :get :url "http://httpbin.org/status/301" :follow-redirects? true}) deref :status)
301
```

expected status 200

KingMob:
Fixes #597.

If you just want to see the relevant changes, ignore the first cleanup commit, and look at the second commit.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,7 @@
                    (map
                      #(vector (symbol "io.netty" (str "netty-" %)) netty-version)
                      netty-modules))
-  :profiles {:dev  {:dependencies [[org.clojure/clojure "1.10.1"]
+  :profiles {:dev  {:dependencies [[org.clojure/clojure "1.10.3"]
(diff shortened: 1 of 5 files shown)


arnaudgeiser (approved):
The fix is fine.
However, I would add the following test:

```clojure
(deftest test-follow-redirects
  (with-handler (fn [{:keys [uri]}]
                  (case uri
                    "/200" {:status 200}
                    "/301" {:status 301
                            :headers {"Location" "/200"}}))
    (is (= 200 (:status @(http/request {:method :get
                                        :url (str "http://localhost:" port "/301")
                                        :follow-redirects? true}))))))
```

PR #601, opened 2022-05-19
Ensure IdleStateHandler is the first handler

arnaudgeiser:
## Description

This ensures the `IdleStateHandler` is always at the first place on the pipeline.

For the complete history : https://github.com/clj-commons/aleph/pull/511

Here are the most relevant parts of the conversation : 

> Having the idle handlers after the other handlers results in them timing out the connection after the timeout without any consideration of actual traffic. This can lead to a socket being closed while it is still actively being used. Attaching the idle handlers first means that a connection will only be treated as idle if no traffic has been sent through the socket for a given period of time.

> This is leading a problem in production for us, where Aleph would terminate connections while a request was in-flight (~20ms after receiving the data) because the connection was older than the request. We have disabled the idle-timeout to get things working for now, but this change will fix things properly.

> In our case, we were seeing the connecting being closed at exactly the idle timeout, every time. We set the idle timeout to 65 seconds, then opened a telnet connection and constantly sent requests and received responses. After 65 seconds the connection was closed, no matter what.

> We're fine with imprecise tracking of the idle timeout, as long as there's some tracking of actual traffic.

### DIFF
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -633,7 +633,7 @@
 (defn attach-idle-handlers [^ChannelPipeline pipeline idle-timeout]
   (if (pos? idle-timeout)
     (doto pipeline
-      (.addLast "idle" ^ChannelHandler (IdleStateHandler. 0 0 idle-timeout TimeUnit/MILLISECONDS))
+      (.addFirst "idle" ^ChannelHandler (IdleStateHandler. 0 0 idle-timeout TimeUnit/MILLISECONDS))
       (.addLast "idle-close" ^ChannelHandler (close-on-idle-handler)))
     pipeline))
 
--- a/src/aleph/http/core.clj
+++ b/src/aleph/http/core.clj
@@ -633,7 +633,7 @@
 (defn attach-idle-handlers [^ChannelPipeline pipeline idle-timeout]
   (if (pos? idle-timeout)
     (doto pipeline
-      (.addFirst "idle" ^ChannelHandler (IdleStateHandler. 0 0 idle-timeout TimeUnit/MILLISECONDS))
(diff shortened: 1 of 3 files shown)


src/aleph/http/core.clj
```diff
@@ -633,7 +633,7 @@
 (defn attach-idle-handlers [^ChannelPipeline pipeline idle-timeout]
   (if (pos? idle-timeout)
     (doto pipeline
-      (.addLast "idle" ^ChannelHandler (IdleStateHandler. 0 0 idle-timeout TimeUnit/MILLISECONDS))
+      (.addFirst "idle" ^ChannelHandler (IdleStateHandler. 0 0 idle-timeout TimeUnit/MILLISECONDS))
       (.addLast "idle-close" ^ChannelHandler (close-on-idle-handler)))
```
KingMob, line 637: Does idle-close need to be moved too?

arnaudgeiser: No, it doesn't have to.
The important part is the `IdleStateHandler` to be at the beginning of the `ChannelPipeline` (the closest to the `Socket.read` and the writes to the `ChannelHandlerContext`. [1]

`idle-close` only implements what to do when an `IdleState` is detected and can be anywhere on the pipeline.

[1] : https://netty.io/4.1/api/io/netty/channel/ChannelPipeline.html

KingMob: Can we safely group them together? It feels tidier that way.

KingMob (approved):

src/aleph/http/core.clj
arnaudgeiser: I finally moved the `attach-idle-handlers`

KingMob (approved):

KingMob:
@arnaudgeiser I just realized, we fixed this for the server, but not the client.

Issue #461, opened 2019-01-03
OutOfMemory error due to high memory usage of io.aleph.dirigiste.Pool (again)


rborer:
Hi,

This is a follow-up of #394. I'm still experiencing random OOM errors while leaving Aleph running for a few hours.

Here is what MAT shows:

![image](https://user-images.githubusercontent.com/361403/50633914-3d7cd380-0f4d-11e9-9363-193073e2a67e.png)

Looking at [dirigiste `_queues` map size](https://github.com/ztellman/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L215), it contains 49'152 items, which seems quite *a lot* to me! So it seems like clearing of unused queues somehow doesn't happens.

I believe I have pinpointed the issue to the following piece of code in dirigiste: https://github.com/ztellman/dirigiste/blob/f5b035094fa947f9779d22d6d08a88f39f4a6e88/src/io/aleph/dirigiste/Pool.java#L329-L343.

More specifically line 331: `entry.getValue().getUtilization(1) == 0`

Here, the code gets the 1 percentile to decide if the queue is unused. By looking at how this percentile gets computed [in dirigiste we have the following piece of code](https://github.com/ztellman/dirigiste/blob/master/src/io/aleph/dirigiste/Stats.java#L178-L180):

```
if (t == 1.0) {
                return (double) vals[cnt-1];
}
```

So for a percentile of 1, it gets the last value of the reservoir of stats and assume it's the latest added metric.

This is fine as long as the reservoir contains 4'096 values or less. As soon as the reservoir has more values, [**every newly added metric is placed at a random location**](https://github.com/ztellman/dirigiste/blob/master/src/io/aleph/dirigiste/Stats.java#L35-L45).

How can this happens? By default, aleph configures dirigiste with the following settings:

### 1. a `control-period` of `60'000`

https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http.clj#L146

### 2. a `sample-period` of `10`

https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/flow.clj#L48

As such, dirigiste control loop attempts to clean unused queues only every 6'000 cycles, which is above the 4'096 reservoir size. So there is a very high probability that `entry.getValue().getUtilization(1)` does **not** represent the 1 percentile anymore.

[...]

Issue #394, opened 2018-06-25
OutOfMemory error due to high memory usage of io.aleph.dirigiste.Pool


rborer:
Hi,

I'm currently incorporating Aleph into https://paper.li/ service as a replacement for our venerable crawler (based on Netty 3.10!).

On our test environment (which is crawling around 350 different URLs per minute, much less than our production system), my crawling component is crashing in OutOfMemory after a few hours of processing 😢 .

I started analysing the heap dumps and it seems to indicate that `io.aleph.dirigiste.Pool` is retaining the majority of heap as shown below (Xmx is set to 9G):

<img width="835" alt="screen shot 2018-06-25 at 09 40 42" src="https://user-images.githubusercontent.com/361403/41836781-04880c58-785c-11e8-9230-fd4fd113bc00.png">

From my understanding so far, this high memory usage seems to be related to the Pool statistics. As shown below, a single `java.util.HashMap` (for the statistics probably) is retaining 40% of the heap:

<img width="710" alt="screen shot 2018-06-25 at 09 43 55" src="https://user-images.githubusercontent.com/361403/41836937-77ba9e3e-785c-11e8-9609-21f9ddf46e9e.png">

And memory is mostly filled with longs & doubles that seems related to the Pool statistics:

<img width="519" alt="screen shot 2018-06-25 at 09 44 03" src="https://user-images.githubusercontent.com/361403/41837089-cd45318e-785c-11e8-82fe-462169577fb3.png">

Am I correct in my analysis to finger pointing aleph Pool? If yes, is there any setting on my side I can modify to relieve this memory usage?

Thanks.

gsnewmark:
May be related https://github.com/ztellman/dirigiste/pull/5#issuecomment-159140122

rborer:
I was finally able to run [Eclipse Memory Analyzer](https://www.eclipse.org/mat/) on my laptop (previously I hit a bug between OSX, SWT & Eclipse that would freeze it at startup).

So I could drill-down into the `Pool` instance, to see that indeed it is the `_stats` field that is the culprit:

<img width="1281" alt="screen shot 2018-06-28 at 10 52 15" src="https://user-images.githubusercontent.com/361403/42023956-63a99578-7ac1-11e8-92c1-749fe223b8d7.png">

Here we can see that this HashMap instance contains 29k entries (not much), but obviously each entry consumes more than 100mb of retained heap!

[...]

PR #602, opened 2022-06-13
add pool customization options

Yneth:
It might not be the best way to add such customisation, as it clashes with other fields (total-connections, target-utilization, max-queue-size, stats-callback) but it should work.

The rationale of this change is to add possibility to pick your own pool implementation. In my case dirigiste pool does not work due to memory leak in 
image

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -103,6 +103,8 @@
    | `control-period` | the interval, in milliseconds, between use of the controller to adjust the size of the pool, defaults to `60000`
    | `dns-options` | an optional map with async DNS resolver settings, for more information check `aleph.netty/dns-resolver-group`. When set, ignores `name-resolver` setting from `connection-options` in favor of shared DNS resolver instance
    | `middleware` | a function to modify request before sending, defaults to `aleph.http.client-middleware/wrap-request`
+   | `pool-builder-fn` | a one arity function which returns a `io.aleph.dirigiste.IPool` from a map containing the following keys: `generate`, `destroy`, `control-period`, `max-queue-length` and `stats-callback`.
+   | `pool-controller-builder-fn` | a zero arity function which returns a `io.aleph.dirigiste.IPool$Controller`.
 
    the `connection-options` are a map describing behavior across all connections:
 
@@ -136,9 +138,7 @@
    | `http-headers` | (HTTP proxy only) an optional map to set additional HTTP headers when establishing connection to the proxy server
    | `tunnel?` | (HTTP proxy only) if `true`, sends HTTP CONNECT to the proxy and waits for the 'HTTP/1.1 200 OK' response before sending any subsequent requests. Defaults to `false`. When using authorization or specifying additional headers uses tunneling disregarding this setting
    | `connection-timeout` | timeout in milliseconds for the tunnel become established, defaults to 60 seconds, setting is ignored when tunneling is not used."
-  [{:keys [pool-builder-fn
-           pool-controller-builder-fn
-           connections-per-host
+  [{:keys [connections-per-host
            total-connections
            target-utilization
            connection-options
@@ -146,7 +146,9 @@
            stats-callback
            control-period
            middleware
-           max-queue-size]
+           max-queue-size
+           pool-builder-fn
+           pool-controller-builder-fn]
     :or {connections-per-host 8
          total-connections 1024
          target-utilization 0.9
@@ -175,7 +177,7 @@
         create-pool-fn      (or pool-builder-fn
                                 flow/instrumented-pool)
         create-pool-ctrl-fn (or pool-controller-builder-fn
-                                #(Pools/utilizationController %1 %2 %3))
+                                #(Pools/utilizationController target-utilization connections-per-host total-connections))
         pool (create-pool-fn
               {:generate (fn [host]
                            (let [c (promise)
@@ -192,10 +194,7 @@
                                     client/close-connection))
                :control-period control-period
                :max-queue-size max-queue-size
-               :controller (create-pool-ctrl-fn
-                            target-utilization
-                            connections-per-host
-                            total-connections)
+               :controller (create-pool-ctrl-fn)
                :stats-callback stats-callback})]
     @(deliver p pool)))
 
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -136,7 +136,9 @@
    | `http-headers` | (HTTP proxy only) an optional map to set additional HTTP headers when establishing connection to the proxy server
    | `tunnel?` | (HTTP proxy only) if `true`, sends HTTP CONNECT to the proxy and waits for the 'HTTP/1.1 200 OK' response before sending any subsequent requests. Defaults to `false`. When using authorization or specifying additional headers uses tunneling disregarding this setting
    | `connection-timeout` | timeout in milliseconds for the tunnel become established, defaults to 60 seconds, setting is ignored when tunneling is not used."
-  [{:keys [connections-per-host
+  [{:keys [pool-builder-fn
+           pool-controller-builder-fn
+           connections-per-host
            total-connections
            target-utilization
            connection-options
@@ -170,7 +172,11 @@
                         (some? log-activity)
                         (assoc :log-activity (netty/activity-logger "aleph-client" log-activity)))
         p (promise)
-        pool (flow/instrumented-pool
+        create-pool-fn      (or pool-builder-fn
+                                flow/instrumented-pool)
+        create-pool-ctrl-fn (or pool-controller-builder-fn
+                                #(Pools/utilizationController %1 %2 %3))
+        pool (create-pool-fn
               {:generate (fn [host]
                            (let [c (promise)
                                  conn (create-connection
@@ -186,7 +192,7 @@
                                     client/close-connection))
                :control-period control-period
                :max-queue-size max-queue-size
-               :controller (Pools/utilizationController
+               :controller (create-pool-ctrl-fn
                             target-utilization
                             connections-per-host
                             total-connections)
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -103,8 +103,8 @@
    | `control-period` | the interval, in milliseconds, between use of the controller to adjust the size of the pool, defaults to `60000`
    | `dns-options` | an optional map with async DNS resolver settings, for more information check `aleph.netty/dns-resolver-group`. When set, ignores `name-resolver` setting from `connection-options` in favor of shared DNS resolver instance
    | `middleware` | a function to modify request before sending, defaults to `aleph.http.client-middleware/wrap-request`
-   | `pool-builder-fn` | a one arity function which returns a `io.aleph.dirigiste.IPool` from a map containing the following keys: `generate`, `destroy`, `control-period`, `max-queue-length` and `stats-callback`.
-   | `pool-controller-builder-fn` | a zero arity function which returns a `io.aleph.dirigiste.IPool$Controller`.
+   | `pool-builder-fn` | an optional one arity function which returns a `io.aleph.dirigiste.IPool` from a map containing the following keys: `generate`, `destroy`, `control-period`, `max-queue-length` and `stats-callback`.
+   | `pool-controller-builder-fn` | an optional zero arity function which returns a `io.aleph.dirigiste.IPool$Controller`.
 
    the `connection-options` are a map describing behavior across all connections:
 


KingMob:
Thanks for submitting this. I think I would prefer to fix Dirigiste, if possible. PRs can be submitted at https://github.com/clj-commons/dirigiste

At a first glance, why isn't this being garbage-collected? Are you collecting stats yourself somehow? 35k seems like an awful lot stats objects.

Yneth:
@KingMob for some reason stats are not disposed. My fix was to implement Pool without stats at all. Unfortunately I do not have much time but I'll try to check what could be the reason.

KingMob:
@Yneth Can you share a minimal test case reproducing it?

arnaudgeiser:
Are you sure the `35'769` arrays of long are all referencing `dirigiste`?
I don't ask you to check for all obviously, but can you randomly pin point some places to ensure it's really the case.

arnaudgeiser:
I'm asking this because a reservoir can contains at most 2048 elements, so :

35769×16408÷1024 = ~573 MiB

and the value you showed is twice as bigger.

KingMob:
@Yneth Even if you don't have a test case, did you save the profiling session? It would be nice to take a look at it.

Yneth:
@KingMob no, but I am sure that it gives problems. My fix was to create custom pool implementation without metrics, those that are using Reservoir.

It is a bit hard to create minimal test as I need to do some investigation beforehand and I am short on time.

My case looks the following way:
5 pools per proxy running http requests all day. it takes around 6 hours when problems arise.

Also I see uncatched error warning from manifold even though I am using m/catch everywhere, just saying in case it could also be a problem.

src/aleph/http.clj
```diff
@@ -186,7 +192,7 @@
                                     client/close-connection))
                :control-period control-period
                :max-queue-size max-queue-size
-               :controller (Pools/utilizationController
+               :controller (create-pool-ctrl-fn
                             target-utilization
                             connections-per-host
                             total-connections)
```
KingMob, line 198: - This setup doesn't support using Dirigiste's fixed pools (or even other utilization pools), because the arities don't match. We should change it to a 0-arity thunk up above.

KingMob (commented):
I wonder if, rather than overriding the fns, users should be allowed to supply a whole pool. hmmm

Yneth:
@KingMob I guess giving a possibility to specify the whole pool would be nice, but it would require making `(create-connection` function public. Not sure if it is ok to expose this so decided to go with the simplest draft solution

src/aleph/http.clj
```diff
@@ -136,7 +136,9 @@
    | `http-headers` | (HTTP proxy only) an optional map to set additional HTTP headers when establishing connection to the proxy server
    | `tunnel?` | (HTTP proxy only) if `true`, sends HTTP CONNECT to the proxy and waits for the 'HTTP/1.1 200 OK' response before sending any subsequent requests. Defaults to `false`. When using authorization or specifying additional headers uses tunneling disregarding this setting
    | `connection-timeout` | timeout in milliseconds for the tunnel become established, defaults to 60 seconds, setting is ignored when tunneling is not used."
-  [{:keys [connections-per-host
+  [{:keys [pool-builder-fn
+           pool-controller-builder-fn
```
KingMob, line 140: Also need to document these options in the docstring

KingMob:
@Yneth Any updates? After doing some work with Dirigiste recently, I want your PR. 😄

Yneth:
will send an update later today

arnaudgeiser (commented):
I'm not sold to the idea of this PR.
To me, we'll have an hard time documenting `create-pool-fn` and `crate-pool-ctrl-fn` as it's not that obvious.

I would rather focus on fixing `Dirigiste` to ensure we are never leaking memory and
it's a stable building block.
A test case, thread dump, heap dump or any that could help us reproduce
this issue would be awesome.

arnaudgeiser:
Any update from your side?

What we actually need from you to determine if the issue is related to `io.aleph.dirigiste.Stats$**` is the
number of those instances when your memory starts growing.

I wasn't able to reproduce on my side but the expected pattern is this:

image

I have three `io.aleph.dirigiste.Executor` (4 reservoirs each) + one `io.aleph.dirigiste.Pool`(4 reservoirs). (16 reservoirs).
Those are actually capped to 32'892 bytes (4096 * 8 bytes = 32'768 + a few other references)

- How many of those `UniformDoubleReservoir` do you have?
- How many different `io.aleph.dirigiste.Executor` they reference?
- How many `io.aleph.dirigiste.Pool` they reference?

arnaudgeiser:
Ok.. ok...
I'm able to create a pattern where it can be problematic.
With the following code, I'm able to create tons of Reservoir :

image

```clojure
(ns dirigiste-leak.core)

(require '[manifold.deferred :as d])
(require '[clj-commons.byte-streams :as bs])

(comment
  (http/start-server (constantly {:body "hello!"}) {:port 8081})


  (defn get [nb]
    (apply d/zip (mapv (fn [_] (d/chain (http/get (format "http://%s.localhost:8081/" (java.util.UUID/randomUUID)))
                                        :body
                                        bs/to-string)) (range nb))))

  (d/loop [d (get 10)]
    (d/chain' d
              (fn [_]
                (prn "Done!")
                (Thread/sleep 300)
                (d/recur (get 10000))))))
```

Each different domain creates me a new Reservoir.

image

They are supposed to be clean by the following code

```java
// clear out any unused queues
_lock.lock();
for (Map.Entry<K,Stats> entry : _stats.entrySet()) {
    K key = entry.getKey();
    if (entry.getValue().getUtilization(1) == 0
        && _queues.get(key).objects.get() == 0) {
        _queues.remove(key).shutdown();

        // clean up stats so they don't remain in memory forever
        _queueLatencies.remove(key);
        _taskLatencies.remove(key);
        _queueLengths.remove(key);
        _utilizations.remove(key);
        _taskArrivalRates.remove(key);
        _taskCompletionRates.remove(key);
        _taskRejectionRates.remove(key);
    }
}
_lock.unlock();
```

https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L331-L332

It does its job when the conditions are met.
I was putting the system under stress here but it might be somewhere along
those lines of code where some objects are not removed from the queues.

https://github.com/clj-commons/aleph/blob/master/src/aleph/http.clj#L321-L375

I would be interested to know if you have the same pattern @Yneth.

arnaudgeiser:
And as @Yneth mentioned this:

> Also I see uncatched error warning from manifold even though I am using m/catch everywhere, just saying in case it could also be a root cause.

KingMob:
@arnaudgeiser Your test case, that crawls a lot of domains, reminds me a _lot_ of:

1. https://github.com/clj-commons/aleph/issues/394#issuecomment-404400781 in #394 
2. https://github.com/clj-commons/aleph/issues/461#issuecomment-453145969 in #461 

As for the objects not being cleaned up:

1. Dumb question: Manually running the GC didn't clean any of that up, I assume?
2. If you look [here](https://github.com/clj-commons/aleph/blob/159d617c93074715d3c9656394af1db8ae383a59/src/aleph/http.clj#L174), the connection pool is keyed off the host. This makes sense, since the connection is immediately usable (returning the conn for different hosts would enable us to avoid object creation, but would still require changing fields, and reconnecting), but the consequence is no pooling between hosts. So I'm not surprised to see usage going up. OTOH, in your example, it looks like 10 hosts generated 1.3GB, which seems high. Hmmm.
3. The code you linked to in `aleph.http/request` will 99% of the time `release`, not `dispose`. Most normally-ended HTTP/1.1 connections are "Connection: keep-alive", so they won't meet the criteria for disposal.
4. The check for `getUtilization(1)` is checking the 100th percentile of the utilization stats. This means that _all_ utilizations in the stats time window must be 0. The reservoir size is 4096 items, and if the sample period is the default 10ms, then we should expect it to take ~41s of no usage for the reservoir to become all zeros. If looking at the memory usage too early is the problem, then lowering the sample period may help. It might be worth checking for bugs here, though, I know there have been some utilization calc issues in the past. (clj-commons/dirigiste#23)

KingMob:
OK, I misread a couple things. 

Your code is recurring with another 10000 domains. In my tests, this works out to ~770k per domain. Still highish, but there's 5 major maps of domain->reservoir, and each reservoir has an array of 4096 longs/double, so we have 10010 x 5 x 4096 x 8 bytes = ~1.6 GB. 770k per domain is still high, though; I know Dirigiste is tunable, we could probably make it consume less.

I also took a closer look at the utilizations, which is where it gets interesting. With one conn per domain, it's continually computing the utilization level as 2 (if I'm computing Double/longBitsToDouble correctly). In [the code](https://github.com/clj-commons/dirigiste/blob/8db17bb6ba1466d7fc9b8d2c66e4378506e06d5b/src/io/aleph/dirigiste/Pool.java#L306-L315), it seems that `queueLength` is always 1, `available` is always 0, and `objects` is always 0. (Not sure what `objects` represents... I think it's the number of _free_ objects.) Anyway, that means `double utilization = 1.0 - ((available - queueLength) / Math.max(1.0, objects));` is always 2.

It feels like there's a bug here. Regardless of whether the conn is released or disposed, you'd think an unused connection shouldn't count towards utilization. I'm not sure if it's an edge artifact of having exactly 1 item per domain, or if this has always been a problem, and that's just where we see it first, when testing zillions of domains.

KingMob:
Just forced it to make 3 conns per domain, and all I got was endless utilization values of 4. I need to set this aside for now.

Even if it's a leak, it mostly affects people talking to many different domains, so I'm not sure how high priority it is.

arnaudgeiser:
Do not get me wrong, everything is fine with my test case.
I'm only in trouble because I'm overwhelming the pool, but eventually,
everything will be garbage collected if I stop doing requests.
I just wanted to demonstrate a situation where a lot of memory could be
allocated on the heap.

This pattern, which is not something we should expect from a production
system is a situation where we could use a bunch of memory.

But again, without a heap dump here, we are stuck doing hypothesis
which might be wrong.

arnaudgeiser:
> Just forced it to make 3 conns per domain, and all I got was endless utilization values of 4. I need to set this aside for now.

Ah yes, interesting lead here.

arnaudgeiser:
I will spend some time to write unit tests for Dirigiste beause some parts of the code seems to be off, really.

A few examples:

```java
public void sample(double n) {
    int cnt = _count.incrementAndGet();
    if (cnt <= RESERVOIR_SIZE) {
        _values.set(cnt-1, Double.doubleToLongBits(n));
    } else {
        int idx = ThreadLocalRandom.current().nextInt(cnt);
        if (idx < RESERVOIR_SIZE) {
            _values.set(idx, Double.doubleToLongBits(n));
        }
    }
}
```

1. We can overflow here 
```clojure
(def atomic-integer (java.util.concurrent.atomic.AtomicInteger. Integer/MAX_VALUE))
(.incrementAndGet atomic-integer) ;; -2147483648
(.incrementAndGet atomic-integer) ;; -2147483647
```

2. Then, it will just blow the Reservoir
```clojure
(def atomic-long-array (java.util.concurrent.atomic.AtomicIntegerArray. 4096))
(.set atomic-long-array -2147483648 0.5) ==> ArrayIndexOutOfBoundsException
```

3. Once the reservoir is filled, you have no guarantee the last `idx` (position 4095) will be
set to the current utilization as we randomly get a slot over the 4096 position.

https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Stats.java#L61-L71

The last `idx` (position 4095) is important here because we rely only on this
one to determine if the queue is no longer in use.

```
if (t == 1.0) {
  return (double) vals[cnt-1];
}
```
https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Stats.java#L166-L168

KingMob:
Yeah, there's some weird stuff going on, with few tests, and a lot of that is why I'm not a fan of Dirigiste.

Based on a sampling period of 10ms, we should expect overflow and crash after ~8 months left untouched. However, the stats are meant to be replaced periodically. For both Executor and Pool, in `startControlLoop`, the stats reservoirs are replaced once each control loop by `updateStats()`. For `Pool.updateStats`, the `.toMap()` calls are actually _destructive_(!). It `remove()`s the keys as it builds the map. So theoretically, overflow isn't an issue, but I find Dirigiste's logic indirect enough that I wouldn't bet on that.

> Once the reservoir is filled, you have no guarantee the last idx (position 4095) will be
set to the current utilization as we randomly get a slot over the 4096 position.

Actually, it's weirder than that. It picks a random number from 0 to cnt-1, and only if that's < 4096 does it store the sample at all. This has the effect of updating the reservoir less and less frequently as time goes on, until it's reset. The longer the control period, the less and less likely that later samples will have an effect. I'm not sure what the rationale is over using a ring buffer, and the commit messages are no help.

That being said, using the value in the last position when looking at the percentiles is fine. Inside the reservoir's `toArray()` fns, you'll see it actually returns them sorted (!), so looking at the last element will always be the maximal utilization. Dirigiste isn't consdiering the _most recent_ utilitzation value, it's considering the _highest_ utilization in the control period.

The problem is that the utilization reservoir contains the same value everywhere, so it doesn't matter what percentile we look at. (The behavior above where later samples in the control period are less likely to be recorded is also a problem, but if the stats are properly reset, it should still work once we have a fallow period.)

So either a) the utilization is never being calculated correctly in our scenario (all scenarios?), or b) it's calculated correctly, but the stats aren't being reset, and thus by the time utilization drops to zero, the odds of it being recorded are low, and the odds of *all non-zero* samples being displaced is vanishingly low. I think it's more likely to be a; if b were true, I should be able to find other values _somewhere_, but I can't find any.

It's really annoying that .toArray and .toMap aren't simple transformation fns. Sorry you lost time on that.

arnaudgeiser:
Thanks for the explanations, I missed so many things here.

> For both Executor and Pool, in startControlLoop, the stats reservoirs are replaced once each control loop by updateStats().

Woow... at least that explains the "high" rate allocation I was seeing.

>  For Pool.updateStats, the .toMap() calls are actually destructive(!). It remove()s the keys as it builds the map.

Right

> Dirigiste isn't consdiering the most recent utilitzation value, it's considering the highest utilization in the control period.

Understood, I missed the fact that `toArray` sorts the values. :face_with_spiral_eyes: 

> The problem is that the utilization reservoir contains the same value everywhere, so it doesn't matter what percentile we look at. (The behavior above where later samples in the control period are less likely to be recorded is also a problem, but if the stats are properly reset, it should still work once we have a fallow period.)

This is currently bugging me. There is a correlation with `Pool.updateStats` but according to what you explain, it's like most of the time the 4096 positions are totally useless.
I need to clean my head here.

> So either a) the utilization is never being calculated correctly in our scenario (all scenarios?), or b) it's calculated correctly, but the stats aren't being reset, and thus by the time utilization drops to zero, the odds of it being recorded are low, and the odds of all non-zero samples being displaced is vanishingly low. I think it's more likely to be a; if b were true, I should be able to find other values somewhere, but I can't find any.

Let's figure this out then!

KingMob:
Keep in mind, this is still low priority to my mind. The Dirigiste pooling problem only applies to outgoing client connections, where talking to zillions of different servers is unlikely. The server uses Netty to pool incoming connections/channels, so it should be unaffected by Dirigiste's pool stats issue. The most likely scenario is a few outside servers are talked to, and a few MBs are wasted, but that's it.

If you're burning to fix it, I won't stop you tho 😄

arnaudgeiser:
Yes, I agree.
But some people are currently having issues with the `io.aleph.dirigiste.Pool` while it's unlikely
they are currently talking to a lot of different servers.

By adding tests, I expect to see patterns where it could go wrong
without having to communicate to thousands of different domains.

Going back to this PR, I'm still not sold to it
but I can see reasons why people don't want to use Dirigiste
on the current situation. According to the discussions we just had, it's quite
an heavy machine under the hood.
Let's provide an escape hatch then!

I know you want to sunset Dirigiste, but I would prefer stabilizing it.
The code base is not that huge and the rationale behind it is still good IMO.

KingMob:
I agree that other memory leaks of Dirigiste should be fixed. I'm just afraid that the pool leak is a red herring for their other issues.

And while I'd like to sunset it, that's more of a "next version" goal. Replacing it in current Aleph makes me think of these lines from the movie State and Main:

> ... you come by next Sunday, we'll give you a second chance.
> 
> Sister, the only second chance I know is the chance to make the same mistake twice.

What kind of escape hatch are you thinking of, if you don't want to use this PR?

arnaudgeiser:
> I agree that other memory leaks of Dirigiste should be fixed. I'm just afraid that the pool leak is a red herring for their other issues.

I'm also afraid of this!

> What kind of escape hatch are you thinking of, if you don't want to use this PR?

We'll use this PR. But I am resigned, not convinced!

KingMob:
@Yneth Are you still interested in this?

Yneth:
factory method of pool is dependant on local variables of the API method, as well as some private functions.
Of course there should be an opportunity to override, but at least in a user friendly way

KingMob:
@Yneth Don't restrict yourself on where or how to fix this. We're open to the right solution, and if you've followed the conversation so far, the most likely correct place is where Dirigiste computes pool utilization and/or cleans up old pool objects.

Also, based on your initial screenshot, which shows the leak is from the Stats reservoirs, I'm pretty sure Arnaud and I are looking at the same leak, so I don't think we need a dump anymore. (Of course, if that's _not_ your leak, please let us know.)

arnaudgeiser:
> Also, based on your initial screenshot, which shows the leak is from the Stats reservoirs, I'm pretty sure Arnaud and I are looking at the same leak, so I don't think we need a dump anymore. (Of course, if that's not your leak, please let us know.)

In any case, the heap dump can help us diagnose whether we are making the correct assumptions.
We'll be able to track the state of those Stats.

arnaudgeiser:
@Yneth : Are you hitting the same ports on those hosts?

https://github.com/clj-commons/aleph/blob/master/src/aleph/http/client.clj#L97

arnaudgeiser:
I spent hours adding (JUnit) tests to Dirigiste and I start to have a better
understanding of the situation and the meaning of those different values.
(Consider what's below as a partial understanding of the situation, but I need to share some
thoughts, some still need to be validated).

So..

## Utilization

Theorically, this calculation is supposed to be at least partially correct:

```
double utilization = 1.0 - ((available - queueLength) / Math.max(1.0, objects));
```

- __available__ refers to the number of the free objects created in the pool. Those are objects that have been created at some point by `.acquire` and then `.release` to the pool.
- __queueLength__ refers to the number of `.acquire` waiting to get an object from the pool. (there is still something I don't understand on the code regarding the pending `_takes`)
- __objects__ refers to the numbers of objects created by `.acquire`

Considering no object are present on the pool at the beginning:

- An utilization of __1__ means we are using all the objects that have been created __available == 0__
  - Scenarios:
    - `.acquire` (1 objects)
    - `.acquire` => `.release` => `.acquire` (1 objects)
    - `.acquire` => `.acquire` => `.acquire` (3 objects)
    - `.acquire` => `.acquire` => `.dispose` (1 objects)
  - __Concerning scenario__:
    - `.acquire` => `.dispose` (__0 objects__, __0 available__)
    
- An utilization of __0.5__ means we are currently using half of the objects that have been created __available*2__ == __objects__
  - Scenarios:
    - `.acquire` => `.acquire` => `.release` (1 available, 2 objects)
 
- An utilization of __0__ means all the objects are not in use on the pool __available__ == __objects__ (but not 0)
  - Scenarios:
    - `.acquire` => `.release` => `.acquire` => `.release` (2 objects)
    - `.acquire` => `.release` => `.acquire` => `.dispose` (1 objects)

- To have an utilization > __1__, we need to have pending `.acquire` and __available == 0__. (it will happen when `controller.shouldIncrement` returns __false__) [1]
  - Scenarios (considering __0 objects__):
    - `.acquire` (2 utilization)
    - `.acquire` => `.acquire` (3 utilization)
    - `.acquire` => `.acquire`  => `.acquire`  (4 utilization)

[1] : https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L269

## Reservoir size

I reached the conclusion that using a fixed value of `4096` is a waste.

```java
static final int RESERVOIR_SIZE = 4096;
```
We should use `controlPeriod/samplePeriod` instead.

By default
- On a Pool __controlPeriod__ is 1000 and __samplePeriod__ is 25 which gives us a size of 40.
- On an Executor __controlPeriod__ is 10000 and __samplePeriod__ is 25 which gives us a size of 400.

## Use a Ring Buffer instead

While it can be seen as an optimization, replacing the stats every second is not ideal.
A Ring Buffer might be more suitable for this use case.

arnaudgeiser:
> So either a) the utilization is never being calculated correctly in our scenario (all scenarios?), or b) it's calculated correctly, but the stats aren't being reset, and thus by the time utilization drops to zero, the odds of it being recorded are low, and the odds of all non-zero samples being displaced is vanishingly low. I think it's more likely to be a; if b were true, I should be able to find other values somewhere, but I can't find any.

Long story short:

It's definitely `a`. When using an utilization controller, the objects will be removed as soon as they are not in 
use which means we'll always end up on the concerning scenario (0 objects, 0 available) => utilization of 1.

The fix might be as simple as this:

```java
double getUtilization(int available, int queueLength, int objects) {
    if(objects==0 && queueLength==0) {
        return 0;
    }
    return 1.0 - ((available - queueLength) / Math.max(1.0, objects));
}
```

However, we are now subject to race condition (youpi...).
Our Queue, that we get on `.acquire` [1] might be destroyed at any moment by the `startControlLoop` [2], and we might end up on IllegalStateException... __already shutdown__. [3]

So close and so far at the same time.

For the good news, I've been able to cover around `90%` of the Pool code base.

[1] : https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L409
[2] : https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L333
[3] : https://github.com/clj-commons/dirigiste/blob/master/src/io/aleph/dirigiste/Pool.java#L163

arnaudgeiser:
I finally reached the bottom of this.

Here is the related PR on Dirigiste : https://github.com/clj-commons/dirigiste/pull/32

arnaudgeiser:
The PR is ready for review.
I applied the @KingMob comments.

arnaudgeiser (approved):

KingMob:
Merged.

Issue #603, opened 2022-06-13
Stacktrace on JVM exit when calling .close on server from SignalHandler thread


ivarref:
Hi and thanks for all the work that is put into this library!

I have a strange issue that only arises when calling `(.close server)` from a `SignalHandler` thread.

Reproducing repo: https://github.com/ivarref/aleph-shutdown-stacktrace

Use `$ ./demo.sh` and after the server is up, hit `Ctrl-C`.
The console will show something like this:
```
2022-06-13 11:20:57:337 +0000 [clojure-agent-send-off-pool-0] [INFO] aleph-shutdown-stacktrace - shutting down in 30 seconds ...
2022-06-13 11:20:57:385 +0000 [main] [INFO] aleph-shutdown-stacktrace - Server listening at 0.0.0.0:8080
^C2022-06-13 11:20:59:508 +0000 [SIGINT handler] [INFO] aleph-shutdown-stacktrace - Received SIGINT
2022-06-13 11:20:59:515 +0000 [SIGINT handler] [INFO] aleph-shutdown-stacktrace - OK shutdown server
2022-06-13 11:20:59:516 +0000 [main] [INFO] aleph-shutdown-stacktrace - About to exit
2022-06-13 11:21:01:522 +0000 [globalEventExecutor-1-1] [WARN] DefaultPromise - An exception was thrown by aleph.netty$wrap_future$reify__4259.operationComplete()
java.lang.ExceptionInInitializerError
        at manifold.deferred.Deferred.success(deferred.clj:399)
        at manifold.deferred$success_BANG_.invokeStatic(deferred.clj:244)
        at manifold.deferred$success_BANG_.invoke(deferred.clj:241)
        at aleph.netty$wrap_future$reify__4259.operationComplete(netty.clj:223)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:578)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:552)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:491)
        at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:616)
        at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:605)
        at io.netty.util.concurrent.DefaultPromise.setSuccess(DefaultPromise.java:96)
[...]

Issue #365, opened 2018-03-10
Shutting down server gracefully


kachayev:
There was a similar [discussion](https://github.com/ztellman/aleph/issues/309) which is already closed, but the issue is still reproducible, i.e. [here](https://travis-ci.org/ztellman/aleph/jobs/351663601#L797).  It's easy to catch "Executor is shutdown!" in [these](https://github.com/ztellman/aleph/blob/d6cb8cbb0b29d0b6d0bea69142551d377a8e67cd/test/aleph/http_test.clj#L303-L312) test cases when closing the server after the client caught timeout **while** response callbacks are still producing answers. 

What's going on here.

1. Calling `close` on the server closes [socket channel](https://github.com/ztellman/aleph/blob/d6cb8cbb0b29d0b6d0bea69142551d377a8e67cd/src/aleph/netty.clj#L921) and [event loop group](https://github.com/ztellman/aleph/blob/d6cb8cbb0b29d0b6d0bea69142551d377a8e67cd/src/aleph/netty.clj#L922).

2. After that we call [`netty/wait-for-close`](https://github.com/ztellman/aleph/blob/d6cb8cbb0b29d0b6d0bea69142551d377a8e67cd/src/aleph/netty.clj#L927).

[...]

PR #604, opened 2022-06-20
Ensure a DynamicClassLoader is the Context Class Loader

arnaudgeiser:
## Description

This PR is a proposal to solve the https://github.com/clj-commons/aleph/issues/603, https://github.com/clj-commons/aleph/issues/365 issues.

The following commit adds a unit test regarding the problem we are currently facing
when some code is running outside of a DynamicClassLoader : https://github.com/clj-commons/aleph/commit/404ebcca64ef1e2e414c3f4e08fb2fdef3a33b3f
Then, the following one is basically the fix that landed on Riemann (https://github.com/riemann/riemann/pull/951) : https://github.com/clj-commons/aleph/commit/3f9fdfb92b9d45ca2b8c6c13c7795e708fbe9895.

All the credits go to @KingMob obviously who did the all work.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -11,6 +11,7 @@
     [clj-commons.primitive-math :as p]
     [potemkin :as potemkin :refer [doit doary]])
   (:import
+    [clojure.lang DynamicClassLoader]
     [java.io IOException]
     [io.netty.bootstrap Bootstrap ServerBootstrap]
     [io.netty.buffer ByteBuf Unpooled]
@@ -200,28 +201,30 @@
   (->> (bs/convert x (bs/stream-of ByteBuf) {:chunk-size chunk-size})
     (s/onto nil)))
 
+(defn ensure-dynamic-classloader
+  "Ensure the context class loader is always a `clojure.lang.DynamicClassLoader`
+  to prevent `ClassNotFoundException`.
+  https://github.com/clj-commons/aleph/issues/603."
+  []
+  (let [thread (Thread/currentThread)
+        context-class-loader (.getContextClassLoader thread)
+        compiler-class-loader (.getClassLoader clojure.lang.Compiler)]
+    (when-not (instance? DynamicClassLoader context-class-loader)
+      (.setContextClassLoader
+        thread (DynamicClassLoader. (or context-class-loader
+                                        compiler-class-loader))))))
+
 (defn wrap-future
   [^Future f]
   (when f
     (if (.isSuccess f)
       (d/success-deferred (.getNow f) nil)
-      (let [d (d/deferred nil)
-            ;; workaround for the issue with executing RT.readString
-            ;; on a Netty thread that doesn't have appropriate class loader
-            ;; more information here:
-            ;; https://github.com/clj-commons/aleph/issues/365
-            class-loader (or (clojure.lang.RT/baseLoader)
-                             (clojure.lang.RT/makeClassLoader))]
+      (let [d (d/deferred nil)]
         (.addListener f
           (reify GenericFutureListener
             (operationComplete [_ _]
-              (try
-                (. clojure.lang.Var
-                   (pushThreadBindings {clojure.lang.Compiler/LOADER
-                                        class-loader}))
-                (operation-complete f d)
-                (finally
-                  (. clojure.lang.Var (popThreadBindings)))))))
+              (ensure-dynamic-classloader)
+              (operation-complete f d))))
         d))))
 
 (defn allocate [x]
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1,6 +1,7 @@
 (ns aleph.netty
   (:refer-clojure :exclude [flush])
   (:require
+    [aleph.netty.impl :refer [operation-complete]]
     [clj-commons.byte-streams :as bs]
     [clojure.tools.logging :as log]
     [manifold.deferred :as d]
@@ -8,9 +9,6 @@
     [manifold.stream :as s]
     [manifold.stream.core :as manifold]
     [clj-commons.primitive-math :as p]
-    [clojure
-     [string :as str]
-     [set :as set]]
     [potemkin :as potemkin :refer [doit doary]])
   (:import
     [java.io IOException]
@@ -29,7 +27,6 @@
      EpollSocketChannel
      EpollDatagramChannel]
     [io.netty.util Attribute AttributeKey]
-    [io.netty.handler.codec Headers]
     [io.netty.channel.nio NioEventLoopGroup]
     [io.netty.channel.socket ServerSocketChannel]
     [io.netty.channel.socket.nio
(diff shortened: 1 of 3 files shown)


src/aleph/netty.clj
```diff
@@ -203,41 +201,30 @@
...
+  to prevent `ClassNotFoundException`.
+  https://github.com/clj-commons/aleph/issues/603."
+  []
+  (let [thread (Thread/currentThread)
+        context-class-loader (.getContextClassLoader thread)
+        compiler-class-loader (.getClassLoader clojure.lang.Compiler)]
+    (when-not (instance? DynamicClassLoader context-class-loader)
+      (.setContextClassLoader
+        thread (DynamicClassLoader. (or context-class-loader
+                                        compiler-class-loader))))))
```
KingMob, line 215: So, my one concern is that `compiler-class-loader` may not have the correct parent classloader set in this code from Riemann, since we need to ensure Netty's classloader is in the chain. I was going to try something like:

```clojure
(let [thread (Thread/currentThread)]
  (when-not (instance? DynamicClassLoader (.getContextClassLoader thread))
    (.setContextClassLoader thread (RT/makeClassLoader))))
```

which seems simpler and (maybe) safer

src/aleph/netty/impl.clj
```diff
@@ -0,0 +1,18 @@
+(ns aleph.netty.impl
```
KingMob, line 1: Aren't `impl` namespaces primarily for protocol implementations?

test/aleph/classpath_test.clj
```diff
@@ -0,0 +1,42 @@
...
+      (require 'aleph.netty :reload)
+      (let [server (http/start-server
+                    (constantly {:body "ok"})
+                    {:port 9999})]
+        (with-handler :term
+          (.close ^java.io.Closeable server))
+        (.exec (Runtime/getRuntime) (format "kill -SIGTERM %s" (pid)))
+        (is @result)
+        (require 'aleph.netty.impl :reload)
+        (require 'aleph.netty :reload)))))
```
KingMob, line 42: This interning and re-requiring seems more complicated than it needs to be. Since we know `aleph.netty.impl/operation-complete` exists, we can use `alter-var-root` or `with-redefs` to temporarily change its root binding, then set it back when done. It's effectively equivalent, but with no need to reload anything. I'm not seeing any benefit to the macro/intern/reload setup over `with-redefs`, but please let me know if I'm missing something. (`with-redefs` is not multithread-safe, but neither is this, since concurrent tests would use the interned version.)

(Incidentally, I have a blog post coming out about in the next week or two highlighting a much better technique for exactly this sort of test redefinition, I'm just waiting on someone to get back to me.)

As I mentioned elsewhere, let's switch to the SIGINT over SIGTERM to be safer.

test/aleph/classpath_test.clj
```diff
@@ -0,0 +1,42 @@
...
+       (catch Throwable _ (do (.complete result false) d))))
+
+(defn pid
+  "Gets this process' PID."
+  []
+  (let [pid (.getName (ManagementFactory/getRuntimeMXBean))]
+    (->> pid
+         (re-seq #"[0-9]+")
+         (first)
+         (Integer/parseInt))))
```
KingMob, line 27: Does this work in all JVMs? I see for a JVM 10+ at least, there's a `.getPid()` method that might be better. Maybe we can use that if present, and fall back to this if not?

KingMob (changes requested):
I think we can make this much simpler. We shouldn't need macros for this, nor interning/reloading.

src/aleph/netty/impl.clj
```diff
@@ -0,0 +1,18 @@
...
+  (:import [io.netty.util.concurrent Future]))
+
+(defmacro operation-complete [^Future f d]
```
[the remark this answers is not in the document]
arnaudgeiser: With the macro, it ensures it's inlined into `wrap-future`.
But I'm overthinking there, let's rely on the JIT.

test/aleph/classpath_test.clj
arnaudgeiser: I will work on simplifying this test.

arnaudgeiser: > I'm not seeing any benefit to the macro/intern/reload setup over with-redefs, but please let me know if I'm missing something. (with-redefs is not multithread-safe, but neither is this, since concurrent tests would use the interned version.)

As long as we are not using a macro for `operation-complete` we can use `with-redefs`.

src/aleph/netty.clj
arnaudgeiser: Applying the following change makes the test fail with the initial issue : `ClassNotFoundException`

test/aleph/classpath_test.clj
arnaudgeiser: I'm using `ProcessHandler` when available now.

src/aleph/netty.clj
KingMob: Huh. Interesting

test/aleph/classloader_test.clj
```diff
@@ -0,0 +1,44 @@
...
+  (testing "classloader: ensure the class loader is always a DynamicClassLoader"
+    (let [result (CompletableFuture.)]
+      (with-redefs [netty/operation-complete (partial operation-complete result)]
+        (let [server (http/start-server
+                      (constantly {:body "ok"})
+                      {:port 9999})]
+          (with-handler :int
+            (.close ^java.io.Closeable server))
+          (.exec (Runtime/getRuntime) (format "kill -SIGINT %s" (pid)))
+          (is @result))))))
```
KingMob, line 44: If the test fails, it'll throw an exception in a Netty thread and the test will hang waiting for the CompletableFuture to be completed, right?

Maybe we should use the version of `deref` with a timeout, and give it ~30s to wrap up?

src/aleph/netty.clj
```diff
@@ -203,41 +200,48 @@
...
     (if (.isSuccess f)
       (d/success-deferred (.getNow f) nil)
       (let [d (d/deferred nil)
-            ;; workaround for the issue with executing RT.readString
-            ;; on a Netty thread that doesn't have appropriate class loader
-            ;; more information here:
-            ;; https://github.com/clj-commons/aleph/issues/365
-            class-loader (or (clojure.lang.RT/baseLoader)
-                             (clojure.lang.RT/makeClassLoader))]
+            bound-operation-complete (bound-fn* operation-complete)]
```
KingMob, line 239: We should keep a comment explaining it all. Since it's on a netty thread, the vars and classloader will be different

src/aleph/netty.clj
```diff
@@ -203,41 +200,48 @@
...
+(defn- operation-complete [^Future f d]
+  (cond
+     (.isSuccess f)
+     (d/success! d (.getNow f))
+
+     (.isCancelled f)
+     (d/error! d (CancellationException. "future is cancelled."))
+
+     (some? (.cause f))
+     (if (instance? java.nio.channels.ClosedChannelException (.cause f))
```
KingMob, line 226: import?

KingMob (approved):
LGTM, other than adding some final comments for future us to read.

Thanks for all your hard work on this!

src/aleph/netty.clj
arnaudgeiser: 7a3410784b28335f8687df42b7f776112f2265fb

src/aleph/netty.clj
arnaudgeiser: 95b67d4a66c100c7b045fc4553241c3bf7a38dbf

Can I ask you if this comment is enough?

KingMob:
@arnaudgeiser Merge whenever you're ready.

Fixes #603

PR #607, opened 2022-07-26
Fix assertions in aleph.tcp-ssl-test

DerGuteMoritz:
Some assertions were ineffective since they would be executed on a
different thread from the test runner. For example, deliberately
introducing a failing assertion wouldn't make the test fail. The bug
can also be observed in the test summary. When running
`test-ssl-echo`, it mentions only one assertion:

> Ran 1 assertions, in 1 test functions

With the patch applied this becomes:

> Ran 3 assertions, in 1 test functions

### DIFF
--- a/test/aleph/tcp_ssl_test.clj
+++ b/test/aleph/tcp_ssl_test.clj
@@ -12,25 +12,27 @@
 
 (set! *warn-on-reflection* false)
 
-(defn ssl-echo-handler
-  [s c]
-  (is (some? (:ssl-session c)) "SSL session should be defined")
-  (s/connect
-    ; note we need to inspect the SSL session *after* we start reading
-    ; data. Otherwise, the session might not be set up yet.
-    (s/map (fn [msg]
-             (is (= (.getSubjectDN ^X509Certificate ssl/client-cert)
-                   (.getSubjectDN ^X509Certificate (first (.getPeerCertificates (:ssl-session c))))))
-             msg)
-      s)
-    s))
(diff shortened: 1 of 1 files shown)


arnaudgeiser (approved):
Thanks a lot!

KingMob (approved):
I had a beautiful long comment, and GitHub ate it. I'm so pissed. 😡 

So, here's the much shorter version:

@DerGuteMoritz Thanks for finding this bug. The issue is not multi-threaded tests, but the fact that that the assertion in the handler takes place on a Netty thread, which doesn't participate in Clojure thread binding. The handler `is` assertion executes, but `clojure.test` can't track it properly because of missing test bindings.

We encountered a similar issue recently with closing callbacks. In general, Netty (and thus Aleph) doesn't play well with bindings. I checked all the other handler tests, and this is the only test that makes an assertion in the handler, so I don't think we need solve a bigger problem. Which is good, because I couldn't think of a general solution (ruled out bound-fn, threadfactory executors, manual binding frame copying, etc.)

👍🏻

DerGuteMoritz:
@KingMob Thanks for the clarification! I guess the best way to put it then is that assertions in any non-Clojure threads are a problem. I had indeed forgotten about binding conveyence of Clojure's thread facilities.

Good call on checking other tests, too! Should have thought of that myself :pray: 

And sorry about your lost comment, that's always very annoying. Channeling some positive energy over to you so that it won't ruin your day too much :smile:

PR #608, opened 2022-07-27
Improve TCP+SSL server connection handling

DerGuteMoritz:
When starting an TCP+SSL server, the handler would be called when the
underlying TCP connection is established but before the SSL handshake
has taken place.  At that point, the client's `:ssl-session` value is
still a placeholder (e.g. `.isValid` returns `false`, the cipher suite
is reported as `"SSL_NULL_WITH_NULL_NULL"` and the peer principal is
not available). Also, the stream is not ready for use, so any writes
would be buffered until after the handshake is complete. Since there
is no way to be notified of the handshake completion, the existing
test case worked around this by looking up the SSL session only after
having received a message for the first time which implies that the
session has been successfully established since. However, this is a
pretty awkward API and only works for protocols in which the client
actually is the first to transmit anything. There's also a bit of a
gotcha: The value of `:ssl-session` mutably changes to a different
value after the handshake, so holding on to the initial placeholder
value and only starting to use it later won't do the trick.

This patch changes the semantics so that the handler is only invoked
after the session was successfully established. While technically a
breaking change, it appears unlikely that any existing users would
rely on the original behavior since the placeholder SSL session value
is pretty much useless. The new behavior should be more in line with
user expectation and arguably is more correct (e.g. the handler
doesn't have to wait for I/O before being able to peruse the SSL
session info).

----

FWIW I find the API for looking up `:ssl-session` a bit questionable: It's using `def-derived-map` to dispatch to a method call under the hood whose return value's identity may change over time which is pretty non-obvious and probably not how `def-derived-map` is intended to be used. Then again, with my patch applied, this shouldn't be the case anymore, so I chose to keep it minimally invasive. But something to keep in mind ...

### DIFF
--- a/src/aleph/tcp.clj
+++ b/src/aleph/tcp.clj
@@ -82,7 +82,7 @@
    |:---|:-----
    | `port` | the port the server will bind to.  If `0`, the server will bind to a random port.
    | `socket-address` | a `java.net.SocketAddress` specifying both the port and interface to bind to.
-   | `ssl-context` | an `io.netty.handler.ssl.SslContext` object. If a self-signed certificate is all that's required, `(aleph.netty/self-signed-ssl-context)` will suffice.
+   | `ssl-context` | an `io.netty.handler.ssl.SslContext` object. If given, the server will only accept SSL connections and call the handler once the SSL session has been successfully established. If a self-signed certificate is all that's required, `(aleph.netty/self-signed-ssl-context)` will suffice.
    | `bootstrap-transform` | a function that takes an `io.netty.bootstrap.ServerBootstrap` object, which represents the server, and modifies it.
    | `pipeline-transform` | a function that takes an `io.netty.channel.ChannelPipeline` object, which represents a connection, and modifies it.
    | `raw-stream?` | if true, messages from the stream will be `io.netty.buffer.ByteBuf` objects rather than byte-arrays.  This will minimize copying, but means that care must be taken with Netty's buffer reference counting.  Only recommended for advanced users."
--- a/src/aleph/tcp.clj
+++ b/src/aleph/tcp.clj
@@ -13,7 +13,10 @@
     [io.netty.channel
      Channel
      ChannelHandler
-     ChannelPipeline]))
+     ChannelHandlerContext
+     ChannelPipeline]
+    [io.netty.handler.ssl
+     SslHandler]))
 
 (p/def-derived-map TcpConnection [^Channel ch]
   :server-name (netty/channel-server-name ch)
@@ -25,7 +28,16 @@
 
 (defn- ^ChannelHandler server-channel-handler
   [handler {:keys [raw-stream?] :as options}]
-  (let [in (atom nil)]
+  (let [in (atom nil)
+        call-handler (fn [^ChannelHandlerContext ctx]
+                       (let [ch (.channel ctx)]
+                         (handler
+                          (doto
+                              (s/splice
+                               (netty/sink ch true netty/to-byte-buf)
+                               (reset! in (netty/source ch)))
+                            (reset-meta! {:aleph/channel ch}))
+                          (->TcpConnection ch))))]
     (netty/channel-inbound-handler
 
       :exception-caught
@@ -40,15 +52,20 @@
 
       :channel-active
       ([_ ctx]
-        (let [ch (.channel ctx)]
-          (handler
-            (doto
-              (s/splice
-                (netty/sink ch true netty/to-byte-buf)
-                (reset! in (netty/source ch)))
-              (reset-meta! {:aleph/channel ch}))
-            (->TcpConnection ch)))
-        (.fireChannelActive ctx))
+       (if-let [^SslHandler ssl-handler (-> ctx .pipeline (.get SslHandler))]
+         (-> ssl-handler
+             .handshakeFuture
+             netty/wrap-future
+             (d/on-realized (fn [_]
+                              (call-handler ctx))
+                            ;; No need to handle errors here since
+                            ;; the SSL handler will terminate the
+                            ;; whole pipeline by throwing a
+                            ;; `javax.net.ssl.SSLHandshakeException`
+                            ;; in this case anyway.
+                            (fn [_])))
+         (call-handler ctx))
+       (.fireChannelActive ctx))
 
       :channel-read
       ([_ ctx msg]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+### Unreleased
+
+* Make TCP+SSL server call handler only after successful SSL handshake
+
 ### 0.5.0
 
 * Add initial clj-kondo hooks
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -2,6 +2,7 @@
   (:require
    [aleph.flow :as flow]
    [aleph.http :as http]
+   [aleph.http.core :as http.core]
    [aleph.netty :as netty]
    [aleph.ssl :as ssl]
    [aleph.tcp :as tcp]
@@ -18,7 +19,8 @@
    (io.netty.handler.codec.http HttpMessage)
    (java.io File)
    (java.util.concurrent TimeoutException)
-   (java.util.zip GZIPInputStream ZipException)))
+   (java.util.zip GZIPInputStream ZipException)
+   (javax.net.ssl SSLSession)))
 
 ;;;
 
@@ -259,6 +261,35 @@
               (:body @(http-put client-url
                                 {:body (io/file "test/file.txt")
                                  :pool client-pool}))))))))
+
+(defn ssl-session-capture-handler [ssl-session-atom]
+  (fn [req]
+    (reset! ssl-session-atom (http.core/ring-request-ssl-session req))
+    {:status 200 :body "ok"}))
+
+(deftest test-ssl-session-access
+  (let [ssl-session (atom nil)]
+    (with-handler-options
+      (ssl-session-capture-handler ssl-session)
+      default-ssl-options
+      (is (= 200 (:status @(http-get (str "https://localhost:" port)))))
+      (is (some? @ssl-session))
+      (when-let [^SSLSession s @ssl-session]
+        (is (.isValid s))
+        (is (not (str/includes? "NULL" (.getCipherSuite s))))))))
+
+(deftest test-ssl-with-plain-client-request
+  (let [ssl-session (atom nil)]
+    (with-handler-options
+      (ssl-session-capture-handler ssl-session)
+      default-ssl-options
+      ;; Note the intentionally wrong "http" scheme here
+      (is (some-> (http-get (str "http://localhost:" port))
+                  (d/catch identity)
+                  deref
+                  ex-message
+                  (str/includes? "connection was closed")))
+      (is (nil? @ssl-session)))))
 
 (deftest test-invalid-body
   (let [client-url (str "http://localhost:" port)]
--- a/test/aleph/tcp_ssl_test.clj
+++ b/test/aleph/tcp_ssl_test.clj
@@ -16,14 +16,8 @@
 
 (defn ssl-echo-handler [ssl-session]
   (fn [s c]
-    (s/connect
-     ;; note we need to capture the SSL session *after* we start
-     ;; reading data. Otherwise, the session might not be set up yet.
-     (s/map (fn [msg]
-              (reset! ssl-session (:ssl-session c))
-              msg)
-            s)
-     s)))
+    (reset! ssl-session (:ssl-session c))
+    (s/connect s s)))
 
 (deftest test-ssl-echo
   (let [ssl-session (atom nil)]
@@ -38,3 +32,18 @@
         (is (some? @ssl-session) "SSL session should be defined")
         (is (= (.getSubjectDN ^X509Certificate ssl/client-cert)
                (.getSubjectDN ^X509Certificate (first (.getPeerCertificates @ssl-session)))))))))
+
+(deftest test-failed-ssl-handshake
+  (let [ssl-session (atom nil)]
+    (with-server (tcp/start-server (ssl-echo-handler ssl-session)
+                                   {:port 10001
+                                    :ssl-context ssl/server-ssl-context})
+      (let [c @(tcp/client {:host "localhost"
+                            :port 10001
+                            :ssl-context (netty/ssl-client-context
+                                          ;; Note the intentionally wrong private key here
+                                          {:private-key ssl/server-key
+                                           :certificate-chain (into-array X509Certificate [ssl/client-cert])
+                                           :trust-store (into-array X509Certificate [ssl/ca-cert])})})]
+        (is (nil? @(s/take! c)))
+        (is (nil? @ssl-session) "SSL session should be undefined")))))


DerGuteMoritz:
Just noticed that the docstring on the `ssl-context` parameter was a bit lacking. Added a commit with an improvement attempt.

KingMob (changes requested):
Good catch!

A few things:
1. "... there is no way to be notified of the handshake completion" Are you sure about this? I see there's a `io.netty.handler.ssl.SslHandler#handshakeFuture` method. Have you checked out `io/netty/example/securechat` in the Netty code? The code in `io.netty.example.securechat.SecureChatServerHandler#channelActive` suggests that the presence of the SslHandler in the pipeline is insufficient, and you may still need to wait on the handshakeFuture to complete.
2. Is this not also a problem for the HTTP code as well as TCP? We should fix both, if so.
3. Oleksii worked on a long-lived branch with a lot of improvements that he never finished and merged, and there's probably some SSL stuff worth looking at (maybe 8c94cf33fc85b357851f99397918696afe4f5dbc?)
4. Minor: It may be preferable to retrieve the SSLHandler in the pipeline via `(.get SslHandler)` instead of by name, like the way `channel-ssl-session` does it. This should be resistant to renaming, and there shouldn't be more than one SSLHandler, anyway.
5. Minor: since we lack the ability to mix secure and insecure connections, we should consider making Netty's `OptionalSslHandler` an option, so people can serve both secure and insecure content if they like.
6. Minor: I don't think `def-derived-map` is a big concern. It's reflecting the underlying mutability of its values. If we use mutable Java objects, we run the risk of stuff changing. (FWIW, I find it helpful to think of derived-maps as more like a "reactive" record pattern than a map.)

DerGuteMoritz:
Hi,

thanks for the quick and thorough feedback :bow: 

> 1. "... there is no way to be notified of the handshake completion" Are you sure about this? I see there's a `io.netty.handler.ssl.SslHandler#handshakeFuture` method. Have you checked out `io/netty/example/securechat` in the Netty code? The code in `io.netty.example.securechat.SecureChatServerHandler#channelActive` suggests that the presence of the SslHandler in the pipeline is insufficient, and you may still need to wait on the handshakeFuture to complete.

Ah no, by this I meant that there was no way *in aleph* to be notified about this. Netty has various ways for this, cf. https://netty.io/4.1/api/io/netty/handler/ssl/SslHandler.html -- I ended up using the user triggered event since that was easiest to integrate.

> 2. Is this not also a problem for the HTTP code as well as TCP? We should fix both, if so.

AFAICT the HTTP server doesn't expose the SSL session object to the handler. Also, unlike the TCP server, the HTTP server invokes the handler when `:channel-read` is fired at which point the handshake has to have happened. And since in HTTP the client always transmits first, the response buffering issue I mention doesn't apply either. So I think we're fine here.

> 3. Oleksii worked on a long-lived branch with a lot of improvements that he never finished and merged, and there's probably some SSL stuff worth looking at (maybe [8c94cf3](https://github.com/clj-commons/aleph/commit/8c94cf33fc85b357851f99397918696afe4f5dbc)?)

Thanks, I will take a look!

> 4. Minor: It may be preferable to retrieve the SSLHandler in the pipeline via `(.get SslHandler)` instead of by name, like the way `channel-ssl-session` does it. This should be resistant to renaming, and there shouldn't be more than one SSLHandler, anyway.

Good call, will change accordingly.

> 5. Minor: since we lack the ability to mix secure and insecure connections, we should consider making Netty's `OptionalSslHandler` an option, so people can serve both secure and insecure content if they like.

I haven't come across this one so far. I suggest we file an issue for this as a potential improvement? FWIW, I would discourage mixing encrypted and unencrypted communication on the same TCP port, this seems to invite a lot of potential for attack vectors (see e.g. https://en.wikipedia.org/wiki/Opportunistic_TLS#Weaknesses_and_mitigations). But of course, there are legacy protocols which require this so making it an option could be worthwhile regardless.

> 6. Minor: I don't think `def-derived-map` is a big concern. It's reflecting the underlying mutability of its values. If we use mutable Java objects, we run the risk of stuff changing. (FWIW, I find it helpful to think of derived-maps as more like a "reactive" record pattern than a map.)

Right, I guess I was concerned about this is because the docstring for `start-server` just says "a map containing information about the client" which reads to a Clojurian like a plain old map. Unknowingly, I captured the value of `:ssl-session` right when the handler was invoked and was severly confused for a while until I found that test case which hinted at the fact that the underlying value might change. Anyway, as mentioned, with my patch applied this shouldn't happen anymore and will hopefully save future users some trouble :sweat_smile: 

BTW, do you want me to add a changelog entry for this or do you prefer to write them up yourself upon release?

DerGuteMoritz:
> * Oleksii worked on a long-lived branch with a lot of improvements that he never finished and merged, and there's probably some SSL stuff worth looking at (maybe [8c94cf3](https://github.com/clj-commons/aleph/commit/8c94cf33fc85b357851f99397918696afe4f5dbc)?)

OK the main patch of that PR (dac2852c73c8ceeecaa15549c9de08dbcd530efd) adds a more convenient API for constructing a Netty `SslContext` which is nice but unrelated to my patch. Other than that, the relevant code on the `1.0.0` branch (assuming this is the one you were referring to) looks unchanged.

KingMob:
> I meant that there was no way in aleph to be notified about this. Netty has various ways for this, cf. 

I'm confused by this. Aleph is built on Netty and Manifold, why couldn't you use the handshake promise? Wouldn't it be `(some-> ch (.pipeline) (.get SslHandler))`, like from `channel-ssl-session`? Checking the user-event SslHandshakeCompletionEvent is fine, but if there's a reason to listen for the handshake promise instead, it seems doable to me.

Also, looking at Netty examples, if you use SslHandshakeCompletionEvent, you still need to call `.isSuccess()` on it to verify the handshake didn't fail.

> AFAICT the HTTP server doesn't expose the SSL session object to the handler. Also, unlike the TCP server, the HTTP server invokes the handler when :channel-read is fired at which point the handshake has to have happened. And since in HTTP the client always transmits first, the response buffering issue I mention doesn't apply either. So I think we're fine here.

If you think it's not an issue on the HTTP side, can we get a quick test verifying that? What you say makes sense, but we have virtually no HTTPS SSL tests, so it would be nice to know for sure.

> FWIW, I would discourage mixing encrypted and unencrypted communication on the same TCP port, this seems to invite a lot of potential for attack vectors

Sorry, I wasn't clear on this point. I'm not suggesting we handle secure and insecure connections on the same port, I was more interested in being able to reuse the same code (since other than TLS, the HTTP parts will be the same). OTOH, the way the HTTP/S pipeline works, maybe this is a complete non-issue.

You may want to grep the git history for other ssl/tls stuff on Oleksii's unmerged branch, could still be something of interest there, fyi.

Adding a line or two to the changelog line is fine. I usually take care of it when making a release, though, so you can skip it if you like.

DerGuteMoritz:
> I'm confused by this. Aleph is built on Netty and Manifold, why couldn't you use the handshake promise? Wouldn't it be `(some-> ch (.pipeline) (.get SslHandler) (.handshakeFuture))`, like from `channel-ssl-session`? Checking the user-event SslHandshakeCompletionEvent is fine, but if there's a reason to listen for the handshake promise instead, it seems doable to me.

I guess I was holding it wrong when I first tried it - works indeed and is arguably a bit more idiomatic, too :+1: 


> Also, looking at Netty examples, if you use SslHandshakeCompletionEvent, you still need to call `.isSuccess()` on it to verify the handshake didn't fail.

Oooouch, thanks a lot, what a blunder :see_no_evil: This alone makes it worthwhile to switch to `handshakeFuture` since the success check is implied :sweat_drops: Pushed a change to that effect!

This made me wonder what exactly happens when the handshake fails. Turns out it results in a pretty noisy exception getting logged. As it happens, the merge commit of the PR you cited above also contains [a patch to improve this a bit](https://github.com/clj-commons/aleph/pull/546/commits/f2c1b030b7104f79ae0adea3d478a1752140eb58#diff-58b3da8a70a7d700653c11e56630106278a1ed8451ccdf59c3cb0b661a27ea22R213-R219). How about we pick this over? The patch only touches the HTTP server, I would then adjust it for the TCP server accordingly.  We could also allow passing in a handler for these similar to how the HTTP server accepts a `rejected-handler`... But I don't think this is too useful other than perhaps for monitoring purposes, though people could then just as well scan the log output for these errors. WDYT?
 
I also added a `test-failed-ssl-handshake` case to `aleph.tcp-ssl-test` for good measure.

> If you think it's not an issue on the HTTP side, can we get a quick test verifying that? What you say makes sense, but we have virtually no HTTPS SSL tests, so it would be nice to know for sure.

Alright, will see what I can do!


> Sorry, I wasn't clear on this point. I'm not suggesting we handle secure and insecure connections on the same port, I was more interested in being able to reuse the same code (since other than TLS, the HTTP parts will be the same). OTOH, the way the HTTP/S pipeline works, maybe this is a complete non-issue.

Hmm still not quite sure I follow :sweat_smile: Do you mean being able to use the same Aleph server instance (and thus Netty pipeline) for listening on both port 80 for plain and 443 for secure connections?


> You may want to grep the git history for other ssl/tls stuff on Oleksii's unmerged branch, could still be something of interest there, fyi.

Did so now! It seems like the two commits from the merge commit you referenced above are the only ones pertaining to SSL improvements, really. These seem worthwhile picking over IMHO, nice quality of life improvements with small chance of breaking things. Looks like they've accumulated some conflicts, though. If you like, I'll prepare a separate PR with rebases of them onto latest `master`. I would also add the improved error logging for the TCP server mentioned above.

 
> Adding a line or two to the changelog line is fine. I usually take care of it when making a release, though, so you can skip it if you like.

Alrighty, pushed a commit with a suggestion for your discretion :pray:

DerGuteMoritz:
> If you think it's not an issue on the HTTP side, can we get a quick test verifying that? What you say makes sense, but we have virtually no HTTPS SSL tests, so it would be nice to know for sure.

OK just pushed a commit with two new HTTPS server tests which should demonstrate that the handler doesn't get called before successul handshake completion. WDYT?

arnaudgeiser:
> AFAICT the HTTP server doesn't expose the SSL session object to the handler. Also, unlike the TCP server, the HTTP server invokes the handler when :channel-read is fired at which point the handshake has to have happened. And since in HTTP the client always transmits first, the response buffering issue I mention doesn't apply either. So I think we're fine here.

I reached the same conclusion.
On TCP -> `:channel-active`
On HTTP -> `:channel-read`

The tests prove we weren't waiting for the handshake to happen.
On the TCP side, it looks good to me.

On the HTTP side, the tests you add are great additions.
I would fix the type hints though.

KingMob:
> As it happens, the merge commit of the PR you cited above also contains [a patch to improve this a bit](https://github.com/clj-commons/aleph/pull/546/commits/f2c1b030b7104f79ae0adea3d478a1752140eb58#diff-58b3da8a70a7d700653c11e56630106278a1ed8451ccdf59c3cb0b661a27ea22R213-R219). How about we pick this over?

Go for it. If you can safely cherry-pick, that's fine. Oleksii's branch was very long-lived, though, and there's no guarantees. Cut'n'paste works too.

I don't think we should support a handler fn to respond to handshake errors. Let's just log a warning.

Thanks for the tests!

> Do you mean being able to use the same Aleph server instance (and thus Netty pipeline) for listening on both port 80 for plain and 443 for secure connections?

No, I just want to be able to maintain the same code for both, but it looks like that's largely true already, since the only real change is an addition in the pipeline.

It _would_ be nice to set up a way for Aleph to serve both secure/insecure content from both ports 80 and 443, rather than start two servers, but that's kind of a moot point going forward, since HTTP/2 and 3 effectively require TLS. Maybe we could make a wrapper `start-both-secure-and-insecure-servers` fn, but... ehhh.

src/aleph/tcp.clj
```diff
@@ -40,15 +52,13 @@
...
-                (netty/sink ch true netty/to-byte-buf)
-                (reset! in (netty/source ch)))
-              (reset-meta! {:aleph/channel ch}))
-            (->TcpConnection ch)))
-        (.fireChannelActive ctx))
+       (if-let [^SslHandler ssl-handler (-> ctx .pipeline (.get SslHandler))]
+         (-> ssl-handler
+             .handshakeFuture
+             netty/wrap-future
+             (d/chain' (fn [_] (call-handler ctx))))
```
KingMob, line 59: @arnaudgeiser @DerGuteMoritz 

We have a slight bug here.

First, Moritz, if you don't need the deferred's value, you probably want to use `on-realized`, and not `chain/chain'`.

Second, it's still not checking for success here. Or rather, `wrap-future` IS, but deceptively, which brings me to the problem. Technically, it's not a bug, since the code works, but it _is_ confusing and wasteful. `wrap-future` uses `.isSuccess` in a way where it really wants to call `.isDone`. 

If the future succeeded, it immediately makes a success-deferred. If it failed, it attaches `operation-complete` as a listener. Note that if the future was cancelled or completed with an exception, it will then call `operation-complete` immediately, which may return an error-deferred. It's wasted work, since it may set up `operation-complete` to run on another thread, though.

More importantly, the call to `.isSuccess` isn't guarding anything for you, since you could still get an error-deferred. `on-realized` will force you to handle both success and error scenarios.

We should probably change `wrap-future` to use `isDone`, and immediately invoke `operation-complete` if true. But not in this PR.

KingMob (changes requested):
Still need to handle handshake failure

src/aleph/tcp.clj
DerGuteMoritz: Oof, tricky business! But I think I successfully applied `wrap-head` to your comment now :sweat_smile: Changing `wrap-future` the way you suggest indeed sounds like a reasonable improvement :+1:

Changing to `d/on-realized` is a good idea, too. Not only does it make the intent clearer, but I think my current implementation would also generate extra overhead if the handler happens to return a deferred. My only question is what to do with the error handler. As you noticed, the code already works without explicit handling of the error because the SSL handler terminates the pipeline in this case anyway (by throwing a `javax.net.ssl.SSLHandshakeException`). I'm passing a (commented) noop function now, just like `d/listener` does in the single arity case. Lemme know if you have a different understanding or better idea!

DerGuteMoritz:
> No, I just want to be able to maintain the same code for both, but it looks like that's largely true already, since the only real change is an addition in the pipeline.

Ah, alright, gotcha. Right, agreed, this is already the case.
 
> It _would_ be nice to set up a way for Aleph to serve both secure/insecure content from both ports 80 and 443, rather than start two servers, but that's kind of a moot point going forward, since HTTP/2 and 3 effectively require TLS. Maybe we could make a wrapper `start-both-secure-and-insecure-servers` fn, but... ehhh.

Yeah doesn't really seem worth it these days. Should somebody need it, they can always file an issue :smile:

arnaudgeiser (approved):
I would go with this even if it's perfectible.

Note however that Aleph's TCP/UDP servers seem to be very little used. So, with the TLS, this seems unlikely that anyone will even noticed if we introduced some kind of overhead.

DerGuteMoritz:
> Note however that Aleph's TCP/UDP servers seem to be very little used. So, with the TLS, this seems unlikely that anyone will even noticed if we introduced some kind of overhead.

Well I ran into the issue here because I want to use it :smile: So scratching my own itch here. I might land some more improvements in the near future!

src/aleph/tcp.clj
KingMob: I think we should log it as a warning.

KingMob:
Other than logging SSL exceptions as a warning, I have no other comments. I'm approving in the interest of time.

I will say, you might want to force-push less. I love force-pushes, but they can make reviews harder, and they definitely make collaboration on a branch harder-to-impossible. My usual strategy is to force-push only before I make a PR, or only AFTER everyone's done reviewing.

KingMob (approved):

src/aleph/tcp.clj
DerGuteMoritz: This is actually already covered via the general exception handler. Results in a log like this:

```
[aleph-netty-server-event-pool-2] WARN aleph.http.server - error in HTTP server
io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: Invalid CertificateVerify signature
[ ... stack trace omitted ... ]
```

See also https://github.com/clj-commons/aleph/pull/611

In other words: Logging it here, too, would result in a redundant warning for the same handshake failure. Would you still prefer to have it?

DerGuteMoritz:
> Other than logging SSL exceptions as a warning, I have no other comments.

See https://github.com/clj-commons/aleph/pull/608#discussion_r938812336

> I will say, you might want to force-push less. I love force-pushes, but they can make reviews harder, and they definitely make collaboration on a branch harder-to-impossible. My usual strategy is to force-push only before I make a PR, or only AFTER everyone's done reviewing.

FWIW, you can see the net changes introduced by a force push using the "Compare" button right next to the force push notification here on GitHub. E.g. [see my latest one here](https://github.com/clj-commons/aleph/compare/2e7f68e9eb3ecbee105c3612bcb3bcd0dd31e971..dd03fee4b08449514a2d3aafdef4903d881ceb8f). That being said, I'm happy to switch to your preferred strategy from now on :+1:

KingMob:
> FWIW, you can see the net changes introduced by a force push using the "Compare" button right next to the force push notification here on GitHub.

@DerGuteMoritz Sure, sure, but it breaks lots of other things. I lose the context of what I'm looking at, since it's a diff between SHAs without commit messages, and I can't even look up the history locally, since I'm guaranteed not to have some of the SHAs. I can't collaborate on a branch without sharing the same history (although we have other issues we need to straighten out there). And in particular, it takes away the "See changes since last review" button that I like to use 😄 

Believe me, I love to rewrite git history to clean it up, but rewriting is fundamentally antisocial, since nobody else shares your view of the commits.

KingMob:
@DerGuteMoritz No worries, I didn't think you were advocating for The One True Way, or anything. And thanks for all your hard work!

PR #611, opened 2022-07-30
Tone down SSL handshake error logging

DerGuteMoritz:
This is a cherry-pick of f2c1b030b7104f79ae0adea3d478a1752140eb58 from the `1.0.0` branch plus the same change for TCP server and client.

As per https://github.com/clj-commons/aleph/pull/608#issuecomment-1200103587

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -108,7 +108,7 @@
     (s/put! response-stream ex)
 
     ;; when SSL handshake failed
-    (http/ssl-handshake-error? ex)
+    (netty/ssl-handshake-error? ex)
     (let [^Throwable handshake-error (.getCause ^Throwable ex)]
       (s/put! response-stream handshake-error))
 
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -100,6 +100,21 @@
           nil))
       (no-url req))))
 
+(defn exception-handler [ctx ex response-stream]
+  (cond
+    ;; could happens when io.netty.handler.codec.http.HttpObjectAggregator
+    ;; is part of the pipeline
+    (instance? TooLongFrameException ex)
+    (s/put! response-stream ex)
+
+    ;; when SSL handshake failed
+    (http/ssl-handshake-error? ex)
+    (let [^Throwable handshake-error (.getCause ^Throwable ex)]
+      (s/put! response-stream handshake-error))
+
+    (not (instance? IOException ex))
+    (log/warn ex "error in HTTP client")))
+
 (defn raw-client-handler
   [response-stream buffer-capacity]
   (let [stream (atom nil)
@@ -117,8 +132,7 @@
 
       :exception-caught
       ([_ ctx ex]
-        (when-not (instance? IOException ex)
-          (log/warn ex "error in HTTP client")))
+        (exception-handler ctx ex response-stream))
 
       :channel-inactive
       ([_ ctx]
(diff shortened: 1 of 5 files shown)


src/aleph/http/client.clj
```diff
@@ -100,6 +100,21 @@
...
+    (instance? TooLongFrameException ex)
+    (s/put! response-stream ex)
+
+    ;; when SSL handshake failed
+    (netty/ssl-handshake-error? ex)
+    (let [^Throwable handshake-error (.getCause ^Throwable ex)]
+      (s/put! response-stream handshake-error))
+
+    (not (instance? IOException ex))
+    (log/warn ex "error in HTTP client")))
```
arnaudgeiser, line 116: On the error-handling branch, Oleksii use to close the channel.
Maybe we should do it on this PR [1]

[1] : https://github.com/clj-commons/aleph/blob/error-handling-1.0.0/src/aleph/http/client.clj#L122

KingMob: I def agree we should be doing more than logging a warning. However, I'm not sure we should do as Oleksii did, and send the exception to the client, at least not by default. It's handy for debugging, but a bad idea for security.

KingMob: @DerGuteMoritz If you want to close the channel, I can wait. Otherwise, is there anything else to this? It's ready, right?

DerGuteMoritz: Hmm yeah closing the channel seems reasonable. I wonder if this means that the current code leaks connections on exception? In any case, will push a change to that effect since it certainly shouldn't hurt either.

> However, I'm not sure we should do as Oleksii did, and send the exception to the client, at least not by default. It's handy for debugging, but a bad idea for security.

Note that the code in question here is the HTTP *client* in which case I think forwarding the exception to the caller is fine? Your point is certainly true wrt the HTTP *server*'s default exception handler which indeed echoes back exceptions verbatim with no way to disable this. In fact, [there is another open PR addressing exactly this issue](https://github.com/clj-commons/aleph/pull/521).

DerGuteMoritz: Pushed! Should be good to go then... :crossed_fingers: Another sanity-check welcome, of course.

src/aleph/tcp.clj
```diff
@@ -30,8 +30,14 @@
...
-        (when-not (instance? IOException ex)
-          (log/warn ex "error in TCP server")))
+       (cond
+         ;; do not need to log an entire stack trace when SSL handshake failed
+         (netty/ssl-handshake-error? ex)
+         (log/warn "SSL handshake failure:"
+                   (.getMessage ^Throwable (.getCause ^Throwable ex)))
+
+         (not (instance? IOException ex))
+         (log/warn ex "error in TCP server")))
```
arnaudgeiser, line 40: Any reason the channel is closed everywhere but here?

DerGuteMoritz: [The commit I picked this from](https://github.com/clj-commons/aleph/commit/7aedd987acc2ff4d914c74243132be86c40c68c9) only touched the client handlers. Looks like the `error-handling-1.0.0` branch also has [a commit which does a similar thing for server handlers](https://github.com/clj-commons/aleph/commit/54954b6e8fc05e840154d3bc14ffd5d0bf688ae4). We could look into porting this, too, but I'm starting to feel this PR's scope starts creeping a bit too much. All I wanted was less noisy SSL handshake errors and what I got was the whole jungle :smile: How about I drop the `netty/close` commit again and we make a separate PR for salvaging the good parts from `error-handling-1.0.0`?

arnaudgeiser: > How about I drop the netty/close commit again and we make a separate PR for salvaging the good parts from error-handling-1.0.0?

Ok, let's handle the `netty/close` on another PR.

arnaudgeiser (approved):

src/aleph/tcp.clj
DerGuteMoritz: Just force-pushed an update to drop the commit :+1:

src/aleph/http/client.clj
KingMob: Yes, you're right, the _client_ should get exceptions, the _server_ should not send them.

The connections should be a dynamic pool. If too many are closed, more will be created as necessary, up to the limits.

PR #612, opened 2022-07-30
Improved API for SSL context management (server and client)

DerGuteMoritz:
This is a cherry-pick of dac2852c73c8ceeecaa15549c9de08dbcd530efd from the `1.0.0` adjusted to latest `master` and with some additions and corrections to docstrings.

As per https://github.com/clj-commons/aleph/pull/608#issuecomment-1200103587

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -34,12 +34,15 @@
      NioSocketChannel
      NioDatagramChannel]
     [io.netty.handler.ssl
+     ClientAuth
      SslContext
      SslContextBuilder
-     SslHandler]
+     SslHandler
+     SslProvider]
     [io.netty.handler.codec DecoderException]
     [io.netty.handler.ssl.util
-     SelfSignedCertificate InsecureTrustManagerFactory]
+     SelfSignedCertificate
+     InsecureTrustManagerFactory]
     [io.netty.resolver
      AddressResolverGroup
      NoopAddressResolverGroup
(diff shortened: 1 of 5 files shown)


src/aleph/netty.clj
```diff
@@ -725,63 +728,192 @@
 
 ;;;
 
+(defn coerce-ssl-provider [provider]
+  (case provider
+    :jdk SslProvider/JDK
+    :openssl SslProvider/OPENSSL
+    :openssl-refcnt SslProvider/OPENSSL_REFCNT))
+
+(set! *warn-on-reflection* false)
```
KingMob, line 737: This should go at the top, right after the `ns`

src/aleph/netty.clj
```diff
@@ -725,63 +728,192 @@
...
+              trust-store
+              ssl-provider
+              ciphers
+              protocols
+              session-cache-size
+              session-timeout]}]
+     (let [^SslContextBuilder builder (SslContextBuilder/forClient)
+           certificate-chain' (if-not (sequential? certificate-chain)
+                               certificate-chain
+                               (into-array X509Certificate certificate-chain))]
```
KingMob, line 779: Indentation is off. Can you fix it in all the added code?

src/aleph/netty.clj
KingMob: Wait, I just reread this. It's being set to false temporarily. We should be able to set the necessary hints to avoid reflection.

E.g., something like `^"[Ljava.security.cert.X509Certificate;"` for X509Certificate arrays.

src/aleph/netty.clj
```diff
@@ -959,7 +1091,7 @@ initialize an DnsAddressResolverGroup instance.
       epoll?
       nil))
   ([pipeline-builder
-    ^SslContext ssl-context
+    ssl-context
```
KingMob, line 1094: Was the type hint removed because sometimes it's a Builder?

KingMob (changes requested):
Please fix indentation. Also, can we not add type hints?

Otherwise, LGTM

src/aleph/netty.clj
arnaudgeiser: Maybe it could be also applied on all namespaces on the project.clj.

arnaudgeiser (approved):
Fine for me.
Regarding TLS, an improvement would be to support `javax.net.ssl SSLContext` out of the box.
Some people on the Clojure community are still using heavily `less-awful-ssl`.

For the record, here are the parameters supported by `hato` when creating a ssl-context:

```
ssl-context Sets the SSLContext. If not specified, uses the default (SSLContext/getDefault). Accepts:

    a map of :keystore :keystore-pass :trust-store :trust-store-pass, :insecure?. See client authentication examples for more details.
    an [javax.net.ssl.SSLContext](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/net/ssl/SSLContext.html)
```
https://github.com/gnarroway/hato

And for telex

```
:exoscale.telex.client/ssl-context: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/net/ssl/SSLContext.html (do yourself an favor an use less-awful-ssl)
```

https://github.com/exoscale/telex

src/aleph/netty.clj
arnaudgeiser: I would remove all the `(set! *warn-on-reflection* false)`, most of them should be fixable though.

DerGuteMoritz: Oof, that was trickier than expected :sweat_smile:  All done now. Perhaps there is a less pedestrian way to solve this?

This even turned up a bug: The code was calling `.provider`  while the method is actually called `.sslProvider`.

I also noticed that passing in a  `javax.net.ssl.TrustManager` as `trust-store` was missing, so added that for good measure.

DerGuteMoritz:
> Regarding TLS, an improvement would be to support `javax.net.ssl SSLContext` out of the box.

@arnaudgeiser Seems like this should be doable via https://netty.io/4.1/api/io/netty/handler/ssl/JdkSslContext.html which one could then pass as `:ssl-context`. Some builtin support for this in Aleph would be more convenient, of course. Should be easy to do now by adding it to `coerce-ssl-context`. Ideally with a test case, of course... I suggest we make this a separate issue?

arnaudgeiser:
> I suggest we make this a separate issue?

Yes, obviously.
Just surprised it wasn't supported already.

src/aleph/netty.clj
```diff
@@ -725,63 +728,192 @@
...
+
+(let [cert-array-class (class (into-array X509Certificate []))]
+  (defn- check-ssl-args! [private-key certificate-chain]
+    (when-not (or
+               (and (instance? File private-key)
+                    (instance? File certificate-chain))
+               (and (instance? InputStream private-key)
+                    (instance? InputStream certificate-chain))
+               (and (instance? PrivateKey private-key)
+                    (instance? cert-array-class certificate-chain)))
```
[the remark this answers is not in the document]
arnaudgeiser: Ah yes, good point.
I was just surprised by the choice of those combinations.

DerGuteMoritz:
@KingMob FYI: I'll resolve the conflicts (and squash where appropriate) once you've approved my latest changes!

src/aleph/netty.clj
KingMob: > Perhaps there is a less pedestrian way to solve this?

@DerGuteMoritz Generally, Clojure has crappy support for arrays, and Java has crappy support for array types, so... ¯\\\_(ツ)_/¯

One way you can do this is to use a snippet like `(println (str "^\"" (pr-str (class (make-array SomeClass 1))) "\""))` to get the array type hint for SomeClass.

src/aleph/netty.clj
```diff
@@ -728,26 +729,31 @@
...
+    (cond (instance? File trust-store)
+          (.trustManager builder ^File trust-store)
+          (instance? InputStream trust-store)
+          (.trustManager builder ^InputStream trust-store)
+          (instance? TrustManager trust-store)
+          (.trustManager builder ^TrustManager trust-store)
+          (instance? TrustManagerFactory trust-store)
+          (.trustManager builder ^TrustManagerFactory trust-store)
+          (instance? cert-array-class trust-store)
+          (.trustManager builder ^"[Ljava.security.cert.X509Certificate;" trust-store)
```
KingMob, line 749: Side note: This is so ugly, but then making a protocol and extending it just to exploit some minor polymorphism would be just as ugly. gah

KingMob (approved):
> @KingMob FYI: I'll resolve the conflicts (and squash where appropriate) once you've approved my latest changes!

It looks good to me. No need to squash on my account, I think the existing commits represent a good history.

arnaudgeiser:
I tried to rebase this PR on master but unfortunately I don't have the rights to perform commits on your branch @DerGuteMoritz.
Can I ask you to do so then I will merge?

src/aleph/netty.clj
DerGuteMoritz: Indeed :smile:  I was referring to the whole approach of using `instance?` checks actually which you also commented on below. Resolving this thread then!

src/aleph/netty.clj
DerGuteMoritz: Oh yeah, using protocols for this purpose didn't occur to me :facepalm:  But I agree, this would only trade one ugliness for another, maybe slightly less ugly version :smile: Let's leave it at that then!

DerGuteMoritz:
@arnaudgeiser There you go! Only squashed the indentation commit after all.

DerGuteMoritz:
Hold on a moment, I accidentlly clobbered Oleksii's original commit authorship, restoring ...

DerGuteMoritz:
Alright, good to go now from my end!

PR #622, opened 2022-08-24
Improve short-circuiting behavior of `wrap-future`

### LINKED ISSUE #614: Improve wrap-future
This was uncovered in the process of reviewing #608. See my comments there for more.

In short, `wrap-future` checks `.isSuccess` on a future, when it really means to check `.isDone`. If it succeeded, it creates an immediate success-deferred. If not, it attaches handlers. In the case of a failed/canceled future, it does a lot of unnecessary work before immediately invoking the handler: fn thread frame binding, handler object creation, and scheduling the handlers to run (potentially) on another thread. This is also kind of confusing because we're checking the wrong condition.

To do:
- [ ] Switch to `.isDone` from `.isSuccess` in `wrap-future`
- [ ] If `.isDone`, choose between returning success- and error-deferreds based on the future's status. For consistency, we can reuse the `operation-complete` handler, but add a flag to tell it to use/ignore executors (so we can run it on the current thread)
- [ ] Verify all tests still work, `test-classloader` in particular

DerGuteMoritz:
If the Netty future passed to `wrap-future` is already done, resolve
the deferred result immediately. This short-circuits the Netty future
listener indirection and all the overhead it involves in such a
situation.

Note that this optimization was already partially implemented, namely
for the success case. This patch merely generelizes it to all "done"
cases.

Fixes #614.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -243,10 +243,10 @@
     (let [d (d/deferred nil)]
       (if (.isDone f)
         (operation-complete f d)
-        (let [;; Ensure the same bindings are installed on the Netty thread (vars,
-              ;; classloader) than the thread registering the
-              ;; `operationComplete` callback.
-              bound-operation-complete (bound-fn* operation-complete)]
+        ;; Ensure the same bindings are installed on the Netty thread (vars,
+        ;; classloader) than the thread registering the
+        ;; `operationComplete` callback.
+        (let [bound-operation-complete (bound-fn* operation-complete)]
           (.addListener f
                         (reify GenericFutureListener
                           (operationComplete [_ _]
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -240,19 +240,19 @@
 (defn wrap-future
   [^Future f]
   (when f
-    (if (.isSuccess f)
-      (d/success-deferred (.getNow f) nil)
-      (let [d (d/deferred nil)
-            ;; Ensure the same bindings are installed on the Netty thread (vars,
-            ;; classloader) than the thread registering the
-            ;; `operationComplete` callback.
-            bound-operation-complete (bound-fn* operation-complete)]
-        (.addListener f
-          (reify GenericFutureListener
-            (operationComplete [_ _]
-              (ensure-dynamic-classloader)
-              (bound-operation-complete f d))))
-        d))))
+    (let [d (d/deferred nil)]
+      (if (.isDone f)
+        (operation-complete f d)
+        (let [;; Ensure the same bindings are installed on the Netty thread (vars,
+              ;; classloader) than the thread registering the
+              ;; `operationComplete` callback.
+              bound-operation-complete (bound-fn* operation-complete)]
+          (.addListener f
+                        (reify GenericFutureListener
+                          (operationComplete [_ _]
+                            (ensure-dynamic-classloader)
+                            (bound-operation-complete f d))))))
+      d)))
 
 (defn allocate [x]
   (if (instance? Channel x)


src/aleph/netty.clj
```diff
@@ -240,19 +240,19 @@
...
-            bound-operation-complete (bound-fn* operation-complete)]
-        (.addListener f
-          (reify GenericFutureListener
-            (operationComplete [_ _]
-              (ensure-dynamic-classloader)
-              (bound-operation-complete f d))))
-        d))))
+    (let [d (d/deferred nil)]
+      (if (.isDone f)
+        (operation-complete f d)
```
DerGuteMoritz, line 245: @KingMob in #614 you mentioned:

> For consistency, we can reuse the operation-complete handler, but add a flag to tell it to use/ignore executors (so we can run it on the current thread)

AFAIUI, `(d/deferred nil)` should already take care of this (the `nil` value is the `executor` argument). But please verify :pray:

src/aleph/netty.clj
```diff
@@ -240,19 +240,19 @@
...
+        (let [;; Ensure the same bindings are installed on the Netty thread (vars,
+              ;; classloader) than the thread registering the
+              ;; `operationComplete` callback.
+              bound-operation-complete (bound-fn* operation-complete)]
+          (.addListener f
+                        (reify GenericFutureListener
+                          (operationComplete [_ _]
+                            (ensure-dynamic-classloader)
+                            (bound-operation-complete f d))))))
```
DerGuteMoritz, line 254: Oops, only now noticed that this form got reindented. However, I think this is in line with common style guides anyway :sweat_smile:

src/aleph/netty.clj
KingMob: Yeah, I don't know if I had something else in mind, but setting the deferred's executor to nil should work, and replicate the existing behavior in all cases.

I wondered if we should be doing something different for the non-immediate case. By setting the executor to nil, it runs later on a netty thread, where we ran into those issues with the classloader not supporting clojure. If we didn't pass an executor param at all, it'll just default to the current thread's executor, but that's retrieved from the `manifold.executor/executor-thread-local` ThreadLocal, which will be null if we call `wrap-future` from a netty thread.

Ehhh... never mind, the future's listener runs on a netty thread, so even if we set the deferred's executor to `wait-pool` or set the executor ThreadLocal, we'd _still_ have to do the whole classloader dance anyway.

src/aleph/netty.clj
```diff
@@ -240,19 +240,19 @@
...
-            (operationComplete [_ _]
-              (ensure-dynamic-classloader)
-              (bound-operation-complete f d))))
-        d))))
+    (let [d (d/deferred nil)]
+      (if (.isDone f)
+        (operation-complete f d)
+        (let [;; Ensure the same bindings are installed on the Netty thread (vars,
+              ;; classloader) than the thread registering the
+              ;; `operationComplete` callback.
```
KingMob, line 248: Can you move the double semicolon comments above the `let`?

src/aleph/netty.clj
KingMob: ...it's indented correctly in my code. Did it get pushed out of the correct form level somewhere else?

KingMob (approved):
Approved, but please move those comments

src/aleph/netty.clj
arnaudgeiser: On my side, I always disable all the tooling I have when working on aleph (parinfer/paredit and so on) to avoid
having this behaviour.

I'm the last person who touches this part of the code : https://github.com/clj-commons/aleph/commit/6143d5e5360e2ce00002a6ab66f2ff62eb216490#diff-ef7e90d6943fd305084d91366461caabf23c2ec88466cd4719fd66ea862417b4R244-R246

src/aleph/netty.clj
arnaudgeiser: I confirm, with `nil` we are sure we will be good.

arnaudgeiser (approved):

src/aleph/netty.clj
DerGuteMoritz: @KingMob Ah sorry, the line I attached that comment to wasn't the most obvious. I was referring to the indentation of the second argument to the `(.addListener f ...)` call, i.e. the `(reify ...)`. It wasn't indented to the same column as `f` before.

@arnaudgeiser 
> On my side, I always disable all the tooling I have when working on aleph (parinfer/paredit and so on) to avoid having this behaviour.

Probably a good idea to keep patches minimally invasive :thinking: But preserving the idiosyncratic formatting indefinitely doesn't seem too attractive either. We discussed before to possibly introduce `cljfmt` and do a big honking re-indent commit. So I suggest to just do what seems most reasonable until then. In this case here I'd say the reindent doesn't really hurt.

src/aleph/netty.clj
DerGuteMoritz: Here you go!

PR #624, opened 2022-09-08
Remove epoll request in test

### LINKED ISSUE #620: `aleph.udp-test/test-echo` broken on macs
It mandates epoll, which is now an error if unavailable.

KingMob:
Fixes #620

### DIFF
--- a/test/aleph/udp_test.clj
+++ b/test/aleph/udp_test.clj
@@ -16,7 +16,7 @@
          (.close ^java.io.Closeable server#)))))
 
 (deftest test-echo
-  (let [s @(udp/socket {:port 10001, :epoll? true})]
+  (let [s @(udp/socket {:port 10001})]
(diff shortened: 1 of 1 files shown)


arnaudgeiser (approved):

DerGuteMoritz:
FTR, [I just had the same error](https://app.circleci.com/pipelines/github/clj-commons/aleph/217/workflows/94f4080c-fe66-4712-beca-91ef7670c81d/jobs/212) on another PR. Maybe we need to bump the timeout value here a bit?

KingMob:
I'll rerun the test, but I don't think we should bump the timeout before checking for a regression. The `slow-stream` in the test writes every 50 ms, but the idle timeout is 200 ms, so it shouldn't timeout. Plus, both tests failed at the `"Client is slow to write"` subsection, which suggests it's repeatable and specific.

I'll rerun, but let's open a new issue to check this.

PR #625, opened 2022-09-08
Add minimal tools.deps support

DerGuteMoritz:
Just enough to allow using aleph as a git dependency via tools.deps. Both the deps.edn file and build.clj's inputs are generated from project.clj which remains the source of truth for now. Can be used as a starting point for a potential full switch to tools.deps in the future.

Also included is a script for re-generating these files, as well as a CI check which should ensure that the generated files don't diverge. Note that these are intentionally not written as tools.deps or babashka tasks so as not to introduce additional dev and build dependencies.

----

FWIW, I've already tested this in practice, too.

Let me know if you have any ideas for improvement or disagree with e.g. the naming or the `deps` directory structure :pray:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,5 +37,7 @@ jobs:
             - ~/.m2
           key: v1-dependencies-{{ checksum "project.clj" }}
 
+      - run: deps/ensure-deps-up-to-date
+
       # run tests!
       - run: lein do clean, test
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
-* @arnaudgeiser @kingmob
+* @arnaudgeiser @kingmob @dergutemoritz
+/deps/ @dergutemoritz
--- a/README.md
+++ b/README.md
@@ -123,6 +123,12 @@ Where incoming packets will have a `:message` that is a byte-array, which can be
 
 To learn more, [read the documentation](http://aleph.io/examples/literate.html).
 
+### Development
+
+Aleph uses [Leinigen](https://leiningen.org/) for managing dependencies, running REPLs and tests, and building the code.
+
+Minimal [`tools.deps`](https://github.com/clojure/tools.deps.alpha) support is available in the form of a `deps.edn` file which is generated from `project.clj`. It provides just enough to be able to use Aleph as a git or `:local/root` dependency. When committing changes to `project.clj`, run `deps/lein-to-deps` and commit the resulting changes, too.
+
 ### License
 
 Copyright © 2010-2020 Zachary Tellman
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ To learn more, [read the documentation](http://aleph.io/examples/literate.html).
 
 ### Development
 
-Aleph uses [Leinigen](https://leiningen.org/) for managing dependencies, running REPLs and tests, and building the code.
+Aleph uses [Leiningen](https://leiningen.org/) for managing dependencies, running REPLs and tests, and building the code.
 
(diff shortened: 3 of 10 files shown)


KingMob:
A couple people have asked for similar stuff over the years. I can't find my reply from the last time, but I'll sum up my stance like this: any switch to tools.deps must replicate _all_ the functionality we currently have with lein before switching. Tools.deps has a few niceties like local/url/sha deps, but nothing that's a game-changer for aleph's needs. 

There's also things that tools.deps lacks, and may never properly support, that we use. (E.g., setting global vars like `*warn-on-reflection*`. There's a [hack](https://stackoverflow.com/questions/69790638/is-it-possible-to-set-warn-on-reflection-in-deps-edn) involving `:main-opts`, but it breaks with anything else setting `:main-opts`).

That being said, I'm totally fine with an auto-generated deps.edn for users who don't want to use maven-based jars.

However, I'm not sure about the inclusion of the two `deps/aleph/build.*` files. It's not like we have a lot of devs to get confused about them yet, but they're not the official build tooling at the moment. I'd prefer to keep them out for now, unless you can think of a good reason to include them.

DerGuteMoritz:
> A couple people have asked for similar stuff over the years. I can't find my reply from the last time, but I'll sum up my stance like this: any switch to tools.deps must replicate _all_ the functionality we currently have with lein before switching. Tools.deps has a few niceties like local/url/sha deps, but nothing that's a game-changer for aleph's needs.

Right, that's why I took the minimal approach here: Switching fully would be tremendous effort for little gain at the moment.

> That being said, I'm totally fine with an auto-generated deps.edn for users who don't want to use maven-based jars.
> 
> However, I'm not sure about the inclusion of the two `deps/aleph/build.*` files. It's not like we have a lot of devs to get confused about them yet, but they're not the official build tooling at the moment. I'd prefer to keep them out for now, unless you can think of a good reason to include them.

This is actually a minimum requirement for making it usable as a git dependency (which was my main goal here): Since Aleph ships a handful of Java source files and Clojure doesn't transparently compile these like it does `.clj` sources,  we need to provide a [`:deps/prep-lib`](https://clojure.org/guides/deps_and_cli#prep_libs) task. IMHO, being able to use Aleph as a git dependency is desirable for allowing people to try out unreleased versions without going through the trouble of putting a custom built jar into a maven repo somewhere. Case in point: I was more annoyed by the prospect of that than by having to code up this patch here :smile: 

How about I add a comment to `deps/aleph/build.clj` which explains this? `deps/aleph/build.edn` already has a comment at the top which at least should deter any future developers from touching it. Ah and I'll also add a paragraph about it to `README.md`. WDYT?

KingMob (approved):
Changes look good to me. I didn't even think about the issues around compiling Java. 

You might want to add yourself to CODEOWNERS for that directory, so people know who to complain to 😄

DerGuteMoritz:
Good call, done! Also added myself to the main list of owners while at it :+1:

KingMob:
CI might need rerunning. Can you merge, btw?

DerGuteMoritz:
Yep, I can, but I can't trigger reruns of failed CI jobs. This reminds me I should give the CI check I added a shot in practice. Lemme just do that on here - will merge it myself afterwards if it works as intended!

KingMob:
I restarted it for you. 

@slipset Can you add @DerGuteMoritz to CircleCI for job-restarting perms?

DerGuteMoritz:
Alright, the check failed successfully :smile: Merging away!

DerGuteMoritz:
Failed test is https://github.com/clj-commons/aleph/issues/626 once more. Proceeding with the merge regardless.

PR #627, opened 2022-09-10
Stabilize `aleph.http-test/test-idle-timeout`

### LINKED ISSUE #626: Look into intermittently failing slow client test
Discussion in #624.  

E.g.: https://app.circleci.com/pipelines/github/clj-commons/aleph/217/workflows/94f4080c-fe66-4712-beca-91ef7670c81d/jobs/212

It's happened at least twice in the same spot, which is suspicious. 

- [ ] Is there a regression? 
- [ ] Did CircleCI downgrade our test images? (unlikely)
- [ ] The idle timeout pipeline changed recently. 
- [ ] Could also be in manifold.
- [ ] Can we replicate locally?

Suggest bisection.

DerGuteMoritz:
Fixes #626.

More like a workaround but given how `byte-streams` works, I can't see how to better fix this. Maybe look into optimizing the conversion path construction?

### DIFF
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -541,7 +541,14 @@
         (d/chain' (fn [_] (s/close! body))))
     body))
 
+(defn force-stream-to-string-memoization! []
+  (bs/to-string (doto (s/stream 1) (s/put! "x") s/close!)))
+
 (deftest test-idle-timeout
+  ;; Required so that the conversion initialization doesn't count
+  ;; toward the idle timeout. See
+  ;; https://github.com/clj-commons/aleph/issues/626 for background.
+  (force-stream-to-string-memoization!)
   (let [url (str "http://localhost:" port)
         echo-handler (fn [{:keys [body]}] {:body (bs/to-string body)})
         slow-handler (fn [_] {:body (slow-stream)})]


test/aleph/http_test.clj
```diff
@@ -541,7 +541,14 @@
         (d/chain' (fn [_] (s/close! body))))
     body))
 
+(defn init-stream-to-string-conversion! []
+  (bs/to-string (doto (s/stream 1) (s/put! "x") s/close!)))
```
KingMob, line 545: Maybe call it something like `force-memoization`?

Side note: I'm not really sure why Zach tried to delay it. Sure, you don't pay the cost of conversions you don't use, but most of them are just fns. I'd be curious to know the actual overhead.

KingMob (approved):
I think the name could be improved, but I don't care all that much.

test/aleph/http_test.clj
DerGuteMoritz: > Maybe call it something like `force-memoization`?

Sure, why not! Went with `force-stream-to-string-memoization!`

 
> Side note: I'm not really sure why Zach tried to delay it. Sure, you don't pay the cost of conversions you don't use, but most of them are just fns. I'd be curious to know the actual overhead.

My guess is that it's actually tricky to do this eagerly but still keep the conversion graph extensible. If you eagerly instantiate the converters (e.g. at the bottom of the `clj-commons.byte-streams` namespace), any third-party `def-conversion` declarations which would influence the optimal conversion paths would be ineffective. Alternatively, if you eagerly instantate the converters whenever a new conversion is declared, you'd have a lot of redundant instantiations during load time (factorial?) while all the stock conversions are being registered. I guess one way out of this pickle would be to provide an API function for eagerly instantiate the converters. This way, the user could call it at a point in their program when they know that all dependencies have been loaded. I'll file an issue over at https://github.com/clj-commons/byte-streams about it and will conduct some experiments to verify my hypotheses here.

DerGuteMoritz: OK I just created https://github.com/clj-commons/byte-streams/issues/61 which picks up on this theme. Let's continue this thread over there :+1:

PR #630, opened 2022-10-06
Upgrade to netty 4.1.82

valerauko:
Netty 4.1.80 [apparently had some pretty significant performance improvements](https://netty.io/news/2022/08/26/4-1-80-Final.html), so it should be a priority target for getting a release out with.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -3,22 +3,22 @@
  :deps
  {org.clojure/tools.logging
   {:mvn/version "1.1.0", :exclusions [org.clojure/clojure]},
-  io.netty/netty-handler-proxy {:mvn/version "4.1.79.Final"},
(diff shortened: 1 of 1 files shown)


DerGuteMoritz:
Thanks! As mentioned on Slack, this is a very welcome update. CircleCI seems to have some trouble checking the commit out right now though:

```
fatal: Couldn't find remote ref refs/pull/630/head
```

I already hit retry but same thing. Any idea what might be going on?

KingMob (approved):

KingMob:
For future reference, we should consider https://github.com/borkdude/tools#lein2deps. Using Babashka, it'll probably be fast enough to use as a precommit hook.

PR #634, opened 2022-10-11
dependency: bouncycastle must be a dev dependency

arnaudgeiser:
## Description

Aleph cannot be compiled on Java > 9 when `bouncycastle` is not included as a dependency. [1]
As `lein test` loads both the `dev` and `test` profile, everything is fine.
However, a REPL without `with-profile +test` results on `bouncycastle` not being
included which fallback to  `OpenJdkSelfSignedCertGenerator` and thus :

```
Execution error (IllegalAccessError) at io.netty.handler.ssl.util.OpenJdkSelfSignedCertGenerator/generate (OpenJdkSelfSignedCertGenerator.java:52).
class io.netty.handler.ssl.util.OpenJdkSelfSignedCertGenerator (in unnamed module @0x3bb50eaa) cannot access class sun.security.x509.X509CertInfo (in module java.base) because module java.base does not export sun.security.x509 to unnamed module @0x3bb50eaa
```

Let's add `bouncycastle` as a `dev` dependency instead.
To help diagnose this issue, I also changed `-Dorg.slf4j.simpleLogger.defaultLogLevel=debug` on `:dev` profile.

[1] : https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/util/SelfSignedCertificate.java#L238-L255

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -30,8 +30,8 @@
                                    [spootnik/signal "0.2.4"]
                                    [me.mourjo/dynamic-redef "0.1.0"]
                                    ;; This is for self-generating certs for testing ONLY:
-                                   [org.bouncycastle/bcprov-jdk15on "1.69"]
-                                   [org.bouncycastle/bcpkix-jdk15on "1.69"]]
+                                   [org.bouncycastle/bcprov-jdk18on "1.72"]
+                                   [org.bouncycastle/bcpkix-jdk18on "1.72"]]
                     :jvm-opts ["-Dorg.slf4j.simpleLogger.defaultLogLevel=debug"]}
              :lein-to-deps {:source-paths ["deps"]}
              :test {:jvm-opts ["-Dorg.slf4j.simpleLogger.defaultLogLevel=off"]}}
--- a/project.clj
+++ b/project.clj
@@ -28,12 +28,13 @@
                                    [org.slf4j/slf4j-simple "1.7.30"]
                                    [com.cognitect/transit-clj "1.0.324"]
                                    [spootnik/signal "0.2.4"]
-                                   [me.mourjo/dynamic-redef "0.1.0"]]}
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -28,12 +28,13 @@
                                    [org.slf4j/slf4j-simple "1.7.30"]
                                    [com.cognitect/transit-clj "1.0.324"]
                                    [spootnik/signal "0.2.4"]
-                                   [me.mourjo/dynamic-redef "0.1.0"]]}
-             :lein-to-deps {:source-paths ["deps"]}
-             ;; This is for self-generating certs for testing ONLY:
-             :test {:dependencies [[org.bouncycastle/bcprov-jdk15on "1.69"]
+                                   [me.mourjo/dynamic-redef "0.1.0"]
+                                   ;; This is for self-generating certs for testing ONLY:
+                                   [org.bouncycastle/bcprov-jdk15on "1.69"]
```
DerGuteMoritz, line 33:
```suggestion
                                   [org.bouncycastle/bcprov-jdk18on "1.72"]
```

How about this while we're at it?

DerGuteMoritz (changes requested):
Ah yes, I ran into this with our use of BC, as well. FTR, this also [the solution Netty devs recommend for this issue](https://github.com/netty/netty/issues/10317).

project.clj
KingMob: We're not supporting anything below 1.8, so the upgrade should be good on that front.

KingMob (approved):
LGTM

project.clj
arnaudgeiser: https://github.com/clj-commons/aleph/pull/634/commits/82a39b815d9aac4bcfe6375a83fdded7895c0b57

PR #635, opened 2022-10-14
Switch to lein2deps

DerGuteMoritz:


arnaudgeiser (approved):
Thanks borkdude :)

KingMob (changes requested):
Can we automate this now? Either via git pre-commit hook or in CI?

arnaudgeiser:
My 2 cents: I would prefer the automation on the CI if possible.

arnaudgeiser:
So basically, replacing `deps/ensure-deps-up-to-date` [1] implementation by something that
would regenerate the `deps.edn` (instead of throwing an error) file.

Either by:
- Adding an additional commit with only the `deps.edn` file change
- Updating `deps.edn` as part of the previous commit

I don't know how difficult is the latter, but I would prefer that way
to keep the history clean.

In any case, I don't have any strong opinion on this subject...

[1] : https://github.com/clj-commons/aleph/pull/625/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R40

KingMob:
What's the latest word with this?

DerGuteMoritz:
Oh yeah, I'm actually not strongly opposed to relaxing the "atomic change" rule. However, I have no experience in setting up a CircleCI job to allow it to push commits so this would require extra effort I can't expend right now. We could just file an issue about this, tho!

arnaudgeiser:
On my side, it was just a proposition.
I don't really think it's worth spending time on it.
Our time is better invested doing something else.

KingMob (approved):

KingMob:
I don't want to hold it up, and I agree it's not high priority. 

Still, it really feels like this is something computers should be able to do 😄 

Approving for now

DerGuteMoritz:
> Still, it really feels like this is something computers should be able to do smile

I feel you :smile: But yeah, probably not worth pushing on this case.

That said, I just thought of a potential way to achieve `tools.deps` compat without pre-generating a `deps.edn` file: Add a (static) dependency on a `:local/root` which points to a non-existing subdir and is populated by a `:deps/prep-lib` task which bascially just runs `lein2deps`. Drawback is that we'd push a `bb` dependency onto the user but I think it should also be possible to run `lein2deps` via `clojure` (possibly requires some tweaks). Seems a bit hacky, tho. Will give it a try anyway when I find a minute to spare :smile: 

Merging away!

PR #636, opened 2022-10-14
Use str instead of .toString

saguywalker:


arnaudgeiser (approved):

KingMob (commented):
@saguywalker First, thanks for contributing. I'm OK with this PR because it's so small, but you should know that tiny little clean-up PRs aren't really a good use of reviewers' time, relative to the benefits.

It's always better to tackle an actual problem or feature, and clean up along the way.

Issue #638, opened 2022-11-02
Accept shutdownGracefully parameters


arnaudgeiser:
Currently, shutdownGracefully cannot be configured and we rely on Netty's default behavior which will accept new connections during 2 seconds and stop the ~`EventExecutor`~ (EDIT: wrong `EventLoopGroup`) after 15 seconds. [1]

Some workloads need this behavior to be tweakable.

- [ ]  Add those two constants inside `netty.clj` so people can rely on them [2]
- [ ]  Make `start-server` accept those new options (please refactor, add another arity function with the parameter as a map) [2]

[1] : https://github.com/netty/netty/blob/b61d7d40f40e3a797e8a60cd567f849a9799c771/common/src/main/java/io/netty/util/concurrent/AbstractEventExecutor.java#L40-L41
[2] : https://github.com/clj-commons/aleph/blob/master/src/aleph/netty.clj#L1282

DerGuteMoritz:
Oh this is great! The default timeouts are particularly egregious when running tests. This actually made me remove the `netty/wait-for-close` call at some point. Glad to hear that there is a way to parameterize this.

arnaudgeiser:
While those parameters are now configurable, they will work as expected when you are trying to decrease them, not increase them.

Here is an example:

```clojure
(defn handler [{:keys [body]}]
  (let [res (s/reduce (fn [acc _]
                        (prn acc)
                        (Thread/sleep 1000)
                        (inc acc))
                      0 body)]
    {:body (str @res)}))

(def server (http/start-server
              #'handler
              {:port 9999
               :raw-stream? true
               :shutdown-timeout 200
               :shutdown-quiet-period 100}))
```

Then upload a huge file

```
curl localhost:9999 --data-binary @/tmp/100GiB
```

Wait five seconds

```
1
2
3
4
5
```

Stop the server

```
(.close server)
```

It's stopped after 5 seconds again instead of 200

```
6
7
8
9
10
```

Connection is closed client side

```
curl: (55) Send failure: Broken pipe
```

But this is the expected behaviour of Netty according to this comment : https://github.com/netty/netty/pull/3706#issuecomment-97715011

cc @pyr 

Closing the issue as it has been addressed by #639

DerGuteMoritz:
> While those parameters are now configurable, they will work as expected when you are trying to decrease them, not increase them.

Hm when reading your example scenario, it appears that it's rather the other way around, i.e. decreasing isn't possible? :thinking:

[...]

PR #641, opened 2022-11-05
Support graceful shutdown with ChannelGroup

arnaudgeiser:
## Description

This PR is an attempt to address #638 where we noticed that `EventLoopGroup#shutdownGracefully` timeouts are not waiting for active connections but is closing all the `EventExecutor` right away. [1]
It's not an issue on the Netty side but an expected behaviour. [2]

Still, people using a `shutdown-timeout` would expect active connections to be finished before shutting down the server.
The proposition is to track all `Channel` on an `ChannelGroup` [3], and wait for them to be closed before shutting down the server. This use case is mentioned on the `ChannelGroup` Java documentation.

- [x] : Track all Channel on `channel-active` for HTTP and TCP servers
- [x] : Wait the `ChannelGroup` for at most the `shutdown-timeout`
- [x] : Take into account the `elasped` time when calling `EventLoopGroup#shutdownGracefully`
- [x] : Ensure it's backward compatible with public exposed functions
- [x] : Deprecate `shutdown-quiet-period` (which hasn't been released)   
- [x] : Reformat the code

## Open questions

- ~Shall we apply this behavior only when `shutdown-timeout` is explicitly set? And keep the current weird behavior?~
- ~Do we have to be backward compatible with `raw-ring-handler` and `ring-handler`? Having that `when channel-group` is bothering me...~

Let's keep the current behavior when `shutdown-timeout` is not explicitly set.

[1] : https://github.com/netty/netty/blob/4.1/transport-classes-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java#L420
[2] : https://github.com/netty/netty/pull/3706#issuecomment-1303066857
[3] : https://netty.io/4.0/api/io/netty/channel/group/ChannelGroup.html

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -52,8 +52,7 @@
    | `idle-timeout` | when set, connections are closed after not having performed any I/O operations for the given duration, in milliseconds. Defaults to `0` (infinite idle time).
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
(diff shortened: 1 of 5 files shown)


arnaudgeiser:
Closing this PR in favor of another that will come soon.

PR #642, opened 2022-11-09
netty: configurable server graceful shutdowns

pyr:
The recently closed #641 and previous #639 were both attempting to improve the server shutdown procedure.

The intent of both these PRs is to make it easily viable to provide the classic blue-green deployment mode where:

- Servers stop listening for incoming requests
- In-flight requests are still processed for a configurable grace period
- All resources are then cleaned up

After more digging (my Netty-foo being a bit rusty of late :-)) an approach relying on
Netty's [*ChannelGroup*](https://netty.io/4.1/api/io/netty/channel/group/ChannelGroup.html) functionality seems to be the right mechanism.

Of course, not everyone will want to keep a set of all active connections, including its memory and latency impact.

This revised PR modifies the previous approach in the following way:

- Allow for hooking into the shutdown process by way of the `shutdown-hook` parameter (accepted by `http/start-server`,` `tcp/start-server`, and `netty/start-server`).
- Provide a maximum time interval for `shutdown-hook` to run in, defaulting to 15 seconds
- ~Walk back configurability of `shutdownGracefully` parameters as they do not give a reliable way to control shutdown.~

In addition to this, a few utilities are provided to handle channel groups, as well as a ready made channel handler to populate a channel group, since this will be the most common use of `shutdown-hook`.

If this PR goes through, I am happy to open a subsequent one with updated docs with a walk-through on how to achieve these stop semantics.

(useful disclosure, @arnaudgeiser and I work together so take his approval seal with a grain of salt :smile:)

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -52,8 +52,7 @@
    | `idle-timeout` | when set, connections are closed after not having performed any I/O operations for the given duration, in milliseconds. Defaults to `0` (infinite idle time).
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
-   | `shutdown-quiet-period` | optional period in seconds for which new connections will still be serviced after a scheduled shutdown via `java.io.Closeable#close`. Defaults to 2 seconds.
-   | `shutdown-timeout` | optional grace period in seconds on which to wait for the event loop group to empty during a scheduled shutdown. Defaults to 15 seconds."
+   | `shutdown-timeout` | interval in seconds within which in-flight requests must be processed, defaults to 15 seconds. A value of 0 bypasses waiting entirely."
   [handler options]
   (server/start-server handler options))
 
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -22,6 +22,8 @@
      ChannelOutboundHandler
      ChannelHandlerContext
      ChannelInitializer]
+    [io.netty.channel.group
+     ChannelGroup DefaultChannelGroup]
     [io.netty.channel.epoll Epoll EpollEventLoopGroup
      EpollServerSocketChannel
      EpollSocketChannel
@@ -57,7 +59,8 @@
      ResourceLeakDetector$Level]
     [java.net URI SocketAddress InetSocketAddress]
     [io.netty.util.concurrent
-     AbstractEventExecutor FastThreadLocalThread GenericFutureListener Future]
+     AbstractEventExecutor FastThreadLocalThread GenericFutureListener Future
+     GlobalEventExecutor]
     [java.io InputStream File]
     [java.nio ByteBuffer]
     [java.nio.channels ClosedChannelException]
@@ -121,15 +124,11 @@
 (defn channel-remote-address [^Channel ch]
   (some-> ch ^InetSocketAddress (.remoteAddress) .getAddress .getHostAddress))
 
-;;;
-
-(def ^:const default-shutdown-quiet-period
-  "Default quiet period in seconds when shutting down an eventloop group"
-  2) ;; As in https://github.com/netty/netty/blob/b61d7d40f40e3a797e8a60cd567f849a9799c771/common/src/main/java/io/netty/util/concurrent/AbstractEventExecutor.java#L40
+;;; Defaults defined here since they are not publically exposed by Netty
 
 (def ^:const default-shutdown-timeout
-  "Default timeout in seconds when shutting down an eventloop group"
-  15) ;; As in https://github.com/netty/netty/blob/b61d7d40f40e3a797e8a60cd567f849a9799c771/common/src/main/java/io/netty/util/concurrent/AbstractEventExecutor.java#L41
+  "Default timeout in seconds to wait for graceful shutdown complete"
+  15)
 
 (def ^:const array-class (class (clojure.core/byte-array 0)))
 
@@ -294,6 +293,12 @@
     x
     (.channel ^ChannelHandlerContext x)))
 
+(defn ^ChannelGroup make-channel-group
+  "Create a channel group which can be used to perform channel operations on
+   several channels at once."
+  []
+  (DefaultChannelGroup. GlobalEventExecutor/INSTANCE))
+
 (defmacro safe-execute
   "Executes the body on the event-loop (an executor service) associated with the Netty channel.
 
@@ -433,8 +438,6 @@
           d))))
   (put [this msg blocking? timeout timeout-value]
     (.put this msg blocking?)))
-
-
 
 (defn sink
   ([ch]
@@ -693,6 +696,17 @@
             (.readableBytes ^ByteBuf msg)))
         (.write ctx msg promise)))))
 
+(defn ^ChannelHandler channel-tracking-handler
+  "Yields an inbound handler, ready to be added to a pipeline,
+   which keeps track of requests in a provided channel group.
+   The channel-group can be created via `make-channel-group`."
+  [^ChannelGroup group]
+  (channel-inbound-handler
+   :channel-active
+   ([_ ctx]
+    (.add group (channel ctx))
+    (.fireChannelActive ctx))))
+
 (defn pipeline-initializer [pipeline-builder]
   (proxy [ChannelInitializer] []
     (initChannel [^Channel ch]
@@ -702,6 +716,16 @@
   (when (some? (.get pipeline handler))
     (.remove pipeline handler))
   pipeline)
+
+(defn append-handler-to-pipeline
+  "Convenience function to add a handler to the tail of a netty pipeline"
+  [^ChannelPipeline pipeline handler-id ^ChannelHandler handler]
+  (.addLast pipeline (str handler-id) handler))
+
+(defn prepend-handler-to-pipeline
+  "Convenience function to add a handler to the head of a netty pipeline"
+  [^ChannelPipeline pipeline handler-id ^ChannelHandler handler]
+  (.addFirst pipeline (str handler-id) handler))
 
 (defn instrument!
   [stream]
@@ -1226,6 +1250,21 @@ initialize an DnsAddressResolverGroup instance.
               (let [ch (.channel ^ChannelFuture f)]
                 (maybe-ssl-handshake-future ch)))))))))
 
+(defn- add-ssl-handler
+  [pipeline-builder ssl-ctx]
+  (fn [^ChannelPipeline pipeline]
+    (append-handler-to-pipeline
+     pipeline "ssl-handler"
+     (.newHandler ^SslContext ssl-ctx (-> pipeline .channel .alloc)))
+    (pipeline-builder pipeline)))
+
+(defn- add-channel-tracker-handler
+  [pipeline-builder chan-group]
+  (fn [pipeline]
+    (prepend-handler-to-pipeline pipeline "channel-tracker"
+                                (channel-tracking-handler chan-group))
+    (pipeline-builder pipeline)))
+
 (defn start-server
   ([pipeline-builder
     ssl-context
@@ -1233,26 +1272,27 @@ initialize an DnsAddressResolverGroup instance.
     on-close
     socket-address
     epoll?]
-   (start-server {:pipeline-builder pipeline-builder
-                  :ssl-context ssl-context
+   (start-server {:pipeline-builder    pipeline-builder
+                  :ssl-context         ssl-context
                   :bootstrap-transform bootstrap-transform
-                  :on-close on-close
-                  :socket-address socket-address
-                  :transport (if epoll? :epoll :nio)}))
+                  :on-close            on-close
+                  :socket-address      socket-address
+                  :transport           (if epoll? :epoll :nio)}))
   ([{:keys [pipeline-builder
             ssl-context
             bootstrap-transform
             on-close
             ^SocketAddress socket-address
             transport
-            shutdown-quiet-period
-            shutdown-timeout]}]
+            shutdown-timeout]
+     :or {shutdown-timeout default-shutdown-timeout}}]
    (when (= transport :epoll)
      (ensure-epoll-available!))
    (let [num-cores      (.availableProcessors (Runtime/getRuntime))
          num-threads    (* 2 num-cores)
          thread-factory (enumerating-thread-factory "aleph-netty-server-event-pool" false)
          closed?        (atom false)
+         chan-group     (make-channel-group)
 
          ^EventLoopGroup group
          (condp = transport
@@ -1272,13 +1312,8 @@ initialize an DnsAddressResolverGroup instance.
                        (coerce-ssl-server-context ssl-context))
 
          pipeline-builder
-         (if ssl-context
-           (fn [^ChannelPipeline p]
-             (.addLast p "ssl-handler"
-                       (.newHandler ssl-context
-                                    (-> p .channel .alloc)))
-             (pipeline-builder p))
-           pipeline-builder)]
+         (cond-> (add-channel-tracker-handler pipeline-builder chan-group)
+           (some? ssl-context) (add-ssl-handler ssl-context))]
 
      (try
        (let [b (doto (ServerBootstrap.)
@@ -1298,10 +1333,25 @@ initialize an DnsAddressResolverGroup instance.
            java.io.Closeable
            (close [_]
              (when (compare-and-set! closed? false true)
+               ;; This is the three step closing sequence:
+               ;; 1. Stop listening to incoming requests
                (-> ch .close .sync)
-               (-> group (.shutdownGracefully (long shutdown-quiet-period)
-                                              (long shutdown-timeout)
-                                              TimeUnit/SECONDS))
+               (-> (if (pos? shutdown-timeout)
+                     ;; 2. Wait for in-flight requests to stop processing within the supplied timeout
+                     ;;    interval.
+                     (-> (wrap-future (.newCloseFuture chan-group))
+                         (d/timeout! (* shutdown-timeout 1000) ::timeout))
+                     (d/success-deferred ::noop))
+                   (d/chain'
+                    (fn [shutdown-output]
+                      (when (= shutdown-output ::timeout)
+                        (log/error
+                         (format "Timeout while waiting for requests to close (exceeded: %ss)"
+                                 shutdown-timeout)))))
+                   (d/finally'
+                    ;; 3. At this stage, stop the EventLoopGroup, this will cancel any
+                    ;;    in flight pending requests.
+                    #(.shutdownGracefully group 0 0 TimeUnit/SECONDS)))
                (when on-close
                  (d/chain'
                   (wrap-future (.terminationFuture group))
--- a/src/aleph/tcp.clj
+++ b/src/aleph/tcp.clj
@@ -80,15 +80,13 @@
    | `bootstrap-transform` | a function that takes an `io.netty.bootstrap.ServerBootstrap` object, which represents the server, and modifies it.
    | `pipeline-transform` | a function that takes an `io.netty.channel.ChannelPipeline` object, which represents a connection, and modifies it.
    | `raw-stream?` | if true, messages from the stream will be `io.netty.buffer.ByteBuf` objects rather than byte-arrays.  This will minimize copying, but means that care must be taken with Netty's buffer reference counting.  Only recommended for advanced users.
-   | `shutdown-quiet-period` | optional period in seconds for which new connections will still be serviced after a scheduled shutdown via `java.io.Closeable::close`. Defaults to 2 seconds.
-   | `shutdown-timeout` | optional grace period in seconds on which to wait for the event loop group to empty during a scheduled shutdown. Defaults to 15 seconds.  "
+   | `shutdown-timeout` | interval in seconds within which in-flight requests must be processed, defaults to 15 seconds. A value of 0 bypasses waiting entirely."
   [handler
    {:keys [port socket-address ssl-context bootstrap-transform pipeline-transform epoll?
-           shutdown-quiet-period shutdown-timeout]
+           shutdown-timeout]
     :or {bootstrap-transform identity
          pipeline-transform identity
          epoll? false
-         shutdown-quiet-period netty/default-shutdown-quiet-period
          shutdown-timeout netty/default-shutdown-timeout}
     :as options}]
   (netty/start-server
@@ -103,7 +101,6 @@
                       socket-address
                       (InetSocketAddress. port))
     :transport (if epoll? :epoll :nio)
-    :shutdown-quiet-period shutdown-quiet-period
     :shutdown-timeout shutdown-timeout}))
 
 (defn- ^ChannelHandler client-channel-handler
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -180,25 +180,25 @@
            (.shutdown *pool*))))))
 
 (defmacro with-handler [handler & body]
-  `(with-server (http/start-server ~handler {:port port})
+  `(with-server (http/start-server ~handler {:port port :shutdown-timeout 0})
      ~@body))
 
 (defmacro with-compressed-handler [handler & body]
   `(do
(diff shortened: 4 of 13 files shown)


arnaudgeiser (approved):
The approach looks pretty solid.
I think we should write some documentation to explain how to gracefully shutdown a Netty server. It doesn't seem to be that well documented on the wild.

I leave the final word to @KingMob on this PR.
On our side, we are currently monkey-patching Aleph to introduce what @pyr is proposing here.
If this PR got merged, you can definitely land a 0.5.1 version... and trust me, we'll be the first user of it.

arnaudgeiser:
Funny... I just added a commit on top of this PR and the tests ran....

KingMob (changes requested):
So, the approach generally looks OK, with a few caveats.

First, I dislike all the orchestration people will have to go through to get everything correct. Correct me if I'm wrong, but if it's like the tests, people will need to:

1. Use `channel-tracking-handler` and manually transform the pipeline to ensure any shutdown that waits on the channels actually waits on them. Preferably, this should be invisible to users.
2. Create a channel-group, and obtain its closing Future
3. Make sure their shutdown hook waits on that future.

There's a lot of user education and steps to get right.

Second, if users don't do all that, shutdowns won't be "graceful", in the sense of waiting for the channels to wrap up within the time limit. Do we want that to be the default, when it doesn't match the intuitive understanding of shutdown timeouts?

Is there a good reason we shouldn't handle all this automatically when there's no shutdown hook? And when there _is_ a hook, can we not do it invisibly on the user's behalf so they don't have to manually coordinate all this?

src/aleph/netty.clj
```diff
@@ -294,6 +297,26 @@
...
+
+(defn channel-group-close-future
+  "Yields a new future realized once all active channels in a channel group are closed"
+  [^ChannelGroup group]
+  (.newCloseFuture group))
+
+(defn append-handler-to-pipeline
+  "Convenience function to add a handler to the tail of a netty pipeline"
+  [^ChannelPipeline pipeline handler-id ^ChannelHandler handler]
+  (.addLast pipeline (str handler-id) handler))
```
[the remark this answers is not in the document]
pyr: Most of the interaction with aleph is type-hint free for downstreams consumers (at the odd exception of `.close` which is part of the equation here :-))

This helps guarantee it stays that way.

pyr:
@KingMob I agree with you, I still think hooking into the process could be important so I left the possibility open but now defaulted to creating a sensible hook.

The parameters need rewording and the docstrings need to be updated, but before going further I want to know if we agree on the flow:

- by default, behave as before
- if a `shutdown-hook-timeout` is provided but no hook, track requests and let them be processed before shutting down within the supplied interval
- if both shutdown-hook-timeout and shutdown-hook are provided, run the provided hook

I suppose the parameters could be named: `graceful-shutdown-period` and `graceful-shutdown-hook`, would this work?

KingMob:
For future reference, please don't force-push in the middle of reviewing a large PR. I personally love to rewrite history, but it has to be done carefully.

It makes it impossible to see only new changes since the last review. I only force push when a) nobody's started reviewing, b) everybody's done reviewing, c) _all_ the code needs to be re-reviewed, or d) nobody's going to review the changes.

KingMob:
Let's take a step back and think about the usage and the API. 

Will users _ever_ want/need NON-graceful shutdown? (I.e., giving channels time to wrap up)  

If not, we should wrap their hook in one that automatically takes care of the graceful details, and eliminate fns they can mess with, either by inlining them, declaring them private, or at a bare minimum, apply `^:no-doc` metadata. In this scenario, everyone always wants a graceful shutdown.

If they _do_ need immediate shutdown, can we still use the graceful shutdown method (channel group, pipeline xform, etc), but tell users to set the timeout to 0 if they want it closed ASAP? It's intuitive and doesn't require two code paths.

If we _are_ maintaining a separate code path, should there be a default, which should it be, and what should that look like?
       - `:graceful-shutdown?` (maybe defaulting to true)
       - `:immediate-shutdown?` (maybe defaulting to false)
       - `:wait-for-channels-during-shutdown?`
       - etc.

I would prefer to avoid separate code paths if avoidable. Conditionally making channel groups, altering the pipeline, and maintaining different hooks is really unappealing to me if we don't have to. My preference is to make it all graceful, and let people set the timeout to 0 if they're in a hurry.

src/aleph/netty.clj
```diff
@@ -1271,6 +1330,13 @@ initialize an DnsAddressResolverGroup instance.
...
+         pipeline-builder
+         (if (some? chan-group)
+           (fn [^ChannelPipeline p]
+             (append-handler-to-pipeline p "group-tracker" (channel-tracking-handler chan-group))
+             (pipeline-builder p))
+           pipeline-builder)
+
          pipeline-builder
          (if ssl-context
            (fn [^ChannelPipeline p]
```
KingMob, line 1342: For future reference, `cond->` would be the better way to do this, since we can combine them.

pyr:
> Will users ever want/need NON-graceful shutdown? (I.e., giving channels time to wrap up)

Yes, high traffic proxies will not want to pay the memory and latency cost of maintaining the channel set.

pyr:
> If we _are_ maintaining a separate code path, should there be a default, which should it be, and what should that look like? 
> - `:graceful-shutdown?` (maybe defaulting to true)
> - `:immediate-shutdown?` (maybe defaulting to false)
> - `:wait-for-channels-during-shutdown?` 
> - etc

Figuring this out was the intent of my last comment. 

My initial hunch was that we cannot change the default behavior since aleph updates are
usually relatively silent. If you feel this is warranted and provided we make it clear in
the release notes, why not. I leave the judgment of what you feel is OK to you. If the
behavior change is OK, I would indeed make it the default as it is the sanest thing to do.

The ideal option from my PoV:

- Graceful shutdown by default with a 15 second timeout interval
- No waiting on the event loop group (i.e: `(.shutdownGracefully 0 0 TimeUnit/SECONDS)`)

Control-hooks:

- `graceful-shutdown-fn`: change the hook if something else, additional 
- `graceful-shutdown?`: 
- `graceful-shutdown-timeout`: timeout period for the channel group close future, defaults to 15 seconds
- `group-shutdown-timeout`: control the parameters to `.shutdownGracefully group`, quiet period would be `(min 2 timeout)` . Defaulting to 0.

If we want to preserve the existing behavior, `graceful-shutdown?` would need to be `false` by default, and `group-shutdown-timeout` set to 15.

So in essence, I guess the two good couple of defaults most people would want to use are:

- `[false 15]`
- `[true 0]`

KingMob:
It's getting late in my TZ, so I have to go, but on one point:

> we cannot change the default behavior since aleph updates are usually relatively silent.

You're right, this is usually true, but I don't think we've ever promised any behavior in regards to shutdown, so I think we have some extra leeway here.

arnaudgeiser:
So, if we agree to change the current shutdown behaviour, we can pivot only around `shutdown-timeout` which will default to 15 seconds.

So I totally concur with this:

> The ideal option from my PoV:
Graceful shutdown by default with a 15 second timeout interval
No waiting on the event loop group (i.e: (.shutdownGracefully 0 0 TimeUnit/SECONDS))

src/aleph/http.clj
```diff
@@ -52,8 +52,8 @@
...
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
-   | `shutdown-quiet-period` | optional period in seconds for which new connections will still be serviced after a scheduled shutdown via `java.io.Closeable#close`. Defaults to 2 seconds.
-   | `shutdown-timeout` | optional grace period in seconds on which to wait for the event loop group to empty during a scheduled shutdown. Defaults to 15 seconds."
+   | `shutdown-hook` | An optional 0-arity function to run during `java.io.Closeable#close`, will be called once the server has stopped listening to incoming requests.
+   | `shutdown-hook-timeout` | Interval in seconds within which the shutdown hook must complete."
```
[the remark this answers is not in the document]
arnaudgeiser: > Removing the shutdown-quiet-period option is a breaking change. We can't just delete it.

This option never got released in any Aleph version, it has been introduced by @pyr with his last commit.
I would accept removing it.

DerGuteMoritz:
> Will users ever want/need NON-graceful shutdown? (I.e., giving channels time to wrap up)

Just to throw my 2 cents into the hat: I wanted this in the past for use in integration test suites. When there are a lot of tests which start a server at the beginning and stop it again at the end, you want to await their proper shutdown so that different tests don't collide (e.g. because they always bind to the same listening port). Having each test wait for at least 2 seconds adds up quickly then. This is a case where stopping the server as god intended doesn't work either :smile:

arnaudgeiser:
With the approach that has been discussed, it will still be supported @DerGuteMoritz.

KingMob:
Graceful by default sounds good to me. I think we have a solution.

Thanks for reminding me about test suites. Yeah, we definitely want a way to kill the server ASAP in that situation.

Having `graceful-shutdown-timeout` and `eventloop-shutdown-timeout` is OK, but the naming is confusing in regards to their relationship. Can we come up with better names? Failing that, we need some good docstrings.

I don't think we need to remove a user-supplied `shutdown-hook` option, as long as the user doesn't have to orchestrate "graceful-ness", and the docs are clear on _when_ their hook is called. OTOH, if nobody needs the hook, we can skip it until someone asks for it. I leave that up to you.

arnaudgeiser:
> I don't think we need to remove a user-supplied shutdown-hook option, as long as the user doesn't have to orchestrate "graceful-ness", and the docs are clear on when their hook is called. OTOH, if nobody needs the hook, we can skip it until someone asks for it. I leave that up to you.

Agreed, but depending on where the `shutdown-hook` is placed, it should not be an aleph responsibility, but the responsibility of your favorite dependency management library (component, integrant, mount etc...)

- Whether the `shutdown-hook` is placed between [1] and [2], it definitely has this place on Aleph
- Whether the `shutdown-hook` is placed between [2] and [3], it should be the responsibility of a dependency management library

[1] : Stop listening to incoming requests
[2] : Run the shutdown hook within the supplied timeout interval when provided. This can be used to finish processing in-flight requests for instance.
[3] : At this stage, stop the EventLoopGroup

pyr:
> Agreed, but depending on where the `shutdown-hook` is placed, it should not be an aleph responsibility, but the responsibility of your favorite dependency management library (component, integrant, mount etc...)

I agree with this, and it would speak against the current place where `on-close` gets ran. `on-close` is necessary internally but maybe it shouldn't have been exposed since it can be fully orchestrated outside of aleph.

If we were to consider changing the placement of the hook point I think it would make sense. Otherwise, after having given it more thought I think it's too many knobs to tweak for library consumers.

> Having `graceful-shutdown-timeout` and `eventloop-shutdown-timeout` is OK, but the naming is confusing in regards to their relationship. Can we come up with better names? Failing that, we need some good docstrings.

I agree with the poor naming, thinking about it, the knob that most people will want to tweak is the graceful one, so I would go with `shutdown-timeout` for this one (the existing one hasn't made it to a release yet).

After a bit of thought and looking at the feedback here, it seems is that the default parameters to `EventLoopGroup::shutdownGracefully` will work out for everyone **except** in odd scenarios like test ones, maybe we can go with `fast-eventloop-shutdown?` being a boolean. 

So to sum it up one possible strategy would be:

- `shutdown-timeout`: Controlling the amount of time the channel group
- `fast-eventloop-shutdown?`: Would set the parameters of `EventLoopGroup::shutdownGracefully` to `0, 0, TimeUnit/SECONDS`. Defaults to `false`

If everyone wants to consider the change of location of the hook I would make the current `on-close` parameter internal and use the same name for the hook that can run between server channel closing and EventLoopGroup shutdown.

arnaudgeiser:
> I agree with this, and it would speak against the current place where on-close gets ran. on-close is necessary internally but maybe it shouldn't have been exposed since it can be fully orchestrated outside of aleph.

Exactly, it should never have been exposed. But what has been done cannot be undone... :(

> I agree with the poor naming, thinking about it, the knob that most people will want to tweak is the graceful one, so I would go with shutdown-timeout for this one (the existing one hasn't made it to a release yet).

Fine for me!

> fast-eventloop-shutdown?: Would set the parameters of EventLoopGroup::shutdownGracefully to 0, 0, TimeUnit/SECONDS. Defaults to false

What about :

```clojure
(if (zero? shutdown-timeout)
  (.shutdownGracefully group 0 0)
  (.shutdownGracefully group))
```

> If everyone wants to consider the change of location of the hook I would make the current on-close parameter internal and use the same name for the hook that can run between server channel closing and EventLoopGroup shutdown.

To me, that's the way to go (if we are talking about putting the `on-close` parameter between [1] and [2] - what I understood).

[1] : Stop listening to incoming requests
[2] : Run the shutdown hook within the supplied timeout interval when provided. This can be used to finish processing in-flight requests for instance.
[3] : At this stage, stop the EventLoopGroup

pyr:
Your proposal kills the opportunity to have fast test shutdowns unfortunately. For the rest we agree.

arnaudgeiser:
And why not `(.shutdownGracefully group 0 0)` in all cases?
Whether waiting for active connections is the default, it's acceptable to have this behavior as non-parameterizable, no?

pyr:
I adapted to the proposed approach, taking into account @arnaudgeiser's comment

test/aleph/http_timeout_test.clj
```diff
@@ -0,0 +1,97 @@
...
+
+(deftest test-shutdown-timeout-3
+  (testing "shutdown with a timeout of 1 seconds with streaming body"
+    (let [server (http/start-server streaming-handler (server-options :shutdown-timeout 1))]
+      (try
+        (let [resp (http-get)]
+          (Thread/sleep 50) ;; wait a bit for the request to be initiated
+          (.close ^AutoCloseable server)
+          (is (= 200 (:status @resp)))
+          (is (= "0.1.2.3.4.5." (bs/to-string (:body @resp)))))
```
arnaudgeiser, line 82: Is it the correct behavior?
200 and only part of the expected response?

src/aleph/netty.clj
```diff
@@ -1298,10 +1331,25 @@ initialize an DnsAddressResolverGroup instance.
            java.io.Closeable
            (close [_]
              (when (compare-and-set! closed? false true)
+               ;; This is the three step closing sequence:
+               ;; 1. Stop listening to incoming requests
                (-> ch .close .sync)
-               (-> group (.shutdownGracefully (long shutdown-quiet-period)
-                                              (long shutdown-timeout)
-                                              TimeUnit/SECONDS))
+               (-> (if (pos? shutdown-timeout)
```
arnaudgeiser, line 1337: All the other integration tests are damn slow on Aleph now.
I would consider adding `shutdown-timeout 0` everywhere else where we are starting a server.

arnaudgeiser (approved):
I'm really happy with the state of this PR.
Approving but would be nice to do the adaptations on the other tests to make them fast again.

src/aleph/netty.clj
```diff
@@ -693,6 +694,17 @@
             (.readableBytes ^ByteBuf msg)))
         (.write ctx msg promise)))))
 
+(defn- ^ChannelHandler channel-tracking-handler
+  "Yields an inbound handler, ready to be added to a pipeline,
+   which keeps track of requests in a provided channel group.
+   The channel-group can be created via `channel-group`."
```
arnaudgeiser, line 700:
```suggestion
   The channel-group can be created via `make-channel-group`."
```

src/aleph/netty.clj
```diff
@@ -703,6 +715,16 @@
     (.remove pipeline handler))
   pipeline)
 
+(defn- append-handler-to-pipeline
+  "Convenience function to add a handler to the tail of a netty pipeline"
+  [^ChannelPipeline pipeline handler-id ^ChannelHandler handler]
+  (.addLast pipeline (str handler-id) handler))
+
+(defn- prepend-handler-to-pipeline
+  "Convenience function to add a handler to the tail of a netty pipeline"
```
arnaudgeiser, line 724:
```suggestion
  "Convenience function to add a handler to the head of a netty pipeline"
```

test/aleph/http_timeout_test.clj
pyr: For a streaming body yes. We waited for the grace period, then closed the connection. streaming-handler says 200 right away then yields a stream for the body.

src/aleph/netty.clj
pyr: Ok will do (much later today though)

pyr:
> Approving but would be nice to do the adaptations on the other tests to make them fast again.
> Furthermore, the title of the PR has to be changed to reflect what has been done.

I will do now that the direction seems to work for everyone, once I get a second approval I will also do a last force push to have proper commit messages

src/aleph/http.clj
```diff
@@ -52,8 +52,7 @@
...
    | `continue-handler` | optional handler which is invoked when header sends \"Except: 100-continue\" header to test whether the request should be accepted or rejected. Handler should return `true`, `false`, ring responseo to be used as a reject response or deferred that yields one of those.
    | `continue-executor` | optional `java.util.concurrent.Executor` which is used to handle requests passed to :continue-handler.  To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread.
-   | `shutdown-quiet-period` | optional period in seconds for which new connections will still be serviced after a scheduled shutdown via `java.io.Closeable#close`. Defaults to 2 seconds.
-   | `shutdown-timeout` | optional grace period in seconds on which to wait for the event loop group to empty during a scheduled shutdown. Defaults to 15 seconds."
+   | `shutdown-timeout` | interval in seconds within which in-flight requests must be processed, defaults to 15 seconds. A value of 0 bypasses waiting entirely."
```
KingMob, line 55: Maybe this got lost in the conversation, but we shouldn't just delete an existing option. Can we make `shutdown-quiet-period` a subordinate alias for `shutdown-timeout`. (Subordinate in the sense that `shutdown-timeout wins if both are present.)

We can then either deprecate the option in the docstring, or remove it from the docstring entirely.

src/aleph/netty.clj
```diff
@@ -1226,33 +1248,49 @@ initialize an DnsAddressResolverGroup instance.
...
+     pipeline "ssl-hanlder"
+     (.newHandler ^SslContext ssl-ctx (-> pipeline .channel .alloc)))
+    (pipeline-builder pipeline)))
+
+(defn- add-channel-tracker
+  [pipeline-builder chan-group]
+  (fn [pipeline]
+    (prepend-handler-to-pipeline pipeline "channel-tracker"
+                                (channel-tracking-handler chan-group))
+    (pipeline-builder pipeline)))
```
KingMob, line 1264: For naming parity, `add-ssl-handler` and `add-channel-tracker` should match. Maybe `add-channel-tracker-handler`?

src/aleph/tcp.clj
```diff
@@ -80,15 +80,13 @@
    | `bootstrap-transform` | a function that takes an `io.netty.bootstrap.ServerBootstrap` object, which represents the server, and modifies it.
    | `pipeline-transform` | a function that takes an `io.netty.channel.ChannelPipeline` object, which represents a connection, and modifies it.
    | `raw-stream?` | if true, messages from the stream will be `io.netty.buffer.ByteBuf` objects rather than byte-arrays.  This will minimize copying, but means that care must be taken with Netty's buffer reference counting.  Only recommended for advanced users.
-   | `shutdown-quiet-period` | optional period in seconds for which new connections will still be serviced after a scheduled shutdown via `java.io.Closeable::close`. Defaults to 2 seconds.
-   | `shutdown-timeout` | optional grace period in seconds on which to wait for the event loop group to empty during a scheduled shutdown. Defaults to 15 seconds.  "
+   | `shutdown-timeout` | interval in seconds within which in-flight requests must be processed, defaults to 15 seconds. A value of 0 bypasses waiting entirely."
```
KingMob, line 83: Same as above

KingMob (changes requested):
Looking good, I think we're almost there.

Thanks for all your hard work!

src/aleph/http.clj
pyr: It wasn't lost, I'm on @arnaudgeiser's side: we added this option two commits ago (9 days ago) and it never made it into a release, so there are no consumers out there

src/aleph/tcp.clj
pyr: Likewise :-)

src/aleph/http.clj
arnaudgeiser: For the record : https://github.com/clj-commons/aleph/pull/642#discussion_r1018233370

src/aleph/netty.clj
pyr: done

pyr:
All comments have been addressed except for @KingMob's regarding `shutdown-quiet-period` which I think doesn't apply. Once we find an agreement on this last one I'll rework the commit messages and push a cleaned rebased set of commits.

@arnaudgeiser tests are indeed way faster now :-)

Thanks for the thorough reviews everyone!

src/aleph/http.clj
KingMob: Ah, OK, I missed Arnaud's comment about it. nm

KingMob (commented):
I withdraw my comment about the extra option. 

Lmk when the final version is ready to look at.

pyr:
@KingMob The code is ready for review, I will leave the rebase + force-push for after your review/approval

src/aleph/netty.clj
```diff
@@ -1226,33 +1248,49 @@ initialize an DnsAddressResolverGroup instance.
               (let [ch (.channel ^ChannelFuture f)]
                 (maybe-ssl-handshake-future ch)))))))))
 
+(defn- add-ssl-handler
+  [pipeline-builder ssl-ctx]
+  (fn [^ChannelPipeline pipeline]
+    (append-handler-to-pipeline
+     pipeline "ssl-hanlder"
```
KingMob, line 1255: typo in "ssl-hanlder"

test/aleph/http_test.clj
```diff
@@ -180,25 +180,25 @@
            (.shutdown *pool*))))))
 
 (defmacro with-handler [handler & body]
-  `(with-server (http/start-server ~handler {:port port})
+  `(with-server (http/start-server ~handler {:port port :shutdown-timeout 0})
```
KingMob, line 183: Wouldn't it be simpler to add this to the `default-options` map at the top?

KingMob (approved):
@pyr Last minor details: please fix the handler name typo, and I'd suggest moving the timeout to the `default-options` maps in the test namespaces.

Thanks for all your work on this!

test/aleph/http_test.clj
pyr: `default-options` at the top is used for HTTP clients, not servers.

KingMob: We should probably make one for servers, then. Still, if you're sick of this, it doesn't have to be in this PR 😄

pyr: I'll re-prepare the commits this morning and then will add this.

pyr: After a second read I think I will not tackle this refactor for this PR

pyr:
OK, from my perspective it's now clean and ready to go.
A side question while I have your attention: the fact that external contributions do not get their tests ran is a bit inconvenient, would you be open to a PR that runs tests from Github actions which wouldn't need any particular setup? This could be done *in addition* to the Circle CI one but at least provide a functional test run for contributions such as this PR

arnaudgeiser (approved):
I confirm, those tests are running pretty fast now.

```
❯ time lein test
Compiling 5 source files to /home/arnaudgeiser/code/aleph/target/classes
warning: [options] bootstrap class path not set in conjunction with -source 8
1 warning
Compiling 5 source files to /home/arnaudgeiser/code/aleph/target/classes
warning: [options] bootstrap class path not set in conjunction with -source 8
1 warning

lein test aleph.classloader-test

lein test aleph.http-continue-test

lein test aleph.http-test

lein test aleph.http-timeout-test

lein test aleph.http.client-middleware-test

lein test aleph.http.client-test

lein test aleph.http.core-test

lein test aleph.http.multipart-test

lein test aleph.netty-test

lein test aleph.ring-test

lein test aleph.tcp-ssl-test

lein test aleph.tcp-test

lein test aleph.udp-test

lein test aleph.websocket-test

Ran 82 tests containing 923 assertions.
0 failures, 0 errors.
lein test  59.93s user 2.20s system 240% cpu 25.833 total
```

PR #651, opened 2023-01-02
multipart: support mime-type when encoding memory data

### LINKED ISSUE #649: The client ignores content type when using byte arrays in multipart data
`clj-http`  uses `org.apache.http.entity.mime.content.ByteArrayBody` for byte arrays and it allows setting content type.

Netty uses attributes and file uploads. The former don't allow setting content type, the latter do.
The issue with Aleph is that it uses file uploads only for on-disk files, even though Netty provides `io.netty.handler.codec.http.multipart.MemoryFileUpload`.

Seems like the solution here is to change `aleph.http.multipart/encode-request` in such a way so that:
- For on-disk files it uses the current approach with `.addBodyFileUpload`
- For in-memory data with no content type it also uses the current approach with `MemoryAttribute`
- For in-memory data with content type it creates `MemoryFileUpload`

arnaudgeiser:
Fixes #649

### DIFF
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -179,7 +179,7 @@
                       (.addContent (netty/to-byte-buf content) true)))]
     (.addBodyHttpData encoder attr)))
 
-(defn encode-request [^DefaultHttpRequest req parts]
+(defn ^:no-doc encode-request [^DefaultHttpRequest req parts]
   (let [^HttpPostRequestEncoder encoder (HttpPostRequestEncoder. req true)]
     (doseq [{:keys [content mime-type] :as part} parts]
       (cond
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -10,7 +10,8 @@
    [manifold.stream :as s])
   (:import
    [java.util
-    Locale]
+    Locale
+    UUID]
    [java.io
     File]
    [java.nio
@@ -28,6 +29,7 @@
    [io.netty.handler.codec.http.multipart
     Attribute
     MemoryAttribute
+    MemoryFileUpload
     FileUpload
     DefaultHttpDataFactory
     HttpPostRequestDecoder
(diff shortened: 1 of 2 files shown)


KingMob (commented):
Looks good, but should there be tests to verify it's encoding as expected?

arnaudgeiser:
Perfect, going to work on it later during the week-end to add more tests.
I wanted to validate the idea first.

arnaudgeiser:
I had to rebase on master (why it was force-pushed).

KingMob (approved):

src/aleph/http/multipart.clj
```diff
@@ -127,37 +129,68 @@
...
+                        content
+                        content-type
+                        false)))
+
+(defn- encode-memory-file-upload
+  "Encode the `content` as a `MemoryFileUpload`."
+  [^HttpPostRequestEncoder encoder {:keys [part-name name content mime-type charset]}]
+  (let [content-type (mime-type-descriptor mime-type charset)
+        charset      (ensure-charset charset)
+        filename     (str (UUID/randomUUID))
```
p-himik, line 164: Just saw this. Why is `filename` a random string? Why not expose it as a parameter to achieve parity with Apache's `ByteArrayBody`?

KingMob: Arnaud will know better, but my guess is it's because it's for a MemoryFileUpload, which mimics a file in memory, and as such, the "filename" is just for compatibility with the FileUpload interface, since there's not really a file.

OTOH, I can imagine a real filename being useful. Want to submit a PR?

arnaudgeiser: > Arnaud will know better, but my guess is it's because it's for a MemoryFileUpload, which mimics a file in memory, and as such, the "filename" is just for compatibility with the FileUpload interface, since there's not really a file.

Exactly, that was the intent behind it.

p-himik: Created a PR: https://github.com/clj-commons/aleph/pull/668

PR #652, opened 2023-01-03
netty: do not push bindings on operation-complete

arnaudgeiser:
## Description

This removes the usage of `bound-fn*` which has been introduced by the following PR. [1]
It was added for safety while there might be no open issue regarding it.
For some workloads, this operation can take up to 40% of CPU.

Here is the reason why we used to push the bindings [2], but I suppose we fixed it by using `ensure-dynamic-classloader` instead.

image

Here is the gist of the discussion we had with @KingMob :

> Oh yeah, because the code runs on netty and signal threads that aren’t started by the Clojure test, so there’s no automatic thread frame propagation, defeating the use of binding under the hood. Well, the usual solution to interop with Java threads is bound-fn , and that works here. Truthfully, we could probably use bound-fn more when handing things off to Netty to run, it’s always safer.
Anyway, this works:

```clojure
(deftest test-classloader
  (testing "classloader: ensure the class loader is always a DynamicClassLoader"
    (let [result (CompletableFuture.)]
      (with-dynamic-redefs [netty/operation-complete (partial operation-complete result)]
        (let [server (http/start-server
                      (constantly {:body "ok"})
                      {:port 9999})]
          (on-signal :int
                     (bound-fn [_]
                       (.close ^java.io.Closeable server)))
          (.exec (Runtime/getRuntime) (format "kill -SIGINT %s" (pid)))
          (is (= (deref result 30000 ::timeout) true)))))))
```

```clojure
(defn wrap-future
  [^Future f]
  (when f
    (if (.isSuccess f)
      (d/success-deferred (.getNow f) nil)
      (let [d (d/deferred nil)
            bound-operation-complete (bound-fn* operation-complete)]
        (.addListener f
          (reify GenericFutureListener
            (operationComplete [_ _]
              (ensure-dynamic-classloader)
              (bound-operation-complete f d))))
        d))))
```

> [9 h 28] The use of bound-fn in the test ensures it inherits the dynamic redef frame, and bound-fn* in wrap-future ensures the listener will work even if .close->wrap-future is called from a signal thread
[> 9 h 29] We shold probably check everywhere we accept user callbacks for netty and wrap them in bound-fn for safety

[1] : https://github.com/clj-commons/aleph/pull/604
[2] : https://github.com/clj-commons/aleph/pull/425

### DIFF
--- a/test/aleph/classloader_test.clj
+++ b/test/aleph/classloader_test.clj
@@ -3,7 +3,6 @@
    [aleph.http :as http]
    [aleph.netty :as netty]
    [clojure.test :refer [deftest is testing]]
-   [dynamic-redef.core :refer [with-dynamic-redefs]]
    [manifold.deferred :as d]
    [manifold.utils :refer [when-class]]
    [signal.handler :refer [on-signal]])
@@ -41,7 +40,7 @@
 (deftest test-classloader
   (testing "classloader: ensure the class loader is always a DynamicClassLoader"
     (let [result (CompletableFuture.)]
-      (with-dynamic-redefs [netty/operation-complete (partial operation-complete result)]
+      (with-redefs [netty/operation-complete (partial operation-complete result)]
         (let [server (http/start-server
                       (constantly {:body "ok"})
                       {:port 9999 :shutdown-timeout 0})]
--- a/test/aleph/classloader_test.clj
+++ b/test/aleph/classloader_test.clj
@@ -45,6 +45,6 @@
                       (constantly {:body "ok"})
                       {:port 9999 :shutdown-timeout 0})]
           (on-signal :int
-                     (bound-fn [_] (.close ^java.io.Closeable server)))
+                     (fn [_] (.close ^java.io.Closeable server)))
           (.exec (Runtime/getRuntime) (format "kill -SIGINT %s" (pid)))
           (is (= (deref result 10000 ::timeout) true)))))))
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,6 @@
                                    [org.slf4j/slf4j-simple "1.7.30"]
                                    [com.cognitect/transit-clj "1.0.324"]
                                    [spootnik/signal "0.2.4"]
-                                   [me.mourjo/dynamic-redef "0.1.0"]
                                    ;; This is for self-generating certs for testing ONLY:
                                    [org.bouncycastle/bcprov-jdk18on "1.72"]
                                    [org.bouncycastle/bcpkix-jdk18on "1.72"]]
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -250,15 +250,11 @@
     (let [d (d/deferred nil)]
       (if (.isDone f)
         (operation-complete f d)
-        ;; Ensure the same bindings are installed on the Netty thread (vars,
-        ;; classloader) than the thread registering the
-        ;; `operationComplete` callback.
-        (let [bound-operation-complete (bound-fn* operation-complete)]
-          (.addListener f
-                        (reify GenericFutureListener
-                          (operationComplete [_ _]
-                            (ensure-dynamic-classloader)
-                            (bound-operation-complete f d))))))
+        (.addListener f
+                      (reify GenericFutureListener
+                        (operationComplete [_ _]
+                          (ensure-dynamic-classloader)
+                          (operation-complete f d)))))
       d)))
 
 (defn allocate [x]


KingMob:
Just breezing through on a spare moment. I haven’t had time to take a closer look, but I did notice one mistake: the Clojure classloader isn’t responsible for thread frame propagation. They’re separate issues, and using the classloader alone isn’t sufficient. Even with the classloader in place, binding frame propagation requires the use of Clojure fns to start threads; with Java code, frames won’t be propagated.

arnaudgeiser:
Yes, I totally agree with your observation and your conclusion.
The question is more: Do we really need thread bindings propagation?
That's unlikely people using an asynchronous mechanism expect to have the bindings of the calling threads propagated down, right? And we are talking only about dynamic bindings, here, right?

I mean, a lot of people are not relying on manifold and use CachedThreadPool or FixedThreadPool directly without trouble.
I have the impression that "we" (not us actually) have done a lot of tinkering around this issue in the past... I would like to reconsider the status-quo.

KingMob:
OK, I took a closer look to refresh my memory, and I'd forgotten why we started frame propagation in the first place.

While it's true that the classloader is not responsible for frame propagation, I had the direction reversed. It's the classloader we care about, not frame propagation _per se_. We were only propagating frames to try and make the `LOADER` var used by the compiler available to the signal handler thread. (I think.)

We need the Clojure class loader set for all netty threads (not just the signal handler thread) but we switched from setting the LOADER var, to using `.setContextClassLoader` on the thread. I _think_ that should be sufficient. The comments say we use `bound-fn*` to "Ensure the same bindings are installed on the Netty thread (vars, classloader)", except... the `operation-complete` fn we're binding doesn't use *any* dynamic vars itself. Maybe there was a concern that the call to `.getNow()` would trigger some code, but I just looked at all definitions in Netty, and that's not an issue.

All of which is to say, yeah, we can probably stop using `bound-fn` on `operation-complete`

test/aleph/classloader_test.clj
```diff
@@ -41,7 +40,7 @@
 (deftest test-classloader
   (testing "classloader: ensure the class loader is always a DynamicClassLoader"
     (let [result (CompletableFuture.)]
-      (with-dynamic-redefs [netty/operation-complete (partial operation-complete result)]
+      (with-redefs [netty/operation-complete (partial operation-complete result)]
```
KingMob, line 43: This is an unrelated change. Essentially, `with-dynamic-redefs` is always preferable to plain `with-redefs` in testing, and wasn't related to whether we use `bound-fn` or not.

KingMob (changes requested):

test/aleph/classloader_test.clj
arnaudgeiser: Not sure your comment is accurate here.
But I will have a look. With `with-dynamic-redefs`, the `netty/operation-complete` was not available on the Netty threads and it failed.

I will refactor the test a bit.

arnaudgeiser:
> All of which is to say, yeah, we can probably stop using bound-fn on operation-complete

Perfect, polishing the PR then.

test/aleph/classloader_test.clj
arnaudgeiser: Mhhh... I don't know exactly how to fix this.
The test was relying on the fact we were using `bound-fn*` actually.

```clojure
(let [bound-operation-complete (bound-fn* operation-complete)]   <== Here we are on a SIGINT Thread
  (.addListener f
                (reify GenericFutureListener
                  (operationComplete [_ _]
                    (ensure-dynamic-classloader)
                    (bound-operation-complete f d)))))) <== Here we are on a FastThreadLocalThread
```

Without using `bound-fn*`, I have no idea how to propagate the local binding of `operation-complete` to the `FastThreadLocalThread`. 

Any help is welcome here.

arnaudgeiser:
For the record @KingMob, here is where everything started : https://github.com/clj-commons/aleph/pull/425, where we put two mechanisms in place.

test/aleph/classloader_test.clj
KingMob: ...or did switching from `bound-fn` to `fn` in `classloader-test` fix it, because CircleCI isn't complaining?

arnaudgeiser: > Hmmm, test-classloader is failing? Are you saying it fails using with-dynamic-redefs, but works using with-redefs? 

Yes, it used to work thanks to `bound-fn*`.
This test was actually doing this job quite nicely, ensuring the bindings of the calling threads were pushed to the callback. Since we are no longer using `bound-fn*`, we won't have access to the `netty/operation-complete` rebinding which is present on the `SIGINT Thread` only.
Having to use `with-redefs` makes sense to me here.

> ...or did switching from bound-fn to fn in classloader-test fix it,

This `bound-fn` was actually to push the bindings from the `test` Thread to the `SIGINT Thread` which was then propagated with `bound-fn*` when calling operation-complete.
Since `with-redefs` is used, this is no longer necessary. (it was just a clean up commit)

> ...or did switching from bound-fn to fn in classloader-test fix it,

Tests are passing, like it was the case initially. But I have no idea how I can keep using `with-dynamic-redefs` to achieve the
same behavior...

KingMob: Well, something is wrong, because it *should* work, but I can't keep looking at it. I think it's due to the use of non-Clojure threads again (both signal and netty)

My suspicion is that `with-dynamic-redefs` is failing because internally, it relies on `binding` to store the thread-specific override map, but `binding`'s thread frames don't play well with non-Clojure threads, so I think the overrides are never seen, in which case, `with-dynamic-redefs` will fall back to the original, which won't set our test Future.

`with-redefs` works in this case because it truly replaces the root definition, instead of adding some thread-based override-enabling functionality.

Let's just go with `with-redefs` for now.

KingMob (approved):

test/aleph/classloader_test.clj
arnaudgeiser: > My suspicion is that with-dynamic-redefs is failing because internally, it relies on binding to store the thread-specific override map

That's the point. It uses local bindings. So any others threads cannot see those bindings unless we propagate.

KingMob: That's only true for non-Clojure threads. And there's no requirement for that, it should be possible to place the override machinery directly in the root var, so it would work, regardless of thread.

KingMob: Hmm, `with-dynamic-redefs` is more limited than I recall. It only works with fns, for starters. Might be able to use `bound-fn` tho.

Real shame Hickey started requiring `^:dynamic` back in 1.3. Used to be simpler.

PR #653, opened 2023-01-06
Update examples and docs

### LINKED ISSUE #629: Update examples and docs
First, while the aleph.io literate documentation site is cool, we aren't currently keeping it up-to-date, and would have to bug Zach for access if we wanted to. On top of that, in recent years, cljdoc.org has emerged as the standard bearer for clj documentation. I'd like to convert the examples and literate docs into articles or something so that users of cljdoc.org can read them.

Along with that, the current routing in `examples` is based off Compojure, which is not very popular these days. I'd like to switch to reitit, or something more data-driven.

To do:
- [x] (Optional) Get local cljdoc building working, if you haven't before
- [x] Bring aleph.io docs to cljdoc.org
- [x] Convert example routing to reitit
- [x] Fix the examples' to follow the clojure style guide

arnaudgeiser:
Fixes #629

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Aleph attempts to mimic the clj-http API and capabilities fully. It supports mul
 
 Aleph client also supports fully async and [highly customizable](https://github.com/clj-commons/aleph/blob/d33c76d6c7d1bf9788369fe6fd9d0e56434c8244/src/aleph/netty.clj#L783-L796) DNS resolver.
 
-To learn more, [read the example code](http://aleph.io/examples/literate.html#aleph.examples.http).
+To learn more, [read the example code](https://github.com/clj-commons/aleph/blob/master/examples/src/aleph/examples/http.clj).
 
 ### WebSockets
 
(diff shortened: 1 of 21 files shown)


KingMob (approved):
👍🏻 Nice work!

Have you tested it with [local cljdoc](https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally-author.adoc)? Fair warning, it can get a bit tricky. I have some scripts for deps.edn-based repos, but not lein.

arnaudgeiser:
> Have you tested it with [local cljdoc](https://github.com/cljdoc/cljdoc/blob/master/doc/running-cljdoc-locally-author.adoc)? Fair warning, it can get a bit tricky. I have some scripts for deps.edn-based repos, but not lein.

Happy to hear that, because it took me a while to understand how to work with a local cljdoc.
The documentation is great, but I didn't get it initially.
Yet, it has been tested.

Please note I had an heavy hand with `^:no-doc` here.

Here is the result :

image

image

image

image

image

image

image

image

arnaudgeiser:
Please let a thumbs up on this comment if you are fine with my choice.
I removed more than 50% of the functions on the documentation, and the whole `http/server.clj`, `http/core.clj`, `http/client.clj`, `http/encoding.clj` namespaces.

I not able to remove `->Cookie` from the documentation as potemkin `p/derived-map` is not supporting metadata so far.

arnaudgeiser:
Actually, it's just documentation.
If I've been too aggressive, we'll change it back.
Merging.

PR #654, opened 2023-01-07
Add support for kqueue and io-uring

### LINKED ISSUE #526: Add support for other native transports - io_uring and kqueue
Can I request a `kqueue` backend? It looks to me pretty simple tack on since we already have an `epoll` interface defined. Also, I should mention, while employing `:epoll? true` we should let people know they need to include `[io.netty/netty-transport-native-epoll "4.1.36.Final"]` or a matching version of said transport that matches the Netty version utilized, for `kqueue`: `s/epoll/kqueue`.

arnaudgeiser:
This adds the support of both `kqueue` and `io-uring` as a transport parameter for the HTTP/TCP/UDP servers and related clients the way describes on #526.

Closes #526


project.clj
```diff
@@ -16,6 +16,8 @@
                  [io.netty/netty-transport ~netty-version]
                  [io.netty/netty-transport-native-epoll ~netty-version :classifier "linux-x86_64"]
                  [io.netty/netty-transport-native-epoll ~netty-version :classifier "linux-aarch_64"]
+                 [io.netty/netty-transport-native-kqueue ~netty-version :classifier "osx-x86_64"]
+                 [io.netty.incubator/netty-incubator-transport-native-io_uring "0.0.12.Final" :classifier "linux-x86_64"]
```
KingMob, line 20: Is the classifier required? Or conversely, are there any ARM versions of io_uring we need to add?

src/aleph/http.clj
```diff
@@ -45,7 +45,7 @@
    | `max-initial-line-length` | the maximum characters that can be in the initial line of the request, defaults to `8192`
    | `max-header-size`         | the maximum characters that can be in a single header entry of a request, defaults to `8192`
    | `max-chunk-size`          | the maximum characters that can be in a single chunk of a streamed request, defaults to `16384`
-   | `epoll?`                  | if `true`, uses `epoll`, defaults to `false`.
+   | `transport`               | the transport to use, one of `:epoll`, `:kqueue` or `:io-uring`
```
KingMob, line 48: Mention which one it defaults to, if unspecified

src/aleph/netty.clj
```diff
@@ -1053,18 +1082,51 @@
...
+    :io-uring (IOUringEventLoopGroup. num-threads thread-factory)
+    :nio      (NioEventLoopGroup. num-threads thread-factory)))
+
+(defn ^:no-doc transport-server-channel [transport]
+  (case transport
+    :epoll    EpollServerSocketChannel
+    :kqueue   KQueueServerSocketChannel
+    :io-uring IOUringServerSocketChannel
+    :nio      NioServerSocketChannel))
+
```
KingMob, line 1129: It feels like we have a bunch of these maps, maybe there should be a ns-wide transport map that groups the values together?

src/aleph/http.clj
arnaudgeiser: > :nio should be added for completeness, and we should mention that :transport defaults to it, if unspecified.

Agreed, I will modify the docstrings accordingly.

> Maybe we should leave in the epoll? option, but deprecate it, and say that :transport supersedes it?

I would remove the `epoll?` option, that way, no-one will ever use it in the future.

src/aleph/netty.clj
arnaudgeiser: That approach minimizes the number of changes across all the namespaces.
I don't think we'll have to support another transport, even so, the changes will be pretty obvious.
I would advice keeping it like this.

Anyway, do not hesitate to push commits on top of this PR with your approach. If I do it myself, I would expect a lot of back and forth.

project.clj
arnaudgeiser: Good question...

According to the documentation on GitHub, it was the _recommendation_ [1].
But, if you look at Maven registry, we are also behind in terms of dependency [2] and the classifier is not mentioned.

I will try to remove it and bump to version "0.0.16.Final".

[1] : https://github.com/netty/netty-incubator-transport-io_uring
[2] : https://mvnrepository.com/artifact/io.netty.incubator/netty-incubator-transport-native-io_uring/0.0.16.Final

arnaudgeiser: But yes, there is an ARM version

```
x86-64 / aarch64 processor
```

src/aleph/netty.clj
KingMob: This works, too.

project.clj
arnaudgeiser: Yes, the classifier is mandatory, otherwise you end up with a `jar` containing only some metadata. [1]
I will add both classifier.

[1] : https://repo1.maven.org/maven2/io/netty/incubator/netty-incubator-transport-native-io_uring/0.0.16.Final/

KingMob (approved):

PR #656, opened 2023-01-08
multipart: manual cleanup to handle temporary files

arnaudgeiser:
## Description

This is a followup PR for #596 (which has been closed in favor of this) to propose a less risky and a backwards compatible approach for multipart requests.

Using temporary files storage is broken with the current `release all resources on stream close` mechanism. Fixing it is barely feasible without introducing a breaking change. 
Still, the current mechanism is really convenient: you consume the stream and you are done! It's working perfectly fine as long as you keep all your data on memory. 
__=> Let's keep it that way.__

I propose here to introduce a `manual-cleanup?` flag (a better name can be proposed) which will return a vector composed of a manifold.stream and a `cleanup` function which can be called after the stream has been consumed.

This PR doesn't aim at bringing a file granularity clean up function (which I think is not necessary) as proposed by #596.

And there is also a cljdoc article attached to this PR : https://github.com/clj-commons/aleph/compare/multipart-requests?expand=1#diff-45f6713a5206dde4c1c69d8ddc5242e5ed09e8256ecec0ab80ff911bb8e967c3R1-R82

Closes : #567

### DIFF
--- a/src/aleph/http/multipart.clj
+++ b/src/aleph/http/multipart.clj
@@ -227,9 +227,20 @@
      :file (when-not memory? (.getFile data))
      :size (.length data)}))
 
-(defn- read-attributes [^HttpPostRequestDecoder decoder parts]
+(defn- read-attributes [^HttpPostRequestDecoder decoder parts manual-cleanup?]
   (while (.hasNext decoder)
-    (s/put! parts (http-data->map (.next decoder)))))
+    (let [{:keys [file] :as part} (http-data->map (.next decoder))]
+      (when (and (not manual-cleanup?) file)
+        (log/warn (str "Temporary files storage is not working using automatic clean up. consider using manual cleanup.` "
+                       "See: https://cljdoc.org/d/aleph/aleph/CURRENT/doc/http/handling-multipart-requests")))
+      (s/put! parts part))))
+
+(defn- destroy-decoder [^HttpPostRequestDecoder decoder destroyed?]
+  (when (compare-and-set! destroyed? false true)
+    (try
+      (.destroy decoder)
+      (catch Exception e
+        (log/warn e "exception when cleaning up multipart decoder")))))
 
 (defn decode-request
   "Takes a ring request and returns a manifold stream which yields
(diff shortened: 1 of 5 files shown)


src/aleph/http/multipart.clj
```diff
@@ -206,23 +217,32 @@
...
    it's better to set `:raw-stream?` to `true` to avoid additional
    input stream coercion."
   ([req] (decode-request req {}))
   ([{:keys [body] :as req}
     {:keys [body-buffer-size
-            memory-limit]
+            memory-limit
+            manual-cleanup?]
      :or {body-buffer-size 65536
           memory-limit DefaultHttpDataFactory/MINSIZE}}]
```
KingMob, line 236: We should be explicit and place `:manual-cleanup? false` in the `:or` map.

test/aleph/http/multipart_test.clj
```diff
@@ -188,52 +189,132 @@
...
+      {:status 200
+       :body (pr-str body)})))
+
+(defn- make-decode-manual-handler [options]
+  (fn [req]
+    (let [[parts cleanup-fn] (mp/decode-request req options)
+          body  (mapv (fn [{:keys [file] :as part}]
+                        (Thread/sleep 20) ;; wait a bit to ensure the files are cleaned when using automatic resource cleanup
+                        (let [packed-part (pack-part part)]
+                          (when (some? file)
```
KingMob, line 219: Trivial: `when-some`

KingMob (approved):
Approved in the interest of no more delays, but let's add the manual cleanup param to the default param map

test/aleph/http/multipart_test.clj
arnaudgeiser: I will use `when file` instead, `when-some` imposes you to create a binding

arnaudgeiser:
I had to rebase on master (and handle merge conflicts), but your comments have been addressed.
I also bumped to Clojure 1.11 to have access to `parse-uuid`.

test/aleph/http/multipart_test.clj
KingMob: Ahh, right, right. A pattern-match failure of the brain!

arnaudgeiser:
Final rebase and merge.

PR #657, opened 2023-01-09
doc: no-doc on ->Cookie

arnaudgeiser:
I find out how to add `no-doc` on `->Cookie` to not expose it on the documentation.


KingMob (approved):
You'd think placing that on `defrecord` might propagate it to the constructor fns, but nope.

PR #661, opened 2023-01-27
Randomize port for TCP tests

arnaudgeiser:
## Description

This ensures port are randomized for udp and tcp servers.
It helps to stabilize the tests when some servers take a little bit of time to be shutdown.

### DIFF
--- a/test/aleph/tcp_test.clj
+++ b/test/aleph/tcp_test.clj
@@ -2,6 +2,7 @@
   (:require
    [aleph.netty :as netty]
    [aleph.tcp :as tcp]
+   [aleph.utils :refer [rand-port]]
    [clj-commons.byte-streams :as bs]
    [clojure.test :refer [deftest testing is]]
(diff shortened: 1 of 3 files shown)


KingMob (changes requested):
This is a good idea, but the implementation isn't quite safe. Since `rand-port` closes the ServerSocket, the port becomes available again, and could be snapped up before we start the server.

I double-checked the docs, and if you pass 0 for `:port` to `start-server`, it will do what we want: find an unused port and grab it simultaneously.

arnaudgeiser:
Yes, definitely, it will work for the server.
But we need to be aware of the port used so it can be exercised by the client on the tests.
Do you have a proposition here?

KingMob:
The object returned from `start-server` implements the AlephServer protocol, which has a method `port`, which returns the port number.

KingMob (approved):

Issue #236, opened 2016-04-24
Add `charset` to `text/plain` responses


pbwolf:
The README's HTTP server example

```
(defn handler [req]
  {:status 200
   :headers {"content-type" "text/plain"}
   :body "hello!"})
```

works because "hello!" consists of 7-bit Latin-1 characters.
Substitute "¡Hola!" and a problem emerges: the web browser displays
"Â¡Hola!".

The cause is a conflict between the Content-Type header and Aleph's
encoding of the response body.  Aleph serves the Content-Type as the
sample says, without a charset, which according to RFC2616 means
ISO-8859-1.[1] But Aleph encodes all String responses in UTF-8
regardless of the Content-Type.  Converting the code points to UTF-8
bytes and then interpreting the bytes as ISO-8859-1 results in the
spurious LATIN CAPITAL LETTER A WITH CIRCUMFLEX.

This behavior is unexpected because it corrupts the response message
in a way that is not required by the Ring SPEC [2].  True, SPEC does
not specify how the String should be encoded, but nor does it require
serving the headers exactly as given!  So a Ring-compliant server may
either make the body to suit the headers, or make the headers to suit
the body.

Ring's Jetty server obeys Content-Type's stated charset, with a
default of ISO-8859-1.  An alternative would be to force
"charset=utf-8" into the Content-Type when serving a String, if the
String is always going to be UTF-8.

[1] HTTP/1.1, RFC 2616, says:  When no explicit charset parameter is
provided by the sender, media subtypes of the "text" type are defined
to have a default charset value of "ISO-8859-1" when received via
HTTP. https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html

[2] https://github.com/ring-clojure/ring/blob/master/SPEC

arnaudgeiser:
> The "charset" parameter is used with some media types to define the character set (section 3.4) of the data. When no explicit charset parameter is provided by the sender, media subtypes of the "text" type are defined to have a default charset value of "ISO-8859-1" when received via HTTP. Data in character sets other than "ISO-8859-1" or its subsets MUST be labeled with an appropriate charset value. See section [3.4.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.4.1) for compatibility problems.

KingMob:
The current RFC on the matter is https://www.rfc-editor.org/rfc/rfc6657.html.

[...]

PR #662, opened 2023-01-29
Add charset to text/plain responses

arnaudgeiser:
This adds `charset=UTF-8` to every `text/plain` responses.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -111,8 +111,11 @@
            (.printStackTrace e (java.io.PrintWriter. w))
            (str w))})
 
-(let [[server-name server-value connection-name keep-alive-value close-value date-name content-type]
-      (map #(HttpHeaders/newEntity %) ["Server" "Aleph/0.6.0" "Connection" "Keep-Alive" "Close" "Date" "Content-Type"])]
+(let [[server-name connection-name date-name content-type]
+      (map #(HttpHeaders/newEntity %) ["Server" "Connection" "Date" "Content-Type"])
+
+      [server-value keep-alive-value close-value]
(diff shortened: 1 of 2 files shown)


src/aleph/http/server.clj
```diff
@@ -111,8 +111,8 @@
            (.printStackTrace e (java.io.PrintWriter. w))
            (str w))})
 
-(let [[server-name connection-name date-name]
-      (map #(HttpHeaders/newEntity %) ["Server" "Connection" "Date"])
+(let [[server-name connection-name date-name content-type]
+      (map #(HttpHeaders/newEntity %) ["Server" "Connection" "Date" "Content-Type"])
 
       [server-value keep-alive-value close-value]
       (map #(HttpHeaders/newEntity %) ["Aleph/0.6.0" "Keep-Alive" "Close"])]
```
KingMob, line 118: ...why do we have two `map` calls?

???

KingMob (approved):

src/aleph/http/server.clj
arnaudgeiser: I will merge them but I guess it was for readability reasons only.

arnaudgeiser: 0fef219efcf69a0a01b582619c994004159c95d7

arnaudgeiser: Actually, I'm going to revert my changes... I understand why it has been done that way.

The first `map` is for the __headers keys__.
The second `map` represents __headers values__.

It's nice to have a quick look at the headers we are manipulating when sending the response instead of a medley.

arnaudgeiser:
Tests failed because of fragile TCP tests.
I'm going to merge this and address the comment here to fix them : https://github.com/clj-commons/aleph/pull/661

arnaudgeiser:
Closes #236

PR #664, opened 2023-01-30
Document how to replace internal scheduler

### LINKED ISSUE #479: manifold-scheduler-pool can be a bottleneck on higher loads
Feel free to mark this as a minor issue, and I don't have a generic solution right now, just throwing around some ideas.

manifold-scheduler-pool that is used whenever `manifold.time/in` is used (and transitively for all `manifold.deferred/timeout!`) is based on Java's `ScheduledThreadPoolExecutor`. That executor uses blocking queues and is not terribly efficient when facing 100k+/sec scheduling rate and grows a sizeable tail in the queue. Aleph is quite trigger-happy about creating timeouts, especially when you use a client and set `:pool-timeout`, `:connection-timeout`, and the other similar parameters.

The immediate solution I came up with is replacing the default scheduler with Netty's HashedTimerWheel, which uses JCTools' lockless MPSC queue under the hood. No extra dependencies are needed since Netty is already there; however, this approach cannot be blindly applied to Manifold itself as people might use it without Aleph (and Netty). Perhaps, we could detect on pool initialization whether Netty is present and create a more efficient executor then? I don't know.

Anyway, here's a hack for those who might run into a similar problem:

```clj
(def ^:private hashed-timer-clock
  "A replacement for the default manifold.time clock that is used for scheduling
  timeouts. We do many timeouts, need something faster. Netty's HashedWheelTimer
  is a good replacement, based on JCTools' MPSC lockless queue."
  (let [;; 10 milliseconds is the minimal resolution - should be enough.
        timer (HashedWheelTimer.
               (thread-factory (fn [] "manifold-timeout-scheduler"))
               10 TimeUnit/MILLISECONDS 1024)
        ;; We also create a regular ScheduledExecutor to serve IClock's `every`
        ;; functionality since WheelTimer doesn't do it. Yet, those periodic
        ;; schedulings are less frequent, so we can afford using the default.
        periodic-clock (mtime/scheduled-executor->clock
                        (Executors/newSingleThreadScheduledExecutor
                         (thread-factory (fn [] "manifold-periodic-scheduler"))))]
    (reify IClock
      (in [_ interval f]
        (.newTimeout timer (reify TimerTask (run [_ _] (f)))
                     interval TimeUnit/MILLISECONDS))
      (every [_ delay period f]
        (.every ^IClock periodic-clock delay period f)))))

;; And now, hack into the root of a dynamic var. Kids, don't try this at home.
(alter-var-root #'mtime/*clock* (constantly hashed-timer-clock))
```

arnaudgeiser:
Closes #479

### DIFF
--- a/docs/advanced/replace-internal-scheduler.md
+++ b/docs/advanced/replace-internal-scheduler.md
@@ -2,7 +2,7 @@
 
 ## Default scheduler
 
-The default scheduler of Manifold (and thus Aleph) is not efficient for handling high volume scheduling requests (+100k/sec) as it's based on [ScheduledThreadPoolExecutor](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html) which uses blocking queues.
+The default scheduler of Manifold (and thus Aleph) is not efficient for handling high volume scheduling requests (+100k/sec) as it's based on [ScheduledThreadPoolExecutor](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ScheduledExecutorService.html) which uses blocking queues.
 
 For optimized performance, you can provide a custom scheduler by redefining `manifold.time/*clock*`.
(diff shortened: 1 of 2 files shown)


KingMob (approved):
LGTM

PR #670, opened 2023-02-07
Randomize UDP ports

arnaudgeiser:
## Description

Since an `UDP` socket doesn't implement the `AlephServer` protocol, the `port` function is unavailable.
Here is a quick solution to stabilize those tests.

### DIFF
--- a/test/aleph/udp_test.clj
+++ b/test/aleph/udp_test.clj
@@ -3,26 +3,40 @@
    [aleph.netty :as netty]
    [aleph.udp :as udp]
    [clj-commons.byte-streams :as bs]
-   [clojure.test :refer [deftest testing is]]
-   [manifold.stream :as s]))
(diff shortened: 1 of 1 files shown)


KingMob:
Hmm, I wonder how hard it would be to make the UDP code use `netty/start-server`, for parity. `udp/socket` takes almost the same options as `tcp/start-server`, it shouldn't be too difficult, in theory.

KingMob (approved):
Approving for now, but I think we should bring UDP into closer parity with TCP.

arnaudgeiser:
> Hmm, I wonder how hard it would be to make the UDP code use netty/start-server, for parity. udp/socket takes almost the same options as tcp/start-server, it shouldn't be too difficult, in theory.

I think so too.
Promise, I will make a PR for it. :)

PR #672, opened 2023-02-12
Fix backwards-compatibility on transport + doc

arnaudgeiser:
## Description

When randomly reviewing some parts of the code I discovered that `netty/create-client` (the backwards-compatibility function) is sending the `epoll?` key instead of the `transport` one.
This function is not used (anymore) inside Aleph but there is a low chance that some Aleph users might still use it with some workarounds... who knows.

So it can be called a regression where the worst case scenario is that the client won't even start. But I don't believe everyone will be impacted.

I also added some `dots` on the documentation for conformance and remove some white spaces.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1298,7 +1298,7 @@ initialize an DnsAddressResolverGroup instance.
                    :bootstrap-transform bootstrap-transform
                    :remote-address      remote-address
                    :local-address       local-address
-                   :epoll?              (if epoll? :epoll :nio)
+                   :transport           (if epoll? :epoll :nio)
                    :name-resolver       name-resolver}))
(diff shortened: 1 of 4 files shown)


src/aleph/netty.clj
```diff
@@ -1298,15 +1298,15 @@ initialize an DnsAddressResolverGroup instance.
...
                    :name-resolver       name-resolver}))
   ([{:keys [pipeline-builder
             ssl-context
             bootstrap-transform
             ^SocketAddress remote-address
             ^SocketAddress local-address
-            name-resolver
-            transport]}]
+            transport
+            name-resolver]}]
```
arnaudgeiser, line 1309: Just changed the order...

KingMob (approved):

Issue #454, opened 2018-12-19
Premature server response causes client to crash


kachayev:
If a server sends response while a client is still transferring request body, the exception raised by netty (see details below) is caught in client handler and causes an appropriate connection to be "stuck" (which is [another](https://github.com/ztellman/aleph/issues/453) problem). Regarding RFC2616, 

> An HTTP/1.1 (or later) client sending a message-body SHOULD monitor the network connection for an error status while it is transmitting the request. If the client sees an error status, it SHOULD immediately cease transmitting the body. If the body is being sent using a "chunked" encoding (section 3.6), a zero length chunk and empty trailer MAY be used to prematurely mark the end of the message. If the body was preceded by a Content-Length header, the client MUST close the connection.

Server:

```clojure
(defn no-ops-handler [_]
  {:status 200
   :body "OK"})

(http/start-server no-ops-handler {:port 2018
                                   :request-buffer-size 16})
```

Client:

```clojure
(http/post "http://localhost:2018" {:body (map str (range 1e6))})
```

Client logs an exception:

```clojure
[aleph-netty-client-event-pool-7] WARN aleph.http.client - error in HTTP client
io.netty.handler.codec.EncoderException: java.lang.IllegalStateException: unexpected message type: DefaultHttpRequest, state: 2
<< … >>
        at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:107)
        at io.netty.channel.CombinedChannelDuplexHandler.write(CombinedChannelDuplexHandler.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:738)
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:730)
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:816)
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:723)
        at io.netty.handler.stream.ChunkedWriteHandler.doFlush(ChunkedWriteHandler.java:305)
        at io.netty.handler.stream.ChunkedWriteHandler.flush(ChunkedWriteHandler.java:135)
        at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:776)
[...]

Issue #453, opened 2018-12-18
Suppressing exceptions caught in the pipeline might lead to undefined behavior


kachayev:
Both [server](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/server.clj#L314-L316) and [client](https://github.com/ztellman/aleph/blob/be1ee04c130660ce69ee7c6603132b36be79ec1f/src/aleph/http/client.clj#L164-L173) handlers simply log all caught exceptions. In case something goes wrong in the pipeline, processing might be stuck. E.g. when a server responds unexpectedly, Netty throws `io.netty.handler.codec.EncoderException` making the connection effectively useless. I'm not sure what is the best approach handling exceptions for the server, but at least for the client we probably need to clean up (close connection?) or propagate exception as the request result.

PR #673, opened 2023-02-20
Stabilize tests

arnaudgeiser:
## Description

Ensure non keep-alived connections are used in `multipart` and `ring` integration tests.

### DIFF
--- a/test/aleph/http/multipart_test.clj
+++ b/test/aleph/http/multipart_test.clj
@@ -138,10 +138,8 @@
 
 (def port1 26023)
 (def port2 26024)
-(def port3 26025)
 (def url1 (str "http://localhost:" port1))
 (def url2 (str "http://localhost:" port2))
-(def url3 (str "http://localhost:" port3))
 
 (def parts [{:part-name "#0-string"
              :content "CONTENT1"}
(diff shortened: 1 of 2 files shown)


KingMob:
How does this stabilize tests? If we rely on closing the conn instead of marking the final chunk, we have a bug, yeah? And if we can't use a persistent connection reliably, we have a different bug 😄 

I don't mind specific tests to check for connection closing and  HTTP/1.0 compatibility with `Connection: keep-alive`/`Connection: close`, but HTTP/1.0 is almost non-existent these days, and 1.1 assumes keep-alive is the default, so disabling keep-alive seems like it should be rare.

arnaudgeiser:
> How does this stabilize tests? If we rely on closing the conn instead of marking the final chunk, we have a bug, yeah? And if we can't use a persistent connection reliably, we have a different bug smile

Yes, we definitely have some __bugs__ with the client. (c.f the discussion on Slack with Ferdinand, where a connection is currently in use and might be closed before/in between)

The thing is... in integration tests, we are starting/restarting the server multiple times while the client is a shared instance. When a server is closed (too fast? - `shutdown-timeout 0`), I observed it might have some impacts on the connection with the client, hence with I propose to not reuse the connection in integration tests.

I'm currently working on this issue [1] which is related to those [2].
While it might not be the culprit, I have the hunch it might be kind of related, at least, I'm able to reproduce similar behavior (closed connection, connection stuck). But I need more time to come up with a proposition/PR here...

[1] : https://github.com/clj-commons/aleph/issues/454
[2] : https://github.com/clj-commons/aleph/issues/453

KingMob:
If you think this change will help those other issues, maybe incorporate it into that future PR instead?

arnaudgeiser:
I will close it. Let's see if we are able to improve the client behavior then (no lead for now).

PR #675, opened 2023-02-21
Fix version number in Readme

ertugrulcetin:
Clojars and Readme versions numbers are out of sync


KingMob:
Oops. Thanks for catching that.

PR #677, opened 2023-02-25
Fix logging dep conflict

### LINKED ISSUE #676: "Possibly confusing dependencies found" regression in 0.6.1
When running `lein deps :tree` in a project which includes aleph 4.6.1 as dependency, I get:

```
Possibly confusing dependencies found:
[aleph "0.6.1"] -> [org.clojure/tools.logging "1.1.0" :exclusions [org.clojure/clojure]]
 overrides
[aleph "0.6.1"] -> [manifold "0.3.0"] -> [org.clojure/tools.logging "1.2.4" :exclusions [org.clojure/clojure]]
```

So aleph includes a transitive dependency which depends on org.clojure/tools.logging 1.2.4 while having a direct dependency on org.clojure/tools.logging 1.1.0. I suggest updating the direct dependency to match the 1.2.4 version. This doesn't happen in aleph 4.6.0.

KingMob:
Also includes some other minor changes:

- core.async dep updated in examples
- websocket examples takes are printed, to be less confusing at REPL
- .eastwood file now gitignored

Fixes #676

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
           command: |
             sudo bash < <(curl -s https://raw.githubusercontent.com/babashka/babashka/3d916df4a0c1e00df94100860b8eb5577e59c56a/install)
 
-      - run: lein deps
+      - run: lein with-profile pedantic deps
 
       - save_cache:
           paths:
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ target/**
 .java-version
 **/.clj-kondo/.cache
 **/.lsp
+/.eastwood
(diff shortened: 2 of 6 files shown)


arnaudgeiser (changes requested):
I would go even further regarding this issue and enable pedantic by adding the following profile :

```
:pedantic {:pedantic? :abort}
```

and then, instead of just invoking `lein deps`, let's use `lein with-profile pedantic deps`.

In all cases, I would also add an exclusion for `org.clojure/tools.logging` when defining the manifold dependency to be the master of the dependency.

arnaudgeiser (approved):

PR #678, opened 2023-02-26
project: housekeeping

arnaudgeiser:
This PR :

- Removes plugins and configuration that is no longer in use on `project.clj`
- Bump Netty to its latest version
- Bump io-uring to its latest version


project.clj
```diff
@@ -33,23 +33,12 @@
                                    [org.bouncycastle/bcprov-jdk18on "1.72"]
                                    [org.bouncycastle/bcpkix-jdk18on "1.72"]]
                     :jvm-opts ["-Dorg.slf4j.simpleLogger.defaultLogLevel=debug"]}
-             :lein-to-deps {:source-paths ["deps"]}
```
KingMob, line 36: There's now a lein-to-deps lein plugin from Borkdude, we should consider using that over our homegrown scripts. In particular, because the plugin can be set up to trigger automatically, and I keep forgetting to run our scripts 😄

KingMob (approved):

project.clj
arnaudgeiser: To be honest, I'm a bit lost here...
Is it not what we are currently using already? [1]
This bit of configuration (the one I proposed to remove) was part of the previous `lein-to-deps` plugin we were using.

On my side, I just expect to not have to care about `deps` when developing Aleph. So a precommit hook would be totally fine for me, but I don't know what you have in mind. You can impose me any workflow, that's fine with me.

[1] : (https://github.com/clj-commons/aleph/blob/master/deps/lein-to-deps#L8)

Issue #162, opened 2015-04-14
aleph.http.request could do a better job erroring if when mandatory arguments are missing


lvh:
Turns out I forgot to pass `:method`; got this:

```
2015-Apr-14 10:37:55 -0400 zygalski.local DEBUG [icecap.handlers.http] - (request args) << … >>
Apr 14, 2015 10:37:55 AM manifold.stream invoke
SEVERE: error in consume
java.lang.NullPointerException
    at clojure.core$name.invoke(core.clj:1522)
    at aleph.http.core$ring_request__GT_netty_request.invoke(core.clj:157)
    at aleph.http.client$http_connection$fn__18161$fn__18162.invoke(client.clj:278)
    at manifold.stream.Callback.put(stream.clj:431)
    at manifold.stream.graph$async_send.invoke(graph.clj:50)
    at manifold.stream.graph$async_connect$this__13640.invoke(graph.clj:172)
    at manifold.stream.graph$async_connect$this__13640$fn__13641$fn__13642.invoke(graph.clj:153)
    at clojure.core$trampoline.invoke(core.clj:6033)
    at manifold.stream.graph$async_connect$this__13640$fn__13641.invoke(graph.clj:153)
    at manifold.deferred.Listener$f__12755__auto____12960.invoke(deferred.clj:158)
    at manifold.deferred.Listener.onSuccess(deferred.clj:158)
    at manifold.deferred.Deferred$fn__13037.invoke(deferred.clj:324)
    at manifold.deferred.Deferred.success(deferred.clj:324)
    at manifold.deferred$success_BANG_.invoke(deferred.clj:184)
    at manifold.stream.default.Stream$fn__13733.invoke(default.clj:96)
    at manifold.stream.default.Stream.put(default.clj:95)
    at manifold.stream.default.Stream.put(default.clj:119)
    at aleph.http.client$http_connection$fn__18161$fn__18173$fn__18174.invoke(client.clj:296)
    at aleph.http.client$http_connection$fn__18161$fn__18173.invoke(client.clj:295)
    at aleph.http.client_middleware$wrap_request_timing$fn__18086.invoke(client_middleware.clj:425)
    at aleph.http.client_middleware$wrap_query_params$fn__18031.invoke(client_middleware.clj:305)
    at aleph.http.client_middleware$wrap_basic_auth$fn__18038.invoke(client_middleware.clj:325)
[...]

PR #679, opened 2023-02-27
Add wrap-validation middleware based on Malli

KingMob:
@arnaudgeiser What do you think of this?

Also moves the schemas to a common aleph.http.schema ns, so they can be shared.

Not sure we need each individual var schema just yet, so I've hidden the ns from cljdoc for now.


KingMob:
We should probably tighten the schemas soon. I've had bad experiences in the past using spec, when people let specs be loose for too long, they became concerned that locking them down would break something (E.g., one team only accepts ports as numbers, but another team also accepts them as strings which it parses...)

KingMob:
...and I'm already reminded of why `:pedantic :abort` was a huge pain. It lead to people excluding things at random because they just want their build to succeed.

src/aleph/http/schema.clj
```diff
@@ -0,0 +1,46 @@
+(ns ^:no-doc aleph.http.schema
+  (:require [malli.core :as m]))
+
+(def server-port [:maybe :int])
+(def server-name [:maybe :string])
+(def remote-addr [:maybe :string])
+(def uri [:maybe :string])
+(def query-string [:maybe :string])
+(def scheme [:enum :http :https])
+(def request-method :keyword)
```
DerGuteMoritz, line 10: Note that Netty actually [defines an enum for this](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java#L88-L101), so allowing arbitrary keywords will lead to an error further down the line. But of course, that property is private so we wold have to redundantly declare it here :grimacing: You call!

src/aleph/http/schema.clj
```diff
@@ -0,0 +1,46 @@
...
+(def content-type [:maybe [:or :string :keyword]])
+(def content-length [:maybe :int])
+(def character-encoding [:maybe :string])
```
DerGuteMoritz, line 13: The `:maybe`s here are probably technically correct but I usually prefer to have only one way to express absence of a key (i.e. via `:optional true`). Otherwise, accidental passing of `nil` (e.g. due to spelling errors) easily slips through.

src/aleph/http/client_middleware.clj
```diff
@@ -4,10 +4,12 @@
   (:require
     ;; leave this dependency to make sure that HeaderMap is already compiled
     [aleph.http.core :as http]
+    [aleph.http.schema :as schema]
     [clj-commons.byte-streams :as bs]
     [clojure.edn :as edn]
     [clojure.string :as str]
     [clojure.walk :refer [prewalk]]
+    [malli.core :as m]
```
DerGuteMoritz, line 12: Unused

src/aleph/http/schema.clj
```diff
@@ -0,0 +1,46 @@
...
+                   [:server-name {:optional true} server-name]
+                   [:remote-addr {:optional true} remote-addr]
+                   [:uri {:optional true} uri]
+                   [:query-string {:optional true} query-string]
+                   [:scheme {:optional true} scheme]
+                   [:content-type {:optional true} content-type]
+                   [:content-length {:optional true} content-length]
+                   [:character-encoding {:optional true} character-encoding]])
+
+(def validate-request (m/validator ring-request))
```
DerGuteMoritz, line 27:
```suggestion
(def valid-request? (m/validator ring-request))
```

src/aleph/http/schema.clj
```diff
@@ -0,0 +1,46 @@
...
+                   [:request-method request-method]
+                   [:server-port {:optional true} server-port]
+                   [:server-name {:optional true} server-name]
+                   [:remote-addr {:optional true} remote-addr]
+                   [:uri {:optional true} uri]
+                   [:query-string {:optional true} query-string]
+                   [:scheme {:optional true} scheme]
+                   [:content-type {:optional true} content-type]
+                   [:content-length {:optional true} content-length]
+                   [:character-encoding {:optional true} character-encoding]])
```
DerGuteMoritz, line 25: What about `:body`?

DerGuteMoritz (changes requested):

src/aleph/http/schema.clj
arnaudgeiser: I'm a bit concerned about breaking call sites if suddenly we forbid `nil` keys.

src/aleph/http/schema.clj
arnaudgeiser: While an enum is defined, nothing prevents you from creating arbitrary HTTP method. [1]
I would say this enum is just to prevent allocations (maybe?).

[1] : https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java#L117-L118

arnaudgeiser:
> @arnaudgeiser What do you think of this?

I'm totally fine with using Malli instead of spec, I don't have strong opinions about them.
However, I'm still wondering if the overhead of validating spec is worth. Actually, only the presence of the  `request-method` is strongly required and currently leads to intelligible error.

> We should probably tighten the schemas soon. I've had bad experiences in the past using spec, when people let specs be loose for too long, they became concerned that locking them down would break something (E.g., one team only accepts ports as numbers, but another team also accepts them as strings which it parses...)

Isn't it too late already?

> ...and I'm already reminded of why :pedantic :abort was a huge pain. It lead to people excluding things at random because they just want their build to succeed.

You're a bit harsh here. :smile: 
We don't add dependencies everyday and I prefer having errors at build time instead of at runtime with libraries incompatibilities. It might be annoying, but I would say it's manageable on a project like Aleph.

In the current example, I'm wondering if we should also pin the `org.clojure/clojure` like metosin does for malli.

src/aleph/http/schema.clj
DerGuteMoritz: Ah thanks, should have read `valueOf` more closely! Works for me then :+1:

src/aleph/http/schema.clj
DerGuteMoritz: Hm yeah, you're probably right, better to be conservative here, not worth releasing a breaking change for this.

KingMob: The schemas were 1-to-1 translations of the spec versions.

KingMob:
Let's take a step back and decide what problem we actually want to solve. 

The original issue (#162) was lvh complaining about the error message he got when he forgot the method key.

What else (if anything) do we want out of validation? Just to return some better error messages for bad requests? To lock down the request format users submit to the client? To capture our current loose schemas, but then transform the input into something more stringent for Aleph's use? Get kondo support for our own use?

Personally, I:
- am slightly in favor of better error messages for users
- think locking it down will cause breakage
- think transforming is nice to have, but not important
- think kondo support would help a lot, given how kondo-unfriendly aleph and manifold are

arnaudgeiser:
> Just to return some better error messages for bad requests?

For me, that should be the focus.

KingMob:
@DerGuteMoritz Any thoughts on the goal of validation?

DerGuteMoritz:
@KingMob Thanks for providing context of the original issue, I was missing that :pray: 

> Personally, I:
> 
> * am slightly in favor of better error messages for users

Looking at the original issue, that definitely should be the priority. Question is whether this alone justifies pulling in malli, especially if it's only for this particular API. We'd end up with a mixed bag of schema-based and hand-rolled validation (and errors). I feel like adopting a schema-based approach should be done in a more holistic fashion. For https://github.com/clj-commons/aleph/issues/162 I now lean towards a minimally invasive patch with hand-rolled validation.

> * think locking it down will cause breakage

Agreed.

> * think transforming is nice to have, but not important

Agreed.

> * think kondo support would help a lot, given how kondo-unfriendly aleph and manifold are

Indeed. As mentioned above, I think doing this via malli is worthwhile since it will also give us opt-in runtime validation (which could then supercede the hand-rolled one) and generative testing support at the same time. Ideally then for all of Aleph's public API.

KingMob:
I'd like to pull in Malli as a vote for the future, then, and because kondo support would help me, personally, a lot. We might not need it for any given small PR, but if it's already a dep, we can start to build and rely on it.

src/aleph/http/schema.clj
KingMob: What would you suggest? It's optional, for starters, but the real issue is it has a ton of representations once middleware gets involved: strings, byte arrays, Clojure maps read from JSON, Manifold streams, InputStreams, etc.

Should we enumerate them? It might be useful when testing middleware to check the contract of what they return, but that will take a bit. For now, we could use `:any` to show we haven't forgotten about it.

arnaudgeiser (approved):

src/aleph/http/schema.clj
KingMob: Added `:body`.

PR #681, opened 2023-04-25
Client HTTP/2 support

KingMob:
WIP. Architecture very much not set in stone, this is somewhat exploratory.

Note that under the hood, Netty does things differently for HTTP/2.


src/aleph/http/client.clj
```diff
@@ -498,401 +957,157 @@
...
+                                        :pipeline pipeline
+                                        :pipeline-transform pipeline-transform
+                                        :raw-stream? raw-stream?
+                                        :remote-address remote-address
+                                        :response-buffer-size response-buffer-size
+                                        :ssl-context ssl-context
+                                        :ssl? ssl?)]
+                  (log/debug (str "HTTP protocol: " protocol))
+
+                  (let [version-specific-req-fn
```
arnaudgeiser, line 1047: Call we call this something like `request-handler`, `req-handler` or `req-fn` instead?

src/aleph/http/client.clj
```diff
@@ -386,95 +392,314 @@
...
+            (if-not (d/realized? response-d)
+              (d/error! response-d ex)
+              (do
+                (d/error! complete ex)
+                (-> (s/put! body-stream ex)
+                    (d/on-realized close-body-handler close-body-handler))))))
+
+        handle-shutdown-frame
+        (fn handle-shutdown-frame [evt]
+          (when (or (instance? Http2ResetFrame evt)
```
arnaudgeiser, line 525: Sad those two classses are not sharing a common interface.

src/aleph/http/client.clj
```diff
@@ -386,95 +392,314 @@
...
+      (throw (IllegalArgumentException. "Proxying HTTP/2 messages not supported yet")))
+
+    (add-non-http-handlers p logger pipeline-transform)
+
+    (log/trace "added all stream-chan handlers")
+    (log/debug "Stream chan:" ch)
+
+    ch))
+
+(defn- h2-stream-chan-initializer
```
arnaudgeiser, line 614: This is not used.

src/aleph/http/client.clj
```diff
@@ -487,7 +712,241 @@
...
+                           :body
+                           (bs/to-input-stream body
+                                               {:buffer-size response-buffer-size}))))))))
+
+(defn- http1-req-handler
+  "Returns a fn that takes a Ring request and returns a deferred containing a
+   Ring response.
+
+   Otherwise, puts/takes to/from the requests/responses streams."
+  [{:keys [ch keep-alive? raw-stream? requests response-buffer-size responses]}]
```
arnaudgeiser, line 759: I would remove all the parameters that are no in use.

src/aleph/http/client.clj
```diff
@@ -487,7 +712,241 @@
...
+                (s/put! requests req)
+                (s/take! responses ::closed))]
+      (log/debug "http1-req-handler - response: " resp)
+      resp)))
+
+
+(defn- make-http1-req-preprocessor
+  "Returns a fn that handles a Ring req map using the HTTP/1 objects.
+
+   Used for HTTP/1, and for HTTP/2 with multipart requests (Netty HTTP/2
```
arnaudgeiser, line 772: Since we are not going to support HTTP/2 multipart requests, at least initially, I would remove the modifications that have been done to support HTTP2.
(No strong opinion)

src/aleph/http/client.clj
```diff
@@ -386,95 +392,314 @@
...
+                  "pause-handler"
+                  ^ChannelHandler (netty/pause-handler))))))
 
 (defn close-connection [f]
   (f
     {:method :get
-     :url "http://example.com"
-     ::close true}))
+     :url    "http://example.com"
+     :aleph/close true}))
```
arnaudgeiser, line 702: Since this got modified, maybe we can remove the `::close` line 1090.

src/aleph/http/client.clj
```diff
@@ -487,7 +712,241 @@
...
+  (if ssl?
+    (if ssl-context
+      (let [^SslContext ssl-ctx (netty/coerce-ssl-client-context ssl-context)]
+        ;; check that ALPN is set up if HTTP/2 is requested (https://www.rfc-editor.org/rfc/rfc9113.html#section-3.3)_
+        (if (and (-> ssl-ctx
+                     (.applicationProtocolNegotiator)
+                     (.protocols)
+                     (.contains ApplicationProtocolNames/HTTP_2)
+                     not)
+                 (some #(= :http2 %) desired-protocols))
```
arnaudgeiser, line 896: Shall we make the `desired-protocols` a set and perform `(desired-protocols :http2)` instead?

src/aleph/http/client.clj
```diff
@@ -498,401 +957,157 @@
...
+                ;; We know the SSL handshake must be complete because create-client wraps the
+                ;; future with maybe-ssl-handshake-future, so we can get the negotiated
+                ;; protocol, falling back to HTTP/1.1 by default.
+                (let [pipeline (.pipeline ch)
+                      protocol (if ssl?
+                                 (or (-> pipeline
+                                         ^SslHandler (.get ^Class SslHandler)
+                                         (.applicationProtocol))
+                                     ApplicationProtocolNames/HTTP_1_1)
+                                 ApplicationProtocolNames/HTTP_1_1)
```
arnaudgeiser, line 1030: Can you confirm we are using HTTP1.1 because of this? https://httpwg.org/specs/rfc7540.html#TLSUsage

src/aleph/http/client.clj
```diff
@@ -498,401 +957,157 @@
...
+                            (ex-info "Channel is inactive/closed."
+                                     {:req     req
+                                      :ch      ch
+                                      :open?   (.isOpen ch)
+                                      :active? (.isActive ch)}))
+
+                          ;; If :aleph/close is set in the req, closes the channel and
+                          ;; returns a deferred containing the result.
+                          (if (or (contains? req :aleph/close)
+                                  (contains? req ::close))
```
arnaudgeiser, line 1090: If think that `::close` can be removed now, no?

test/aleph/http_test.clj
```diff
@@ -790,6 +790,10 @@
 (defn tcp-handler [response]
   (fn [s ch]
     (s/take! s)
+    ;; too fast, have to slow it down - since this handler responds as soon as
+    ;; it's called, a returned error can shut down the client channel before
+    ;; it's even finished sending, leading to UNexpected errors
+    (Thread/sleep 5)
```
arnaudgeiser, line 796: "Nice", I didn't know how to solve this issue.

arnaudgeiser (commented):
I spent most of me time on both the `aleph.http.client` and the `aleph.http.http2` namespace to be sure I have a good understanding.
Overall the changes are looking good, I also performed some manual tests to a HTTP2 Nginx running locally and it's working as expected.
However, I truly think we are lacking some tests, maybe those can be introduced once the server will be HTTP2.0 compatible?

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+      OpenOption
+      Path
+      StandardOpenOption)))
+
+(set! *warn-on-reflection* true)
+
+(def ^:private byte-array-class (Class/forName "[B"))
+
+(def ^:const max-allowed-frame-size 16777215) ; 2^24 - 1, see RFC 9113, SETTINGS_MAX_FRAME_SIZE
+(def ^:dynamic *default-chunk-size* 16384) ; same as default SETTINGS_MAX_FRAME_SIZE
```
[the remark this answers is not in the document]
KingMob: Good point. I need to double-check, but I think I may have done that before exposing it as a param.

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+
+;; See https://httpwg.org/specs/rfc9113.html#ConnectionSpecific
+(def invalid-headers #{"connection" "proxy-connection" "keep-alive" "upgrade"})
+
+(let [illegal-arg (fn [^String emsg]
```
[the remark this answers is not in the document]
KingMob: Just following in Zach's footsteps, I guess. The benefit of the `let` is making clear the fn is scoped to `defn` it wraps, but I'm not sure we care that much. I may even use a similar fn elsewhere, in which case it *really* should be available throughout the ns.

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+          (sequential? header-value)
+          (.add h2-headers ^CharSequence header-name ^Iterable header-value)
+
+          :else
+          (.add h2-headers ^CharSequence header-name ^Object header-value))))))
+
+(defn- ring-map->netty-http2-headers
+  "Builds a Netty Http2Headers object from a Ring map."
+  ^DefaultHttp2Headers
+  [req]
```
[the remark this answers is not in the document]
KingMob: Zach felt that these were hot-enough paths in the HTTP1 code to warrant using `get` directly, and use a ConcurrentHashMap to cache header case transformations, so I did the same for HTTP2. 

I haven't profiled (which I should), but it makes sense. This function will be called for every single request, so especially on the server, it will be hot, and destructuring is (relatively) slow.

If anything, I'm incline to do it more once we profile.

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+(defn- try-set-content-length!
+  "Attempts to set the `content-length` header if not already set.
+
+   Will skip if it's a response and the status code is 1xx or 204.
+   Negative length values are ignored."
+  ^Http2Headers
+  [^Http2Headers headers ^long length]
+  (when-not (.get headers "content-length")
+    (if (some? (.status headers))
+      ;; TODO: consider switching to netty's HttpResponseStatus and HttpStatusClass for clarity
```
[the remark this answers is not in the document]
KingMob: Fair. However, we already use them a bit in the server code, so it might be nice to align our usage.

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+      (let [code (-> headers (.status) (Long/parseLong))]
+        (when-not (or (<= 100 code 199)
```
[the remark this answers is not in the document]
KingMob: Good catch, but sadly, no. The primitive-math fns don't support arities higher than 2. I thought about using two `<=` calls, but thought it would look too weird. OTOH, if this is about performance, I _should_ do it...

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+;; branches based on the class (channel vs context), but that's not ideal. It's slower, and
+;; writing to the channel vs the context means different things, anyway, they're not
+;; usually interchangeable.
+(defn send-request
+  [^Http2StreamChannel ch req response]
+  (log/trace "http2 send-request fired")
+
+  (when (multipart/is-multipart? req)
+    ;; shouldn't reach at the moment, but just in case
+    (throw (ex-info "HTTP/2 code path multipart not supported"
```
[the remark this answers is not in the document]
KingMob: Maybe, but it'll be caught right above in `http2-req-handler`. Honestly, I was planning to delete this code, since it's literally checked right before being called.

src/aleph/http/http2.clj
```diff
@@ -0,0 +1,503 @@
...
+                 ;; RFC #7231 4.3.8. TRACE
+                 ;; A client MUST NOT send a message body...
+                 (do
+                   (when (some? (:body req))
+                     (log/warn "Non-nil TRACE request body was removed"))
+                   nil)
+                 (:body req))
+          headers (ring-map->netty-http2-headers req)
+          chunk-size (or (:chunk-size req) *default-chunk-size*)
+          file-chunk-size (p/int (or (:chunk-size req) file/*default-file-chunk-size*))]
```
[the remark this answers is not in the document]
KingMob: Sure, if you want to set the global default. "Mutable" globals that are set once at startup time and only read from then on, don't really pose a problem. 

Otherwise you have to set a param and attach it to every map everywhere which wastes memory and clutters stuff up.

src/aleph/http/client.clj
KingMob: The problem those names are either taken, or awfully similar to existing names. 😞 

I wanted to add names to lambdas for clarification and improved stack traces, but especially on the HTTP1 side, I started running out of useful names...because what code was in which function was never designed in a principled way that lent itself to a good name.

I guess `req-handler` is technically available...but it sounds a lot like `http-req-fn` right below. What about `http-req-handler` or `http-version-req-handler`?

src/aleph/http/client.clj
KingMob: Yeah...but they're meant for different use cases. From the perspective of a stream handler, they both mean "shut down this stream", but GOAWAY is for the whole connection, and RST_STREAM is just for this stream.

src/aleph/http/client.clj
KingMob: True...it was being used for `server-initiated-stream-chan-initializer`, which I replaced with a placeholder that just logs an error if the server initiates a stream. I was using for the client, but stopped because I needed different code to snc the manifold and netty side of things.

I thought it might still be useful for the server-side, so I haven't deleted it yet.

src/aleph/http/client.clj
KingMob: Yeah, good point. I was shuffling a lot of stuff around, and waiting for the dust to settle, and forgot to clean that up

src/aleph/http/client.clj
KingMob: I think the only change is to the addition of `out-ch`.

I guess removing it is fine. I'm tempted to leave it in, since there's a decent chance the the path to HTTP/2 multipart support involves inserting `Http2StreamFrameToHttpObjectCodec` into the pipeline, at which point, the old HTTP1 code could theoretically work, but it's unclear if it's a minor thing, or will still require major adjustments.

If it still requires major adjustments to the old HTTP1 code, we might as well write a new, simpler handler that does just what we need with the HTTP1 objects.

src/aleph/http/client.clj
KingMob: Good eye, but that might break user code on the off-chance they did used the old key manually. Pretty unlikely, but I figured just in case...

src/aleph/http/client.clj
KingMob: We could convert it to a set in this _particular_ spot, but in general, the desired protocols are ordered, because they indicate which protocols you most prefer. E.g., `[:http1 :http2]` means you prefer HTTP/1 over HTTP/2, and the server will only fall back to HTTP/2 if it doesn't support HTTP/1.

src/aleph/http/client.clj
KingMob: I'm not sure I follow. If you're referring to the final clause of the `or`, we set it to HTTP/1.1 because that's Aleph's default if we're not using SSL/TLS.

The other use of ApplicationProtocolNames/HTTP_1_1, on line 1029, is a fallback in case we ARE using SSL, but the server is NOT using ALPN for some reason. Since HTTP/2 requires ALPN, we shouldn't let that be the default. (And any server old enough to not use ALPN is probably defaulting to HTTP/1 anyway.)

(BTW, RFC 7540 is obsoleted by RFC 9113.)

src/aleph/http/client.clj
KingMob: I thought I replied to this already...I left in `::close` on the off chance some user was manually using it.

test/aleph/http_test.clj
KingMob: It's such a hack, but I was getting the weirdest errors. 😛

KingMob:
> However, I truly think we are lacking some tests, maybe those can be introduced once the server will be HTTP2.0 compatible?

Yeah, my goal is to get most of the existing tests working for both HTTP/1 and /2 once the server side is up and running. There should probably be some HTTP2-specific tests, as well, especially around TLS and connection set up.

src/aleph/http/http2.clj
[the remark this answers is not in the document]
KingMob: Removed.

src/aleph/http/client.clj
KingMob: Done

src/aleph/http/client.clj
KingMob: Removed

src/aleph/http/http2.clj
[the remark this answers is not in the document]
KingMob: OK, converted to use primitive-math, too

src/aleph/http/http2.clj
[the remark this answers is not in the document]
KingMob: Added use of ConcurrentHashMap to HTTP2 headers. Profiling suggests retrieving already-lower-cased headers is 1/3 faster (~750 ns vs 1200 ns) than always lower-casing.

src/aleph/http/client.clj
KingMob: How's `http-req-handler`?

src/aleph/http/client.clj
KingMob: One annoying caveat: there's no guarantee a server will honor your preference order. 

I was debugging something today, and got confused for a hot minute because the postman-echo.com service I've been using _always_ chooses http2 if available.

arnaudgeiser: Understood!

src/aleph/http/client.clj
arnaudgeiser: Thanks for the comment :
> ; Not using ALPN, HTTP/2 isn't allowed

arnaudgeiser (approved):
Thanks for the hard work!

PR #685, opened 2023-07-05
Add a note about breaking changes in #679

danielcompton:
The changes in release 0.6.2 broke some production code which was using `:request-method "GET"` instead of `:request-method :get`. I thought it might be good to warn others about this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,10 +14,13 @@ Contributions by Matthew Davidson and Stefan van den Oord.
 
 * Fix backwards-compatibility for transport options
 * Bump Netty to 4.1.89.Final, and io_uring to 0.0.18.Final
-* Add `wrap-validation` middleware to validate Ring maps 
 * Bump deps and example deps
 * Upgrade CircleCI instance size
 * Switch to pedantic deps for CircleCI
+
+### Breaking changes
+
+* Add `wrap-validation` middleware to validate Ring maps [#679](https://github.com/clj-commons/aleph/pull/679). This adds a stricter interpretation of the ring spec, which may fail on previously valid input, for example `:request-method "GET"`.
 
 Contributions by Arnaud Geiser, Ertuğrul Çetin, Jeroen van Dijk, David Ongaro, 
 Matthew Davidson, and Moritz Heidkamp.


KingMob:
@danielcompton Thanks for pointing this out. At first, I was surprised that string methods even worked, but I took a look at the code, and the transformation from keyword to Netty input is such that strings will usually work, too. (Though string TRACE methods will break some code.) FWIW, the Ring SPEC is clear, and it's been keywords-only for its entire life. 

Regardless, I guess the question is, should we:

1. Add a breaking change note, like this PR?
2. Remove `wrap-validation` from the default middleware, and silently continue to accept strings?
3. Loosen the schema to accept strings, too?

I'd rather not break any other wild code, so I prefer 2 or 3. I think 2 will leave the validation effectively unused by 99.99% of users, which defeats its purpose. I still hope to tighten the schema for areas like the query-string and the uri, so I vote for 3. This is not a strongly-held opinion, though. Thoughts?


----------

From #679:

> > We should probably tighten the schemas soon. I've had bad experiences in the past using spec, when people let specs be loose for too long, they became concerned that locking them down would break something (E.g., one team only accepts ports as numbers, but another team also accepts them as strings which it parses...)
>
> Isn't it too late already?

Well, _that_ comment exchange proved prophetic 😂

danielcompton:
> I'd rather not break any other wild code, so I prefer 2 or 3. I think 2 will leave the validation effectively unused by 99.99% of users, which defeats its purpose. I still hope to tighten the schema for areas like the query-string and the uri, so I vote for 3. This is not a strongly-held opinion, though. Thoughts?

Yeah, 3 seems like a safe middle ground without breaking existing code when people upgrade Aleph versions.

A possible extension to 3 could be to log a warning when callers pass a string instead of a keyword. That way, you could still tighten the spec one day if you ever decided to. Although I don't really know what you'd be gaining at that point, so maybe allowing string + keyword is ok.

arnaudgeiser:
> Loosen the schema to accept strings, too?

I would definitely loosen the schema. And as a follow-up, confirm we didn't break any other parameters along the way.

danielcompton:
Sorry, I dropped this. Happy for you to take it if you'd like.

KingMob:
@danielcompton We can do it, but I have plenty on my plate at the moment, and you're welcome to if you like. Just let us know.

danielcompton:
@KingMob I've updated the PR to expand the validation. LMK what you think.

arnaudgeiser (approved):
It looks good to me, thank you Daniel!

test/aleph/http/client_middleware_test.clj
```diff
@@ -183,6 +183,12 @@
   (doseq [req (mg/sample schema/ring-request)]
     (is (middleware/wrap-validation req)))
 
+  (testing "Request methods can be strings"
```
arnaudgeiser, line 186: modulo the `testing` import that is missing. :)

CHANGES.md
```diff
@@ -34,11 +38,14 @@ Contributions by Matthew Davidson and Stefan van den Oord.
 
 * Fix backwards-compatibility for transport options
 * Bump Netty to 4.1.89.Final, and io_uring to 0.0.18.Final
-* Add `wrap-validation` middleware to validate Ring maps
```
KingMob, line 37: Why is this line removed? A changelog should be conceptually append-only, no? Was it incorrect?

CHANGES.md
```diff
@@ -34,11 +38,14 @@ Contributions by Matthew Davidson and Stefan van den Oord.
...
 * Fix backwards-compatibility for transport options
 * Bump Netty to 4.1.89.Final, and io_uring to 0.0.18.Final
-* Add `wrap-validation` middleware to validate Ring maps 
 * Bump deps and example deps
 * Upgrade CircleCI instance size
 * Switch to pedantic deps for CircleCI
 
+### Breaking changes
+
+* Add `wrap-validation` middleware to validate Ring maps [#679](https://github.com/clj-commons/aleph/pull/679). This adds a stricter interpretation of the ring spec, which may fail on previously valid input. For example, strings (e.g. `"GET"`) and keywords (e.g. `:get`) were both accepted values for `:request-method`, but now only keywords are accepted. This will be fixed in the release after 0.7.1.
```
KingMob, line 47: Let's remove the bit about "a stricter interpretation of the Ring spec". The ring spec never allowed strings, we just accepted them by accident.

KingMob (approved):
Just got back from vacation. Looks cool to me. (Tho `testing` needs to be required)

My only note is that the Ring spec never accepted strings, and the changelog shouldn't say otherwise. It's just an Aleph bug that users might have come to rely on.

DerGuteMoritz:
LGTM now, thanks! :pray:

PR #689, opened 2023-09-18
Enable TLS endpoint identification by default (breaking)

DerGuteMoritz:
A few review notes:

1. See commit message for overall intent
2. I would have preferred to make the option part of `:ssl-context` instead having it sit next to it since conceptually, it belongs there. The reason for not doing this is that we also support passing in an `SslContext` object. Since Netty doesn't offer an API for setting an endpoint identification algorithm on the context, this would make it impossible to reach for users doing so, forcing them to convert their `SslContext` into an options map (which for some reason or another they might not be able to do). Feel free to disagree on this tradeoff, of course.
3. In the tests, I had to give up on `netty/self-signed-ssl-context` because I needed to be able to reach the certificate to properly configure the client's trust store (which in turn allowed me to remove the `:insecure? true`). Unfortunately, `SslContext` has no API for that. I guess we could still keep `netty/self-signed-ssl-context` around since it's exposed as public API. However, perhaps we should also offer an arity for passing in a hostname now?
4. Also in the tests you'll notice that I changed the default connection options to `:http-versions [:http1]`. This is because I wanted to override (only) the `:trust-store` option in `:ssl-context` which means I don't get the "good default" `:application-protocol-config` provided in `aleph.http.client/client-ssl-context` anymore. Perhaps we could lift that default a bit higher so that it also applies when passing in an `:ssl-context` map without `:application-protocol-config`?

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1384,11 +1384,17 @@ initialize an DnsAddressResolverGroup instance.
    (.newHandler ssl-ctx
                 (.alloc ch)))
   (^ChannelHandler
-   [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address]
-   (.newHandler ssl-ctx
-                (.alloc ch)
-                (.getHostName ^InetSocketAddress remote-address)
-                (.getPort ^InetSocketAddress remote-address))))
+   [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address ssl-endpoint-id-alg]
+   (let [ssl-handler (.newHandler ssl-ctx
+                                  (.alloc ch)
+                                  (.getHostName ^InetSocketAddress remote-address)
+                                  (.getPort ^InetSocketAddress remote-address))]
+     (when ssl-endpoint-id-alg
+       (let [ssl-engine (.engine ssl-handler)
+             ssl-params (.getSSLParameters ssl-engine)]
+         (.setEndpointIdentificationAlgorithm ssl-params ssl-endpoint-id-alg)
+         (.setSSLParameters ssl-engine ssl-params)))
+     ssl-handler)))
 
 (defn- add-ssl-handler-to-pipeline-builder
   "Adds an `SslHandler` to the pipeline builder."
@@ -1400,7 +1406,10 @@ initialize an DnsAddressResolverGroup instance.
                 "ssl-handler"
                 (let [ch (.channel p)]
                   (if remote-address
-                    (ssl-handler ch ssl-ctx remote-address)
+                    ;; Always passing `nil` as `ssl-endpoint-id-alg` here since this code path is
+                    ;; only hit in the deprecated `create-client`. To keep it fully backwards
+                    ;; compatible, we don't enable endpoint identification by default.
+                    (ssl-handler ch ssl-ctx remote-address nil)
                     (ssl-handler ch ssl-ctx))))
      (pipeline-builder p))))
 
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1078,10 +1078,12 @@
 
 (defn self-signed-ssl-context
   "A self-signed SSL context for servers."
-  []
-  (let [cert (SelfSignedCertificate.)]
-    (ssl-server-context {:private-key       (.privateKey cert)
-                         :certificate-chain (.certificate cert)})))
+  ([]
+   (self-signed-ssl-context "localhost"))
+  ([hostname]
+   (let [cert (SelfSignedCertificate. hostname)]
+     (ssl-server-context {:private-key       (.privateKey cert)
+                          :certificate-chain (.certificate cert)}))))
 
 (defn insecure-ssl-client-context
   "An insure SSL context for servers."
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1382,14 +1382,14 @@ initialize an DnsAddressResolverGroup instance.
    The 3- and 4-ary version are for clients.
    For these, the `remote-address` must be provided.
    The `ssl-endpoint-id-alg` is the name of the algorithm to use for endpoint identification (see https://docs.oracle.com/en/java/javase/17/docs/specs/security/standard-names.html#endpoint-identification-algorithms). It defaults to \"HTTPS\" in the 3-ary version which is a reasonable default for non-HTTPS uses, too. Pass `nil` to disable endpoint identification."
-  (^ChannelHandler
+  (^SslHandler
    [^Channel ch ^SslContext ssl-ctx]
    (.newHandler ssl-ctx
                 (.alloc ch)))
-  (^ChannelHandler
+  (^SslHandler
    [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address]
    (ssl-handler ch ssl-ctx remote-address "HTTPS"))
-  (^ChannelHandler
+  (^SslHandler
    [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address ssl-endpoint-id-alg]
    (let [ssl-handler (.newHandler ssl-ctx
                                   (.alloc ch)
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1377,12 +1377,18 @@ initialize an DnsAddressResolverGroup instance.
 (defn ssl-handler
   "Generates a new SslHandler for the given SslContext.
 
-   The 2-arity version is for the server.
-   The 3-arity version is for the client. The `remote-address` must be provided"
+   The 2-ary version is for servers.
+
+   The 3- and 4-ary version are for clients.
+   For these, the `remote-address` must be provided.
+   The `ssl-endpoint-id-alg` is the name of the algorithm to use for endpoint identification (see https://docs.oracle.com/en/java/javase/17/docs/specs/security/standard-names.html#endpoint-identification-algorithms). It defaults to \"HTTPS\" in the 3-ary version which is a reasonable default for non-HTTPS uses, too. Pass `nil` to disable endpoint identification."
   (^ChannelHandler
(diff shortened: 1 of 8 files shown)


src/aleph/netty.clj
```diff
@@ -1384,11 +1384,17 @@ initialize an DnsAddressResolverGroup instance.
    (.newHandler ssl-ctx
                 (.alloc ch)))
   (^ChannelHandler
-   [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address]
-   (.newHandler ssl-ctx
-                (.alloc ch)
-                (.getHostName ^InetSocketAddress remote-address)
-                (.getPort ^InetSocketAddress remote-address))))
+   [^Channel ch ^SslContext ssl-ctx ^InetSocketAddress remote-address ssl-endpoint-id-alg]
```
arnaudgeiser, line 1387: I would create another arity instead.
In the current state, it might be a breaking change for some people.

arnaudgeiser (approved):
Except my comment which I consider important, it's good for me.

src/aleph/netty.clj
DerGuteMoritz: Good catch, I didn't realize that it's part of the public API. It also made me notice that the docstring doesn't match the implementation anymore :grimacing: 

I've pushed a commit to reinstate the 3-ary version and update the docstring. I also pushed another one to change the return type to `SslHandler` - seemed like a reasonable refinement.

src/aleph/netty.clj
```diff
@@ -1400,7 +1412,10 @@ initialize an DnsAddressResolverGroup instance.
                 "ssl-handler"
                 (let [ch (.channel p)]
                   (if remote-address
-                    (ssl-handler ch ssl-ctx remote-address)
+                    ;; Always passing `nil` as `ssl-endpoint-id-alg` here since this code path is
+                    ;; only hit in the deprecated `create-client`. To keep it fully backwards
+                    ;; compatible, we don't enable endpoint identification by default.
```
DerGuteMoritz, line 1417: Hm I wonder now whether this is really the wisest thing to do as it leaves users of the deprecated API vulnerable. Why not make it a (mildly) breaking change for them as well ... I think I'll go ahead and change this.

DerGuteMoritz: OK pushed a commit to that effect.

arnaudgeiser (approved):

src/aleph/netty.clj
```diff
@@ -1377,30 +1377,42 @@ initialize an DnsAddressResolverGroup instance.
...
+         (.setEndpointIdentificationAlgorithm ssl-params ssl-endpoint-id-alg)
+         (.setSSLParameters ssl-engine ssl-params)))
+     ssl-handler)))
 
 (defn- add-ssl-handler-to-pipeline-builder
   "Adds an `SslHandler` to the pipeline builder."
   ([pipeline-builder ssl-ctx]
-   (add-ssl-handler-to-pipeline-builder pipeline-builder ssl-ctx nil))
-  ([pipeline-builder ssl-ctx remote-address]
+   (add-ssl-handler-to-pipeline-builder pipeline-builder ssl-ctx nil nil))
```
KingMob, line 1408: I thought we *weren't* defaulting to nil now. Do we only want to set the "HTTPS" default higher up?

Hmmm. Not that it matters much. I've started to think these special handler-inserting functions aren't a great idea, since they lack the big picture of where in the pipeline to insert, so I was planning on removing it...at some point.

src/aleph/netty.clj
```diff
@@ -1471,6 +1483,22 @@ initialize an DnsAddressResolverGroup instance.
   ([pipeline-builder
     ssl-context
     bootstrap-transform
+    remote-address
+    local-address
+    epoll?
+    name-resolver]
+   (create-client pipeline-builder
+                  ssl-context
+                  "HTTPS"
```
KingMob, line 1492: I've counted 3 uses of "HTTPS" so far, we should probably make a const for it.

test/aleph/tcp_ssl_test.clj
```diff
@@ -69,6 +70,31 @@
         (is (nil? @(s/take! c)))
         (is (nil? @ssl-session) "SSL session should be undefined")))))
 
+(def wrong-hostname-cert
+  (SelfSignedCertificate. "some-random-hostname"))
+
+(def wrong-hostname-ssl-server-context
+  (netty/ssl-server-context
+   {:private-key       (.privateKey wrong-hostname-cert)
+    :certificate-chain (.certificate wrong-hostname-cert)}))
```
KingMob, line 79: This stuff is being replicated in a couple places...I suggest movign them to `test/aleph/ssl.clj`, where they can be shared

KingMob:
> * I would have preferred to make the option part of `:ssl-context` instead having it sit next to it since conceptually, it belongs there. The reason for not doing this is that we also support passing in an `SslContext` object. Since Netty doesn't offer an API for setting an endpoint identification algorithm on the context, this would make it impossible to reach for users doing so, forcing them to convert their `SslContext` into an options map (which for some reason or another they might not be able to do). Feel free to disagree on this tradeoff, of course.

I agree. It's not ideal, but it makes sense.

> * In the tests, I had to give up on `netty/self-signed-ssl-context` because I needed to be able to reach the certificate to properly configure the client's trust store (which in turn allowed me to remove the `:insecure? true`). Unfortunately, `SslContext` has no API for that. I guess we could still keep `netty/self-signed-ssl-context` around since it's exposed as public API. However, perhaps we should also offer an arity for passing in a hostname now?

What does "reach the certificate" mean here? You can access the certs in `test/aleph/ssl.clj` for testing purposes. Is that what you need?

We should definitely keep `netty/self-signed-ssl-context` around for users and testing. Especially with HTTP2, where TLS is on by default, and wyou have to jump through hoops not to use it.

I was also thinking about an arity for passing in a hostname. That would be nice.

test/aleph/http_test.clj
```diff
@@ -186,9 +187,20 @@
...
+   :ssl-context (netty/ssl-server-context {:private-key       (.privateKey default-ssl-certificate)
+                                           :certificate-chain (.certificate default-ssl-certificate)})})
+
 (defmacro with-server [server & body]
   `(let [server# ~server]
-     (binding [*pool* (http/connection-pool {:connection-options (merge *connection-options* {:insecure? true})})]
+     (binding [*pool* (http/connection-pool {:connection-options (merge
+                                                                  {:ssl-context {:trust-store (.certificate default-ssl-certificate)}
+                                                                   :http-versions [:http1]}
+                                                                  *connection-options* )})]
```
KingMob, line 203: I'd suggest taking a look at aleph.http-test in the current http2 PR, because I left this alone, but built on it to make it easier to run all tests against both http versions, as needed.

This section now looks like:

```clojure
(def http1-server-options {:port port})
(def http1-ssl-server-options {:port        port
                               :ssl-context (netty/self-signed-ssl-context)})
(def http2-server-options {:port        port
                           :ssl-context ssl/http2-only-ssl-context})

(defmacro with-server [server & body]
  `(let [server# ~server]
     (binding [*pool* (http/connection-pool {:connection-options (merge *connection-options* {:insecure? true})})]
       (try
         ~@body
         (finally
           (.close ^Closeable server#)
           (netty/wait-for-close server#)
           (.shutdown *pool*))))))

(defmacro with-http1-server
  [handler server-options & body]
  `(testing "- http1"
     (with-server (http/start-server ~handler ~server-options)
       ~@body)))

(defmacro with-http2-server
  [handler server-options & body]
  ;; with-redefs used so clj fns running on netty threads will work
  `(testing "- http2"
     (with-redefs [*use-tls?* true]
       (let [server-options# ~server-options]
         (with-server (http/start-server ~handler (merge server-options# http2-server-options))
           ~@body)))))
```

Copying this, and placing your change in `with-http1-server` might cause less merge issues

KingMob (changes requested):
I have a bunch of minor changes and reorganization I'd like, but I think it's fundamentally sound.

src/aleph/netty.clj
DerGuteMoritz: The 2-ary version is for server contexts where the endpoint identification algorithm doesn't have an effect anyway. And yeah, agree that getting rid of these would make sense :+1:

DerGuteMoritz: Another idea in the same vein could be to split up `ssl-handler` into `ssl-server-handler` and `ssl-client-handler`.

test/aleph/tcp_ssl_test.clj
DerGuteMoritz: Good call. Did that and refactored all SSL tests to use the helpers from that namespace consistently.

test/aleph/http_test.clj
DerGuteMoritz: I resolved it by switching all SSL tests to use the `aleph.ssl` utils instead. Now the only diff here is the removal of `:insecure? true` which is good IMHO since it could otherwise mask SSL errors when one fails to set up the client correctly. So I think conflict-wise we should be in a better place now :pray:

DerGuteMoritz:
Hm I think the failing test is a flake - it passes for me locally. It seems I am not allowed to trigger a retry, though!

DerGuteMoritz:
> Hm I think the failing test is a flake - it passes for me locally. It seems I am not allowed to trigger a retry, though!

Never mind, turns out I wasn't logged into CircleCI :facepalm:

src/aleph/netty.clj
KingMob: It's a possibility, especially the more they diverge.

KingMob:
> > Hm I think the failing test is a flake - it passes for me locally. It seems I am not allowed to trigger a retry, though!
> 
> Never mind, turns out I wasn't logged into CircleCI 🤦

We ran into similar issues with pyr. CircleCI *really* wants people to log in...

KingMob (approved):

Issue #696, opened 2023-11-19
Hide application protocol config from users with `http-versions`


KingMob:
The only useful setting is the vector of protocols for ALPN to use, so this will hide ALPN config entirely (maybe available, but undocumented), which should prevent the need to reconcile `http-versions` and the ALPN config.

PR #698, opened 2023-11-22
Centralize :port and :shutdown-timeout settings in http-test

DerGuteMoritz:
Forgetting to pass `:shutdown-timeout` to a test server (as was the case in `test-ssl-session-access` for example) adds 15s extra wait time for each instance. This patch centralizes setting it to 0 so that one doesn't have to remember doing so.

Also centralize the `:port` setting so that tests which pass extra settings only have to declare the ones which are relevant to the test.

Drop redundant with-handler-options macro.

@KingMob This came up while working on https://github.com/clj-commons/aleph/issues/696 but I didn't want to lump it in there since it's a bit noisy.

### DIFF
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -225,11 +225,12 @@
     (partition 2)))
 
 
-(def http-server-options {:port port})
-(def http1-ssl-server-options {:port        port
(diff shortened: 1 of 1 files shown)


KingMob (approved):
LGTM

arnaudgeiser (approved):

PR #699, opened 2023-11-23
Run all HTTP SSL tests against both HTTP versions, too

DerGuteMoritz:
To that end, introduce a new `with-http-ssl-servers` macro which works like `with-http-servers` but with SSL enabled for both versions. It also sets `*use-tls* true` by default.

This also fixes some tests which were passing `http1-ssl-server-options` to `with-http-servers`. This resulted in them running only against HTTP/1 (twice) because the `:ssl-context` of `with-http2-server` would get overridden.

With this change, `test-ssl-session-access` is failing in the HTTP/2 case. This is because it calls `http.core/ring-request-ssl-session` on the ring request. However, this fails because the HTTP/2 server implementation currently doesn't wrap the ring request in `aleph.http.core/NettyRequest`. @KingMob Can we just change it to do that or is there a good reason for not doing so?

Note: To avoid conflicts, this PR is based on https://github.com/clj-commons/aleph/pull/698

### DIFF
--- a/src/aleph/http/http2.clj
+++ b/src/aleph/http/http2.clj
@@ -822,12 +822,13 @@
 
      :protocol              "HTTP/2.0"
 
-     ;; These keys are internal to Aleph and should not be relied on
+     :aleph/keep-alive?     true                            ; not applicable to HTTP/2, but here for compatibility
+     :aleph/request-arrived (System/nanoTime)
+
+     ;; The following keys are internal to Aleph and should not be relied on
      :aleph/channel         ch
      :aleph/writable?       writable?
-     :aleph/h2-exception    h2-exception
-     :aleph/keep-alive?     true                            ; not applicable to HTTP/2, but here for compatibility
-     :aleph/request-arrived (System/nanoTime)}))
+     :aleph/h2-exception    h2-exception}))
 
 (defn- validate-netty-req-headers
   "Netty is not currently checking for missing pseudo-headers, so we do it here."
--- a/src/aleph/http/http2.clj
+++ b/src/aleph/http/http2.clj
@@ -822,6 +822,7 @@
 
      :protocol              "HTTP/2.0"
 
+     ;; These keys are internal to Aleph and should not be relied on
      :aleph/channel         ch
      :aleph/writable?       writable?
      :aleph/h2-exception    h2-exception
--- a/src/aleph/http/http2.clj
+++ b/src/aleph/http/http2.clj
@@ -822,6 +822,7 @@
 
      :protocol              "HTTP/2.0"
 
+     :aleph/channel         ch
      :aleph/writable?       writable?
      :aleph/h2-exception    h2-exception
      :aleph/keep-alive?     true                            ; not applicable to HTTP/2, but here for compatibility
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1190,11 +1190,12 @@
   (partial coerce-ssl-context ssl-client-context))
 
 (defn ^:no-doc channel-ssl-session [^Channel ch]
-  (some-> ch
-          ^ChannelPipeline (.pipeline)
-          ^SslHandler (.get SslHandler)
-          .engine
-          .getSession))
+  (or (some-> ch
+              ^ChannelPipeline (.pipeline)
+              ^SslHandler (.get SslHandler)
+              .engine
+              .getSession)
+      (recur (.parent ch))))
 
 (defn ^:no-doc ssl-handshake-error? [^Throwable ex]
   (and (instance? DecoderException ex)
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1190,13 +1190,14 @@
   (partial coerce-ssl-context ssl-client-context))
 
 (defn ^:no-doc channel-ssl-session [^Channel ch]
-  (or (some-> ch
-              ^ChannelPipeline (.pipeline)
-              ^SslHandler (.get SslHandler)
-              .engine
-              .getSession)
-      ;; Needed for multiplexed child channels (e.g. as present in HTTP/2)
-      (recur (.parent ch))))
+  (when ch
+    (or (some-> ch
+                ^ChannelPipeline (.pipeline)
+                ^SslHandler (.get SslHandler)
+                .engine
+                .getSession)
+        ;; Needed for multiplexed child channels (e.g. as present in HTTP/2)
+        (recur (.parent ch)))))
 
 (defn ^:no-doc ssl-handshake-error? [^Throwable ex]
   (and (instance? DecoderException ex)
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1195,6 +1195,7 @@
               ^SslHandler (.get SslHandler)
               .engine
               .getSession)
+      ;; Needed for multiplexed child channels (e.g. as present in HTTP/2)
       (recur (.parent ch))))
 
 (defn ^:no-doc ssl-handshake-error? [^Throwable ex]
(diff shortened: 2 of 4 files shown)


DerGuteMoritz:
> It also sets `*use-tls*` true by default.

Oh yeah, this variable got me confused for a moment since I thoguht it would enable TLS in the servers, too. How about renaming it to `*use-tls-requests*` or perhaps `*use-tls-request-urls*`?

KingMob:
Unfortunately `NettyRequest` is based on Netty's HTTP1 classes. IIRC, those SSL tests were fragile because they exploited the underlying fields of `NettyRequest`, instead of just treating it like a map, which is why I made them HTTP1 for the time being.

I agree that functionality should be tested, though.

Can you merge #698, or do you need someone else to do it?

Also, are you trying one of the stack tools out? I noticed this is a branch off a branch.

DerGuteMoritz:
> Unfortunately `NettyRequest` is based on Netty's HTTP1 classes. IIRC, those SSL tests were fragile because they exploited the underlying fields of `NettyRequest`, instead of just treating it like a map, which is why I made them HTTP1 for the time being.

I see! How about exposing the channel via `:aleph/netty-channel`  or so in the request map then?

> Can you merge #698, or do you need someone else to do it?

Ah I thought you didn't already merge it because you wanted @arnaudgeiser to confirm it, too (which he did in the meantime :smile:). Merged now!



> Also, are you trying one of the stack tools out? I noticed this is a branch off a branch.

Nope, in this case I just manually did it. And note what GH did after merging the other branch:

> Base automatically changed from cleanup-http-tests to master [1 minute ago](https://github.com/clj-commons/aleph/pull/699#event-11048451828)

Pretty good!

DerGuteMoritz:
> I see! How about exposing the channel via `:aleph/netty-channel` or so in the request map then?

Just realized there is precedence of exposing it via `:aleph/channel` in the stream metadata in `aleph.tcp` and `aleph.udp` as well as the websocket streams. Since we don't necessarily have a stream here and we already have precedence of adding such keys to the ring request map directly (e.g. `:aleph/request-arrived`), I suggest using `:aleph/channel` here too but in the request map.

DerGuteMoritz:
> How about renaming it to `*use-tls-requests*`

> I suggest using `:aleph/channel` here too but in the request map.

Just pushed commits for both of these!

src/aleph/http/http2.clj
```diff
@@ -822,6 +822,7 @@
 
      :protocol              "HTTP/2.0"
 
+     :aleph/channel         ch
      :aleph/writable?       writable?
      :aleph/h2-exception    h2-exception
      :aleph/keep-alive?     true                            ; not applicable to HTTP/2, but here for compatibility
```
KingMob, line 828: Can you add a comment above that these keys are internal to Aleph, and not to be relied on?

DerGuteMoritz: Done! Though now I wonder whether `:aleph/request-arrived` actually was intended to be part of the public API since there is no use of it in Aleph itself :thinking:

src/aleph/netty.clj
```diff
@@ -1190,11 +1190,12 @@
...
-          ^ChannelPipeline (.pipeline)
-          ^SslHandler (.get SslHandler)
-          .engine
-          .getSession))
+  (or (some-> ch
+              ^ChannelPipeline (.pipeline)
+              ^SslHandler (.get SslHandler)
+              .engine
+              .getSession)
+      (recur (.parent ch))))
```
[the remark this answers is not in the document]
KingMob: Hmm, maybe not an infinite loop, but an NPE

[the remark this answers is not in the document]
DerGuteMoritz: Only if the ancestor chain is circular, no? Can this indeed happen? :thinking: 

Comment added!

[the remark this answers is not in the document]
KingMob: I think throwing an IllegalArgumentException (or an ex-info wrapping one) would be preferable

src/aleph/http/http2.clj
KingMob: Huh.

Maybe it's for debugging, since nano time has higher resolution than the `Date` header.

src/aleph/netty.clj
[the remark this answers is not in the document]
DerGuteMoritz: Ah yes, missed the NPE there, indeed. I would actually go with `(some-> ch ^Channel .parent recur)` instead to preserve the nil-punning behavior. Sound good?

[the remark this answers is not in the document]
DerGuteMoritz: Or wrap the whole body in `(when ch ...)` to be less cute.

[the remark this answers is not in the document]
KingMob: > Only if the ancestor chain is circular, no? Can this indeed happen? 🤔

No, but an NPE will happen if the SslHandler isn't present.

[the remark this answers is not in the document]
KingMob: I think in that case, either nil should be returned, or a better exception class thrown.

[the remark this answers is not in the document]
DerGuteMoritz: Opted for returning nil, see https://github.com/clj-commons/aleph/pull/699/commits/ec36caf1430e2f59c9e15c25a03e7d4edd16538c

src/aleph/http/http2.clj
KingMob: Looks like Zach meant for the `:aleph/` prefix to distinguish these keys from Ring-defined keys, not to be internal. See https://github.com/clj-commons/aleph/pull/275

We could use two prefixes for each case, `:aleph/` and `:aleph.internal/` to let people know which ones they can rely on.

src/aleph/http/http2.clj
```diff
@@ -822,6 +822,8 @@
 
      :protocol              "HTTP/2.0"
 
+     ;; These keys are internal to Aleph and should not be relied on
+     :aleph/channel         ch
      :aleph/writable?       writable?
      :aleph/h2-exception    h2-exception
```
KingMob, line 828: If we add a new `:aleph.internal/` prefix, it should definitely apply to `channel`. Not sure about`writable?` and `h2-exception`. My gut says it shouldn't, but I can see scenarios where users might want to know about them in their handlers.

src/aleph/http/http2.clj
DerGuteMoritz: Good call to dig into the history :+1: `:aleph.internal/` sounds good to me, will push a change to that effect momentarily.

src/aleph/http/http2.clj
DerGuteMoritz: Right, the good thing is that people may still use them if they so desire - it's just at their own peril then :smile: I would also make these internal since they are pretty low-level what with them being `AtomicBoolean`. Seems reasonable!

KingMob (approved):
Works for me

src/aleph/http/http2.clj
KingMob: > would also make these internal 

Just so we're on the same page, you mean we leave the prefix `:aleph/`, and keep the comment saying they're internal?

We can't change the `request-arrived` and `keep-alive?` prefixes.

DerGuteMoritz: > Just so we're on the same page, you mean we leave the prefix :aleph/, and keep the comment saying they're internal?

Ah no I actually meant to change them to `:aleph.internal/`. But having just started doing that, I noticed we have a bunch of other `:aleph/` keywords which are probably also only used internally (e.g. `:aleph/omitted` or `:aleph/save-request-message` - haven't checked them all in detail, though). I suggest we leave that can of worms closed for now after all and punt this to a separate issue for future improvement where we can think this through more thoroughly. Let's keep using `:aleph/` throughout for the time being here and only move the comment so that it doesn't cover `:aleph/request-arrived` and `:aleph/keep-alive?`. Agree?

KingMob: 👍

DerGuteMoritz: Alright, pushed a commit to that effect and created https://github.com/clj-commons/aleph/issues/700 for future reconsideration.

PR #701, opened 2023-11-24
Set up ALPN configuration based on http-versions when possible

### LINKED ISSUE #696: Hide application protocol config from users with `http-versions`

DerGuteMoritz:
This also adds an `http-versions` option to `aleph.http/start-server`. Its value will be used to
automatically set up a matching ALPN config when `ssl-context` is a map and doesn't contain a
`:application-protocol-config` key, yet. Support for the latter is retained so that users can still
override the default in case any more interesting settings will become available in the future.

If an `:application-protocol-config` is present, we assert that the configured protocols match the
desired HTTP versions. We need to have that validation code path anyway for the case when an
`io.netty.handler.ssl.SslContext` instance is passed, which made the decision to retain the
`:application-protocol-config` option easier.

Closes #696.

Some review notes:

* Is it too strict to not allow more HTTP versions to be present in a custom ALPN config than specified via `:http-versions`? At first I thought so but then it would allow silly things like `:http-versions []` or might lead to accidental misconfiguration.
* I added a test case which starts a server for `:http2` without an `ssl-context`. I expected it to fail to start the server but it currently succeeds. Should we make this an error?
* Another test case enabled `use-h2c?`  but also passes an `ssl-context` in which case the latter wins. Should this be an error?

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -15,6 +15,8 @@
     [manifold.deferred :as d]
     [manifold.executor :as executor])
   (:import
+    (aleph.http.core
+      HeaderMap)
     (aleph.utils
       ConnectionTimeoutException
       PoolTimeoutException
@@ -500,7 +502,7 @@
 (defn get-all
   "Given a header map from an HTTP request or response, returns a collection of
    values associated with the key, rather than a comma-delimited string."
-  [header-m ^String k]
+  [^HeaderMap header-m ^String k]
   (let [raw-headers (.headers header-m)]
     (condp instance? raw-headers
       HttpHeaders (.getAll ^HttpHeaders raw-headers k)
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -35,16 +35,20 @@
 
    Defaults to HTTP/1.1-only.
 
-   To enable HTTP/2, you must supply an SslContext with ALPN support for HTTP/2.
-   See `aleph.netty/ssl-server-context` and `aleph.netty/application-protocol-config`
-   for more details. (You can also set `use-h2c?` to force HTTP/2 cleartext, but
-   this is strongly discouraged.)
+   To enable HTTP/2, add `:http2` to the `http-versions` option. This requires you to also enable
+   SSL via the `ssl-context` option. If you supply an `io.netty.handler.ssl.SslContext` instance, you
+   have to set it up with ALPN support for HTTP/2.  See `aleph.netty/ssl-server-context` and
+   `aleph.netty/application-protocol-config` for more details. If you supply an SSL options map
+   without an `:application-protocol-config` key instead, the necessary ALPN configuration will be
+   set up automatically. (You can also set `use-h2c?` to force HTTP/2 cleartext, but this is strongly
+   discouraged.)
 
    | Param key                         | Description |
    | ---                               | --- |
    | `port`                            | The port the server will bind to. If `0`, the server will bind to a random port. |
    | `socket-address`                  | A `java.net.SocketAddress` specifying both the port and interface to bind to. |
    | `bootstrap-transform`             | A function that takes an `io.netty.bootstrap.ServerBootstrap` object, which represents the server, and modifies it. |
+   | `http-versions`                   | An optional vector of allowable HTTP versions to negotiate via ALPN, in preference order. Defaults to `[:http1]`. |
    | `ssl-context`                     | An `io.netty.handler.ssl.SslContext` object or a map of SSL context options (see `aleph.netty/ssl-server-context` for more details) if an SSL connection is desired |
    | `manual-ssl?`                     | Set to `true` to indicate that SSL is active, but the caller is managing it (this implies `:ssl-context` is nil). For example, this can be used if you want to use configure SNI (perhaps in `:pipeline-transform` ) to select the SSL context based on the client's indicated host name. |
    | `executor`                        | A `java.util.concurrent.Executor` which is used to handle individual requests. To avoid this indirection you may specify `:none`, but in this case extreme care must be taken to avoid blocking operations on the handler's thread. |
--- a/src/aleph/http/common.clj
+++ b/src/aleph/http/common.clj
@@ -14,7 +14,7 @@
       DateFormatter
       DecoderResult
       DecoderResultProvider)
-    (io.netty.handler.ssl ApplicationProtocolNames SslContext)
+    (io.netty.handler.ssl ApplicationProtocolConfig ApplicationProtocolNames SslContext)
     (io.netty.util AsciiString)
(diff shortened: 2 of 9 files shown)


DerGuteMoritz:
> * Is it too strict to not allow more HTTP versions to be present in a custom ALPN config than specified via `:http-versions`? At first I thought so but then it would allow silly things like `:http-versions []` or might lead to accidental misconfiguration.

Having thought about it some more, I think we should be even more strict. As you can see, we only check for HTTP protocols that we support in the ALPN config and ignore anything else. That means if a user enabled HTTP/3 for example, we would just let that pass. Clients could then negotiate HTTP/3 and run into an error right away because the server is not actually able to handle it! So in conclusion, I will change it to check that the protocols in the ALPN config match the *exactly* the desired HTTP versions.

src/aleph/http/common.clj
```diff
@@ -119,6 +119,61 @@
...
+      (do (assert-consistent-alpn-config! (.supportedProtocols apc) desired-http-versions)
+          ssl-context)
+      (assoc ssl-context
+             :application-protocol-config
+             (netty/application-protocol-config desired-http-versions)))
+    (do (some-> ^SslContext ssl-context
+                (.applicationProtocolNegotiator)
+                (.protocols)
+                (assert-consistent-alpn-config! desired-http-versions))
+        ssl-context)))
```
KingMob, line 175: What happens when an SslContext is passed in with no ALPN config attached, but `desired-http-versions` is set? It looks like the `some->` will terminate early, but no error will be thrown, nor will the SslContext be updated.

src/aleph/netty.clj
```diff
@@ -900,6 +900,10 @@
     :openssl SslProvider/OPENSSL
     :openssl-refcnt SslProvider/OPENSSL_REFCNT))
 
+(def key->application-protocol-name
```
KingMob, line 903: Hmmm, how about just `->application-protocol-name`?

test/aleph/http_test.clj
```diff
@@ -1308,3 +1312,107 @@
           (is (= 200 (:status resp)))
           (is (= "" (bs/to-string (:body resp)))))))))
 
+
+(defn test-start-server-options [options]
```
KingMob, line 1316: Let's not put `test` in the name of plain, non-deftest fns

src/aleph/http/common.clj
KingMob: Hmmmm. It looks like the ALPN of the SslContext _might_ always be set...? But that might not be true if users use non-JDK, non-OpenSSL versions.

test/aleph/http_test.clj
```diff
@@ -1308,3 +1312,107 @@
...
+      (let [result (test-start-server-options
+                    {:http-versions [:http2 :http1]
+                     :ssl-context test-ssl/server-ssl-context})]
+        (is (instance? ExceptionInfo result))
+        (is (= "Some desired HTTP versions are not part of ALPN config." (ex-message result))))))
+  (testing "HTTP/2 without ssl-context"
+    (let [result (test-start-server-options
+                  {:http-versions [:http2]})]
+      ;; TODO: Shouldn't this fail?
+      (is (= :started result))))
```
KingMob, line 1407: Yes, this should fail. Either an use-h2c? or an SslContext is needed somehow.

test/aleph/http_test.clj
```diff
@@ -1308,3 +1312,107 @@
...
+    (let [result (test-start-server-options
+                  {:use-h2c? true
+                   :http-versions [:http2]})]
+      (is (= :started result))))
+  (testing "HTTP/2 with ssl-context *and* with h2c (which has no effect then)"
+    (let [result (test-start-server-options
+                  {:use-h2c? true
+                   :http-versions [:http2]
+                   :ssl-context test-ssl/server-ssl-context-opts})]
+      (is (= :started result)))))
```
KingMob, line 1418: Should also throw an error, because of port incompatibilities, if nothing else.

KingMob (changes requested):
See comments.

Minor style requests:
1. Can you keep `do` on its own line?
2. Can you add some whitespace between the `testing` macros, so it's not a wall of text?

KingMob:
Generally speaking, if we can't eliminate potentially-conflicting options, we should throw errors when they conflict.

> I added a test case which starts a server for :http2 without an ssl-context. I expected it to fail to start the server but it currently succeeds. Should we make this an error?

This should be an error unless `use-h2c?` is true.

> Another test case enabled use-h2c? but also passes an ssl-context in which case the latter wins. Should this be an error? And if not, should use-h2c? perhaps trump ssl-context? The docstring words it as "set use-h2c? to force HTTP/2 cleartext" which somewhat implies it should.

Setting `use-h2c?` and passing an SslContext should also be an error. However, if `use-h2c?` is true, setting `http-versions [:http1 :http2]` should be OK, as long as there's no SslContext.

test/aleph/http_test.clj
DerGuteMoritz: Done! And while at it, I also made it an error to enable h2c without enabling H2 (in a separate commit, so easy to drop if you disgree with that one).

src/aleph/http.clj
```diff
@@ -496,7 +502,7 @@
 (defn get-all
   "Given a header map from an HTTP request or response, returns a collection of
    values associated with the key, rather than a comma-delimited string."
-  [header-m ^String k]
+  [^HeaderMap header-m ^String k]
```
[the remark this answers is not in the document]
DerGuteMoritz: dito

DerGuteMoritz:
> 1. Can you keep `do` on its own line?
> 2. Can you add some whitespace between the `testing` macros, so it's not a wall of text?

Should both be taken care of, too!

src/aleph/http/http2.clj
```diff
@@ -423,7 +423,10 @@
                default-value
                (if (== 1 (.size vs))
                  (.toString (.get vs 0))
-                 (str/join "," vs))))))))
+                 (str/join "," vs)))))))
+  common/HeaderMap
+  (get-header-values [_ k]
+                     (.getAll headers ^String k)))
```
KingMob, line 429: You know... I'm not sure I knew this would work.

I've known for a while that you can sneak extra methods into Clojure-generated Java classes, even if not descirbed by a protocol/interface (iirc), but I kind of assumed `def-map-type` was more restricted. Probably should have taken a closer look at it.

KingMob (approved):
Looks good to me. Let's merge and aim to deploy an RC early next week

Issue #707, opened 2024-01-03
Track down potential idle-handler refcount testing leak


KingMob:
# Summary
When running tests in paranoid leak mode, there are some ByteBuf ref count leaks (where the ByteBuf is GCed, but the reference count wasn't 0). The final access consistently involves the idle handlers, which aren't used in a lot of tests, so I wonder if they're connected. It's also possible the error occurs earlier in the ByteBuf's life, and the idle handler just happens to be the last one

This may not be a real issue for production, since tests tear down things quickly, and simply may not be allowing proper time for clean up. It would be good to track this down, and make sure it's not a real leak.

# To reproduce

`lein with-profile +leak-level-paranoid test`

Interestingly, running `lein with-profile +leak-level-paranoid test :only aleph.http-test/test-idle-timeout` or `lein with-profile +leak-level-paranoid test :only aleph.http-test/test-compressed-response` by themselves don't turn up the problem, which is odd because they're the only tests that add the idle handlers afaict, which lends a little credence to the idea that the idle handlers are not the direct cause.

# Sample output

```
202959 ERROR ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records:
#1:
	io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:300)
	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	aleph.netty$close_on_idle_handler$reify__6723.channelRead(netty.clj:1621)
	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:286)
[...]

Issue #615, opened 2022-08-05
Netty resource leak detector too unreliable?


DerGuteMoritz:
As per [this comment](https://github.com/clj-commons/aleph/pull/484#issuecomment-1206366457), I get the impression that Netty's builtin leak detector is too unreliable to be useful. For example, [when introducing an intionally leaking test case](https://github.com/bevuta/aleph/commit/4f6554fe2310b237f4fd82f76f4f093e7a7e4ade) like this:

```clj
(deftest test-leak-in-raw-stream-handler
  (with-raw-handler basic-handler
    (let [resp @(http-put (str "http://localhost:" port "/string")
                          ;; NOTE: The request handler doesn't consume
                          ;; this body. As per the :raw-stream?
                          ;; contract, this leads to a buffer leak.
                          ;; However, Netty's leak detector in
                          ;; paranoid mode still doesn't pick it up
                          ;; most of the time.
                          {:body "Hello, world!"})]
      (is (= 200 (:status resp)))
      (is (= "String!" (bs/to-string (:body resp)))))))
```

Then `lein test` will happily pass without reporting any leaks even when passing `-Dio.netty.leakDetection.level=paranoid`. Only when repeatedly running the same case (about 10-20x) via a REPL, the detector eventually notices it. It seems that there is some threshold even in paranoid mode which needs to be reached before the detector fires. So far, I was unable to find any indication of that in [the docs](https://netty.io/wiki/reference-counted-objects.html#troubleshooting-buffer-leaks), though. On the contrary, it states there:

> PARANOID - Same with ADVANCED except that it's for every single buffer. Useful for automated testing phase.

Which seems to confirm that we're using it correctly here. Any ideas?

arnaudgeiser:
It's useful and reliable on `PARANOID` mode but you will see the traces only when the GC is triggered.
Some objects are eligible for being GC but still get a `.refcount` of 1.

DerGuteMoritz:
@arnaudgeiser  Ah, right, forgot to mention in the ticket description that even adding an explicit `(System/gc)` to the end of the test doesn't help either. I also tried `-XX:+UseSerialGC` assuming that this one is a bit more deterministic but no dice either. Perhaps there is some other parameter one could tweak to make this work a bit more reliably? :thinking:

arnaudgeiser:
Porting my comment here:

[...]

PR #708, opened 2024-02-09
Add more reliable resource leak detector for tests

DerGuteMoritz:
See docstrings in `aleph.resource-leak-detector` for details.

This patch also includes the following changes:

* Rename the `:leak-level-paranoid` Leiningen profile to `:leak-detection` since it now also sets everything up to use the new custom detector.
* Introduce new test selector `:leak` for tests which intentionally leak (not run by default since they require the `:leak-detection` profile to be active, too).
* Introduce intentionally leaking `aleph.http-test/test-leak-in-raw-stream-handler` as an example and to document a case where users are responsible for releasing buffers.
* Eliminate all `(netty/leak-detector-level! :paranoid)` calls from tests. Use new leak detection API instead (see below).
* Place `(aleph.resource-leak-detector/instrument-tests!)` at the end of all relevant test namespaces to detect leaks (only effective when running with `:leak-detection` profile).

Ideally we'd activate it in the CI test job but we first need to fix the leak in `aleph.http-test/test-idle-timeout` for that (see #707).

Based on the research results from  #615.

### DIFF
--- a/test/aleph/resource_leak_detector.clj
+++ b/test/aleph/resource_leak_detector.clj
@@ -1,0 +1,210 @@
+(ns aleph.resource-leak-detector
+  "Provides a Netty leak detector which is more reliable than the default implementation.
+
+  Its chance of detecting leaks is around 95%. See `with-leak-collection` for details.
+
+  Since it adds considerable runtime overhead, its main purpose is to be used in tests to assert
+  intentional presence of leaks and to check for unintended leaks. The main API entry points for
+  this are `with-leak-collection`, `with-expected-leaks` and `instrument-tests!`.
+
+  To enable it, pass `-Dio.netty.customResourceLeakDetector=aleph.resource_leak_detector` to the
+  JVM. You will also have to pass `-Dio.netty.leakDetection.level=PARANOID`.
+
+  Most reliable results are achieved by also passing the following options:
+  `-Dio.netty.leakDetection.targetRecords=1`
+  `-Dio.netty.allocator.type=unpooled`
+
+  All of the above is also provided by the `:leak-detection` Leiningen profile.
+
+  NOTE: Currently only improves reliability for detecting leaked ByteBufs. Other types of leaked
+  resources will still be detected but not with the same reliability. Search Netty's codebase for
+  invocations of `newResourceLeakDetector` to see which other candidate resources types there are."
+  (:gen-class
+   :extends io.netty.util.ResourceLeakDetector)
+  (:require
+   [clojure.test :as test]
+   [clojure.tools.logging :as log])
+  (:import
+   (io.netty.buffer AbstractByteBufAllocator)
+   (io.netty.util
+    ResourceLeakDetector
+    ResourceLeakDetector$Level
+    ResourceLeakDetectorFactory)))
+
+(defn enabled?
+  "Checks whether the resource leak detector is enabled.
+
+  See `aleph.resource-leak-detector` docstring on how to enable it."
+  []
+  (= "aleph.resource_leak_detector"
+     (System/getProperty "io.netty.customResourceLeakDetector")))
+
+(def active-resource-leak-detector-class
+  (delay
+    (class (.newResourceLeakDetector (ResourceLeakDetectorFactory/instance) String))))
+
+(def active?
+  (delay
+    (= (Class/forName "aleph.resource_leak_detector")
+       @active-resource-leak-detector-class)))
+
+(defn ensure-consistent-config! []
+  (when-not @active?
+    (if (enabled?)
+      (throw (RuntimeException.
+              (str "`aleph.resource_leak_detector` is enabled but the active resource leak detector is "
+                   "`" (.getName ^Class @active-resource-leak-detector-class)"`. This indicates that "
+                   "`io.netty.util.ResourceLeakDetectorFactory` ran into an initialization error. "
+                   "Enable Netty debug logging to diagnose the cause.")))
+      (throw (RuntimeException.
+              (str "Attempted to use `aleph.resource-leak-detector` API but it is not enabled. "
+                   "Pass `-Dio.netty.customResourceLeakDetector=aleph.resource_leak_detector` to enable.")))))
+  (when-not (= ResourceLeakDetector$Level/PARANOID (ResourceLeakDetector/getLevel))
+    (throw (RuntimeException.
+            (str "`aleph.resource_leak_detector` requires `-Dio.netty.leakDetection.level=PARANOID`. "
+                 "Current level is `" (ResourceLeakDetector/getLevel) "`.")))))
+
+(def +max-probe-gc-runs+
+  "Maximum number of times the GC will be run to detect a leaked probe."
+  10)
+
+(def +probe-hint-marker+
+  "ALEPH LEAK DETECTOR PROBE")
+
+(defn hint-record-pattern [hint-pattern]
+  (re-pattern (str "(?m)^\\s*Hint: " hint-pattern "$")))
+
+(def +probe-hint-pattern+
+  (hint-record-pattern (str +probe-hint-marker+ " \\d+")))
+
+(defn probe? [leak]
+  (re-find +probe-hint-pattern+ (:records leak)))
+
+(defn contains-hint? [hint leak]
+  (re-find (hint-record-pattern hint) (:records leak)))
+
+(defn remove-probes [leaks]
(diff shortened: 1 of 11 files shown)


DerGuteMoritz:
Hmm `aleph.http-timeout-test/test-shutdown-timeout-1` fails consistently in CI for me while it works fine locally. Looks like there is a race condition there where the client has not yet established a connection before the server shuts down after the 50ms wait? I can't think of a way to make this more reliable, though... @pyr Since you are the author, do you have an idea?

FWIW I also pushed a branch with only a bogus change to trigger the job and [this one also fails](https://app.circleci.com/pipelines/github/clj-commons/aleph/624/workflows/4c422f9f-886e-4ddc-bdbc-fd5f3f54fc86/jobs/603) so it's not related to any change in this here patch.

DerGuteMoritz:
Well, that fixed it but now the DNS failure tests started timing out, as well (also on `master`). See https://github.com/clj-commons/aleph/pull/710 for an improvement.

DerGuteMoritz:
OK all green now :sweat_drops: Please give it a try!

test/aleph/resource_leak_detector.clj
```diff
@@ -0,0 +1,210 @@
...
+  this are `with-leak-collection`, `with-expected-leaks` and `instrument-tests!`.
+
+  To enable it, pass `-Dio.netty.customResourceLeakDetector=aleph.resource_leak_detector` to the
+  JVM. You will also have to pass `-Dio.netty.leakDetection.level=PARANOID`.
+
+  Most reliable results are achieved by also passing the following options:
+  `-Dio.netty.leakDetection.targetRecords=1`
+  `-Dio.netty.allocator.type=unpooled`
+
+  All of the above is also provided by the `:leak-detection` Leiningen profile.
```
KingMob, line 17: Minor: Excellent docs, but I suggest reversing the order here. Mention the :leak-detection profile first, then go into what it's doing under the hood. (I assume 99% of the time we'll just want to add `with-profile +leak-detection`?)

test/aleph/resource_leak_detector.clj
```diff
@@ -0,0 +1,210 @@
...
+              (str "Attempted to use `aleph.resource-leak-detector` API but it is not enabled. "
+                   "Pass `-Dio.netty.customResourceLeakDetector=aleph.resource_leak_detector` to enable.")))))
+  (when-not (= ResourceLeakDetector$Level/PARANOID (ResourceLeakDetector/getLevel))
+    (throw (RuntimeException.
+            (str "`aleph.resource_leak_detector` requires `-Dio.netty.leakDetection.level=PARANOID`. "
+                 "Current level is `" (ResourceLeakDetector/getLevel) "`.")))))
+
+(def +max-probe-gc-runs+
+  "Maximum number of times the GC will be run to detect a leaked probe."
+  10)
```
KingMob, line 69: Style: No need for Common Lisp-style constants (`+some-const+`) since everything in clj is immutable by default.

KingMob: Might want to consider adding `^:const` though, if they're compile-time literals like numbers or strings, since then they can be inlined.

test/aleph/resource_leak_detector.clj
```diff
@@ -0,0 +1,210 @@
...
+    (->> (ns-interns *ns*)
+         vals
+         (filter (comp :test meta))
+         (run! (fn [tv]
+                 (alter-meta! tv update :test instrument-test-fn))))))
+
+(if (enabled?)
+  (log/info "enabled.")
+  (log/info "disabled. This means resource leaks will be reported less accurately."
+            "Pass `-Dio.netty.customResourceLeakDetector=aleph.resource_leak_detector` to enable."))
```
KingMob, line 210: Minor: add something useful before "enabled/disabled" in case the log settings don't display the namespace.

test/aleph/resource_leak_detector.clj
```diff
@@ -0,0 +1,210 @@
...
+  the same thread, so any asynchronous processes started by and outliving `f` will leak into the
+  parent scope(s)."
+  [f handle-leaks]
+  (ensure-consistent-config!)
+  (with-redefs [current-leaks (atom [])]
+    (f)
+    (let [hint (gen-probe-hint)]
+      (leak-probe! hint)
+      (await-probe! hint)
+      (handle-leaks (remove-probes @current-leaks)))))
```
KingMob, line 139: Interesting technique. Is there some doc on it? 

Why is a probe needed? Why not listen for all leaks when running `(f)`?

Is there a race? Can we miss leaks because we call `f` before setting up leak detection?

If we ever run parallel tests, using https://github.com/mourjo/dynamic-redef might help.

KingMob (changes requested):
Generally looks good to me, but see my comments.

test/aleph/resource_leak_detector.clj
DerGuteMoritz: Oh yeah, good call, also done :+1: I didn't mark `max-probe-gc-runs` as const so that one can experiment with it on the REPL (as suggested by the corresponding error raised by `await-probe!`).

test/aleph/resource_leak_detector.clj
DerGuteMoritz: > Interesting technique. Is there some doc on it?

Nothing beyond what's in the docstring of `with-leak-collection`.

> Why is a probe needed? Why not listen for all leaks when running `(f)`?

We do listen for all leaks when running `(f)` via the `(with-redefs [current-leaks (atom [])] ...)`.

To understand why the probe is needed, you need to know that the detector can only detect leaked resources after they have been marked as garbage by the GC. Now since the probe is only leaked *after* `f` has returned, there's a higher chance that any leaks which occurred during `f` will have been detected by the time the detector reports the probe again. There's still a small chance that the GC will mark the probe as garbage before it marked all resources leaked by `f`, which is why we end up with only ~95% accuracy. 

> Is there a race?

Not really a race but more of a gap of unreliability due to the GC being non-deterministic. Even our explicit `System/gc`  loop doesn't help since that only serves as a hint to the GC. I did extensive research on whether there is a way to force a full GC somehow, but it appears there isn't :confused: 

> Can we miss leaks because we call `f` before setting up leak detection?

As mentioned above, leak detection is indeed set up before `f`!


> If we ever run parallel tests, using https://github.com/mourjo/dynamic-redef might help.

Indeed, though I would tend to always run tests serially for leak detection since the GC is the most unpredictable component involved. Since the GC owuld be shared across all threads, I would expect this to further reduce reliability. But we can cross that bridge when we get to it I guess :sweat_smile:

KingMob: Sounds good. I think the word "probe" confused me, because, IIUC, the probe isn't "probing" for leaks itself, but is used as a marker to indicate which leaks were caused by `f` (i.e., all leaks before the probe). Is that right, or am I still misunderstanding?

test/aleph/resource_leak_detector.clj
DerGuteMoritz: True, done!

test/aleph/resource_leak_detector.clj
DerGuteMoritz: Yep, that's right! Or ... well, it isn't really a *marker* for which leaks came before it (during `f`) but rather as a *trigger* to flush out any leaks which happened during `f` but which have not yet been marked by the GC. So maybe `gc-trigger` could be a better name? Any other ideas?

DerGuteMoritz:
> I'm not sure it will fly long-term (the strategy around calling the GC - since it's just a hint. And manually running the finalizations).

@arnaudgeiser Now that you mention it: running finalizers should actually not be needed, I only threw `runFinalization` in there for good measure. Netty does use a handful of finalizers but the leak detector uses weak references and a reference queue instead. I think leaving it in there for now doesn't hurt for now but having to remove it shouldn't be a problem either - if it ever gets removed completely, Netty will have to adjust as well, anyway. And by then Netty 5 migh thave rolled around which AFAIK gets rid of reference counting anyway :shrug: 

> But it improves the current situation and it's just for testing purpose. Let's see what will be the future.

Right, improving the current testing situation is the main goal here :+1:  


> Thanks for the approach!

You're welcome and thanks for your input on the original ticket :pray: 

 
> > https://openjdk.org/jeps/421

Ah, good to know, wasn't aware of this!

KingMob (approved):

PR #687, opened 2023-08-31
HTTP/2 server

KingMob:
The code is ready for review, though there are admittedly, some details to work out and tests to fix. 

Some things to know:

1. There's extremely little continuity in the Netty code, so a lot of things are more different than they need to be. Some of that is genuinely caused by how HTTP/2 has a completely different binary format and multiplexed streams, unlike HTTP/1, but a lot is just Netty doing it differently over time.

1. Organizationally, truly common code has been moved to a.http.common. Code specific to HTTP2, but shared between client and server is in a.http.http2. HTTP1-specific code is left wherever it was, in a.http.core/server/client. Public functions in a.http.server and a.http.client should handle both HTTP versions. This is messier than I'd like, but since we never hid the http sub-namespaces, there's a chance someone's code is using them. Otherwise, I'd have made a much cleaner design.

1. Missing features: this does not support proxies or multipart. Multipart is kind of pointless in HTTP2 (just open a new stream for each part), but understandably some people want to keep using it. I think it's fine to tell them to wait, since Netty lacks all support for it in HTTP2, and we'd have to figure something out. I was working on shifting to the HTTP1 pathway when multipart is used, but never finished.

1. `pipeline-transform` - fundamentally this assumed a single pipeline, but multiplexing breaks that. We now have 2 types of pipeline: a general connection pipeline that handles common concerns like SSL and HTTP frame decoding, and many (identical) stream pipelines that each handle one stream, aka a request/response pair. There's no safe way to know what the user intended with pipeline-transform, so I think perhaps two new params (conn-pipeline-transform and stream-pipeline-transform?) should be added, and an error should be thrown if the user supplies _only_ pipeline-transform, and is using http2. It's not ideal, but it will prevent subtle problems.

1. Some good news: While I think the multiplexing and Netty changes added some complexity, I believe the fundamental core is actually _simpler_ in the HTTP2 code, because requests/responses are isolated in their own streams. There's no crazy state-tracking to know when one request/resp is finished, nor are there multiple queues feeding in/out on the client side.

1. The broken tests are a mix of fragile, weird, and different.

- - -
This change is [Reviewable](https://reviewable.io/reviews/clj-commons/aleph/687)

PR #709, opened 2024-02-09
Drop fixed HTTP server shutdown timeout

DerGuteMoritz:
This is a WIP artifact from the patch which introduced HTTP/2 support. It is no longer needed and dropping it reduces test suite runtime from roughly 5 minutes to roughly 40 seconds.

See https://github.com/clj-commons/aleph/pull/687#discussion_r1402354975 for discussion.

### DIFF
--- a/src/aleph/netty.clj
+++ b/src/aleph/netty.clj
@@ -1729,9 +1729,7 @@
                    (d/finally'
                      ;; 3. At this stage, stop the EventLoopGroup, this will cancel any
                      ;;    in flight pending requests.
-                     ;;    We still wait 1.5s to gracefully end repeating tasks
-                     ;;    like the date-header-value.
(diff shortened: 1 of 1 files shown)


DerGuteMoritz:
Aha, coincidentally, this patch appears to fix the timeout issue mentioned [here](https://github.com/clj-commons/aleph/pull/708#issuecomment-1935962752), as well! Either that, or some environmental factor at CircleCI has changed in the meantime...

KingMob:
Thanks for following up on this. LGTM. Though I suspect it may only be masking any resource leak issue...

KingMob (approved):

DerGuteMoritz:
> Though I suspect it may only be masking any resource leak issue...

If so, #708 will hopefully uncover it!

PR #714, opened 2024-02-12
HTTP client request cancellation

### LINKED ISSUE #712: Support cancellation of HTTP requests
# Problem

At the moment it is not possible to cancel HTTP requests (be it in-flight or during connection establishment). The only option users have is to just drop the response deferred on the floor and have it GC'ed. However, this doesn't free up the underlying connection until the response is complete. Combined with the fact that there is no default `request-timeout`, this can lead to resource exhaustion. 

Manifold's idiom for cancelling a deferred is to put it into an error state but that only affects *chained* deferreds downstream from the "cancelled" one (see also https://github.com/clj-commons/manifold/issues/167). Since the response deferred returned by `aleph.http/request` sits at the very end of the deferred chain, putting it into an error state has no repercussion on any of the upstream deferreds nor on the underlying connection. 

This also means that placing a timeout on a response deferred to limit the overall request duration (i.e. connection setup, request transmission and response reception) doesn't have the effect a user might expect:

```clojure
@(-> (http/get "...") (d/timeout! 1000))
```

As explained above, this will simply put the response deferred into an error state when the timeout expires but the underlying request operation will still proceed, tying up any resources it has acquired until done.

# Solutions

The most elegant solution would be to change Manifold to also cancel any upstream deferreds which feed into a "cancelled" deferred (if there are no others left). However, [according to Zach](https://github.com/clj-commons/manifold/issues/166#issuecomment-459902958), this cannot be implemented in the current architecture of Manifold and would require a "major reengineering".

A more tractable solution would be to explicitly cancel any upstream operations when the response deferred is set into an error state.

DerGuteMoritz:
This patch changes `aleph.http/request` so that setting the response deferred to an error status will terminate an in-flight request. This allows e.g. for `d/timeout!` to be used without potentially leaking connections.

For convenient explicit cancellation, we provide `aleph.http/cancel-request!`. It puts the given response deferred into error state with an instance of the new `aleph.utils.RequestCancellationException`.

Closes #712.

Review notes:
* ~The patch uses the approach outlined in https://github.com/clj-commons/manifold/issues/167#issuecomment-457996116 for propagating cancellation.~  (not quite, it was merely an inspiration)
* Diff best viewed [without whitespace changes](https://github.com/clj-commons/aleph/pull/714/files?w=1).
* ~Still marked as draft because the patch only pertains to in-flight requests. Connection establishment is not yet cancellable. Combined with the default `connection-timeout` of 60s, this can make for an unpleasant surprise :grimacing:~ -- deferred to a follow-up patch once Netty has responded.

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -397,14 +397,7 @@
                                         ;; connection timeout triggered
                                         (d/catch' TimeoutException
                                                   (fn [^Throwable e]
-                                                    (log/error e "Timed out waiting for connection to be established")
                                                     (d/error-deferred (ConnectionTimeoutException. e))))
-
-                                        ;; connection failed, bail out
-                                        (d/catch'
-                                         (fn [e]
-                                           (log/error e "Connection failure")
-                                           (d/error-deferred e)))
 
                                         ;; actually make the request now
                                         (d/chain'
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -20,6 +20,7 @@
       ConnectionTimeoutException
       PoolTimeoutException
       ReadTimeoutException
+      RequestCancellationException
       RequestTimeoutException)
     (io.aleph.dirigiste Pools)
     (io.netty.handler.codec Headers)
@@ -336,6 +337,9 @@
      by [clj-http](https://github.com/dakrone/clj-http), and returns a deferred representing
      the HTTP response.  Also allows for a custom `pool` or `middleware` to be defined.
 
+     Putting the returned deferred into an error state will cancel the underlying request if it is
+     still in flight.
+
      Param key            | Description
      -------------------- | -----------------------------------------------------------------------------------------------------------------------------------------------------------------
      `connection-timeout` | timeout in milliseconds for the connection to become established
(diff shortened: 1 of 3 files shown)


DerGuteMoritz:
Completion of this patch depends on how the Netty issue linked above will be resolved.

KingMob:
Some impressive spelunking!

> Completion of this patch depends on how the Netty issue linked above will be resolved.

Would we not still want to do this even if Netty doesn't want to change this for v4? After all, persistent conns are pretty common, so canceling the initial `connect` shouldn't be that relevant in those cases, no?

DerGuteMoritz:
> Some impressive spelunking!

You can say that again :see_no_evil: 

> Would we not still want to do this even if Netty doesn't want to change this for v4? After all, persistent conns are pretty common, so canceling the initial `connect` shouldn't be that relevant in those cases, no?

Yeah, the in-flight cancellation on its own would be good to have already, indeed. We could of course already merge this as-is and defer the connection establishment cancellation to a follow-up.

I propose to include one more change, though - commit coming up!

DerGuteMoritz:
@KingMob See https://github.com/clj-commons/aleph/pull/714/commits/9e27caec66b31f8b86813c45ba21fd2b2443e447 :pray:

src/aleph/http.clj
```diff
@@ -336,6 +337,9 @@
      by [clj-http](https://github.com/dakrone/clj-http), and returns a deferred representing
      the HTTP response.  Also allows for a custom `pool` or `middleware` to be defined.
 
+     Putting the returned deferred into an error state will cancel the underlying request if it is
+     still in flight.
+
```
KingMob, line 342: Maybe add a note to the effect that partial processing may still have occurred.

src/aleph/http.clj
```diff
@@ -358,96 +362,104 @@
...
+                                           (->> rsp
+                                                (middleware/handle-cookies req)
+                                                (middleware/handle-redirects request req))))))))))))
+                      req))]
+      (d/connect response result)
+      (d/catch' result
+                (fn [e]
+                  (log/trace e "Request failed. Disposing of connection...")
+                  (@dispose-conn!)
+                  (d/error-deferred e)))
```
KingMob, line 452: Won't it log "request failed" even if the _user_ puts it in an error state? That'll be confusing.

Maybe "Request failed or canceled."?

test/aleph/http_test.clj
```diff
@@ -1438,3 +1442,18 @@
...
+                       (deliver conn-established true)
+                       ;; Required for the client close to be detected
+                       (s/consume identity s)
+                       (s/on-closed s (fn []
+                                        (deliver conn-closed true))))
+      (let [rsp (http-get "/")]
+        (is (= true (deref conn-established 1000 :timeout)))
+        (http/cancel-request! rsp)
+        (is (= true (deref conn-closed 1000 :timeout)))
+        (is (thrown? RequestCancellationException (deref rsp 1000 :timeout)))))))
```
KingMob, line 1459: I'd like to see more testing, for both normal and error conditions. Maybe via overriding the conn-pool in some cases.

What's all the TCP handler/server/etc stuff get us? Why not the the HTTP code?

src/aleph/http.clj
DerGuteMoritz: Good idea! Did that and also touched up the explanation to make explicit that cancellation means closing of the TCP connection. It now also mentions that cancellation will have no effect if the request had already completed by the time of invoking `cancel-request!`.

src/aleph/http.clj
DerGuteMoritz: Fair enough, Or we could log a different message depending on `e`'s type?

BTW I just noticed that I was somewhat wrong about the original error handlers being redundant: the original handlers used `log/error` while this one here uses `log/trace`, so there's a user-visible change. I for one would like to change it this way, though, as I feel the `log/error` doesn't add much for users which I expect to deref the response anyway, in which case they will receive the exception which would then (usually) generate a corresponding error log. And if the user decides to handle e.g. a timeout error explicitly, it would also be nicer to not spam the error log. WDYT?

DerGuteMoritz: Oh yeah, also note that the exception's message will say "Request cancelled" in case the user cancelled the request.

DerGuteMoritz: (and since that's logged right along with the "Request failed ..." it might be alright as-is after all?)

test/aleph/http_test.clj
DerGuteMoritz: > I'd like to see more testing, for both normal and error conditions. 

You mean for the case where `cancel-request!` is invoked after the request had already completed successfully or due to some other error?

> What's all the TCP handler/server/etc stuff get us? Why not the the HTTP code?

IMHO using a raw TCP server which only  accepts and holds the connection is the cleanest way to reliably test whether cancellation closes the underlying TCP connection (via the `s/on-closed` handler). One could probably concoct something like it with a `raw-stream?` HTTP request handler but I don't think it adds much value, does it?

DerGuteMoritz: I'll give the latter idea a shot in any case!

DerGuteMoritz: OK, the test uses the HTTP server(s) now, works just as well and probably causes less confusion :+1:

src/aleph/http.clj
```diff
@@ -358,96 +362,106 @@
...
+                                                      (-> rsp
+                                                          (dissoc :aleph/destroy-conn?)
+                                                          (assoc :connection-time (- end start)))))))))
+
+                                         (fn handle-response [rsp]
+                                           (->> rsp
+                                                (middleware/handle-cookies req)
+                                                (middleware/handle-redirects request req))))))))))))
+                      req))]
+      (d/connect response result)
```
arnaudgeiser, line 447: I'm not sure to understand, why do we need to `connect` here?
We cannot just `catch'` on the `response` deferred?

src/aleph/http.clj
```diff
@@ -358,96 +362,106 @@
...
+      (d/catch' result
+                (fn [e]
+                  (log/trace e "Request failed. Disposing of connection...")
```
arnaudgeiser, line 450: By catching on a central place, we are going to lose the granularity of the log message, right?
Since it's a `trace` one, I don't think it's a big deal though.

arnaudgeiser (commented):
I looks good to me, just a few comments.

src/aleph/http.clj
DerGuteMoritz: Yeah this is a bit subtle: In the context of this patch it's for [handling this race](https://github.com/clj-commons/aleph/blob/75c4cbf3904026d1c4223f8588850c52be9115dd/src/aleph/http.clj#L389-L392) since the `response` deferred cannot refer to itself. In the follow-up patch for cancellation of connection establishment, we will also need to connect the result deferred to `(first conn)` to support it, which again would require `response` to refer to itself... Hope that makes sense!

src/aleph/http.clj
DerGuteMoritz: Indeed. I guess we could add another `log/trace` to the timeout error handler but then again, it's the only place where `RequestTimeoutException` is returned so logging it only here should be good enough.

src/aleph/http.clj
arnaudgeiser: Got it, thanks for the explanation.

KingMob (approved):

PR #715, opened 2024-02-14
Expose Netty's CONNECT_TIMEOUT_MILLIS client channel option

DerGuteMoritz:
Netty has a default connect timeout for client channels (currently 30s) which was only accessible by means of `bootstrap-transform`. Since it's a rather important option to get predictable connection behavior, expose it as `connect-timeout` in all relevant APIs.

This patch came about while working on connection establishment cancellation in the context of #714. Turns out that cancelling a `connect` promise currently has no effect in Netty. Instead, the connection would stick around for roughly 30s. Now I was using the HTTP client for this particular test whose [`connection-timeout` defaults to 60s](https://github.com/clj-commons/aleph/blob/634c0348d0e10572266ab1e5c57db6561c207278/src/aleph/http.clj#L361) so this didn't line up. Grepping Netty's source code for "30000" quickly lead me to the [DEFAULT_CONNECT_TIMEOUT](https://github.com/clj-commons/aleph/blob/634c0348d0e10572266ab1e5c57db6561c207278/src/aleph/http.clj#L361) constant. This of course means that the HTTP client's default of 60s is practically ineffective since Netty's default connect timeout would always fire first! The test case I include in the patch also demonstrates this.

To address this, I thought about changing the default of `connection-timeout` to `aleph.netty/default-connect-timeout`, too, but then you could sometimes end up with Netty's `ConnectTimeoutException` and sometimes with Aleph's `ConnectionTimeoutException` depending on who wins the race. Another option could be to enforce that `connection-timeout < connect-timeout` but that could of course still be racey if the difference is very low. Yet another option would be to wait for [HTTP client request cancellation](https://github.com/clj-commons/aleph/pull/714) to be ready. Then we could set the default pool's `connect-timeout` to 0 (= infinite) and handle it completely on our end via `connection-timeout` (which would then cancel the connection attempt). Thoughts?

### DIFF
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -23,6 +23,7 @@
       RequestCancellationException
       RequestTimeoutException)
     (io.aleph.dirigiste Pools)
+    (io.netty.channel ConnectTimeoutException)
     (io.netty.handler.codec Headers)
     (io.netty.handler.codec.http HttpHeaders)
     (java.net
@@ -396,10 +397,18 @@
                                     (-> (first conn)
                                         (maybe-timeout! connection-timeout)
 
-                                        ;; connection timeout triggered
-                                        (d/catch' TimeoutException
-                                                  (fn [^Throwable e]
-                                                    (d/error-deferred (ConnectionTimeoutException. e))))
+                                        ;; connection establishment failed
+                                        (d/catch'
+                                         (fn [e]
+                                           (if (or (instance? TimeoutException e)
+                                                   ;; Unintuitively, this type doesn't inherit from TimeoutException
+                                                   (instance? ConnectTimeoutException e))
+                                             (do
+                                               (log/error e "Timed out waiting for connection to be established")
+                                               (d/error-deferred (ConnectionTimeoutException. ^Throwable e)))
+                                             (do
+                                               (log/error e "Connection failure")
+                                               (d/error-deferred e)))))
 
                                         ;; actually make the request now
                                         (d/chain'
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -507,7 +507,7 @@
    | `follow-redirects?`  | whether to follow redirects, defaults to `true`; see `aleph.http.client-middleware/handle-redirects`
    | `pool`               | a custom connection pool
    | `pool-timeout`       | timeout in milliseconds for the pool to generate a connection
-   | `connection-timeout` | timeout in milliseconds for the connection to become established, defaults to 60s. Note that this timeout will be ineffective if the pool's `connect-timeout` is lower.
+     `connection-timeout` | timeout in milliseconds for the connection to become established, defaults to `aleph.netty/default-connect-timeout`. Note that this timeout will be ineffective if the pool's `connect-timeout` is lower.
    | `request-timeout`    | timeout in milliseconds for the arrival of a response over the established connection
    | `read-timeout`       | timeout in milliseconds for the response to be completed
    | `response-executor`  | optional `java.util.concurrent.Executor` that will handle the requests (defaults to a `flow/utilization-executor` of 256 `max-threads` and a `queue-length` of 0)")
--- a/src/aleph/http.clj
+++ b/src/aleph/http.clj
@@ -345,7 +345,7 @@
 
      Param key            | Description
      -------------------- | -----------------------------------------------------------------------------------------------------------------------------------------------------------------
-     `connection-timeout` | timeout in milliseconds for the connection to become established, defaults to 60s. Note that this timeout will be ineffective if the pool's `connect-timeout` is lower.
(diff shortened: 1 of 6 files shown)


KingMob:
I thought for a hot minute the two exceptions would share a useful ancestor class, but nope.

Maybe we should handle both exception classes in out `catch`es, and set the netty timeout value from the aleph value.

src/aleph/http.clj
```diff
@@ -482,7 +484,7 @@
    | `follow-redirects?`  | whether to follow redirects, defaults to `true`; see `aleph.http.client-middleware/handle-redirects`
    | `pool`               | a custom connection pool
    | `pool-timeout`       | timeout in milliseconds for the pool to generate a connection
-   | `connection-timeout` | timeout in milliseconds for the connection to become established
+   | `connection-timeout` | timeout in milliseconds for the connection to become established, defaults to 60s. Note that this timeout will be ineffective if the pool's `connect-timeout` is lower.
```
arnaudgeiser, line 487: Instead of only documenting, I would consider throwing an Exception if the `connection-timeout` is higher than the `connect-timeout`.

src/aleph/netty.clj
```diff
@@ -1516,6 +1518,14 @@
                     (ssl-handler ch ssl-ctx))))
      (pipeline-builder p))))
 
+(def ^:const
+  default-connect-timeout
```
arnaudgeiser, line 1522: I would just put the scalar (duplicate) with a comment that points to the class definition.
That's not the first time we have done this : https://github.com/clj-commons/aleph/blob/master/src/aleph/netty.clj#L192-L194

src/aleph/http.clj
DerGuteMoritz: I would have liked to do that but turns out that would require a deeper change: THe pool instance we have at this point doesn't expose the options that were passed to it for creation. I feel it's not worth the trouble for this alone. Or am I missing something perhaps?

src/aleph/netty.clj
DerGuteMoritz: Good point, [did that](https://github.com/clj-commons/aleph/pull/715/commits/8ace007d5f0ab24b75838f2ee87178f8335acb8b). I also moved it to the existing section for this but unlike `default-shutdown-timeout`, I opted to *not* mark it as `:no-doc` so that we can still refer to it in the docstrings. I guess this would make sense for `default-shutdown-timeout`, too. WDYT?

DerGuteMoritz:
> I thought for a hot minute the two exceptions would share a useful ancestor class, but nope.

:see_no_evil: 

> Maybe we should handle both exception classes in out `catch`es, and set the netty timeout value from the aleph value.

Good point, that makes users' lives easier and neatly solves the problem that they would sometimes get one or the other when the two timeouts are close to each other :+1: Also went ahead and changed the default `:connection-timeout` from 60s to Netty's default, see https://github.com/clj-commons/aleph/pull/715/commits/b000d983c8977042427322bcc47b10701166c9bd.

BTW notice how we had `(-> ... (d/catch TimeoutException ...) (d/catch ...))` before. Since the first catch returned a `d/error-deferred` again, the second catch would also always be triggered, resulting in duplicate logging of timeout exceptions.

While at it: if you agree with my rationale [here](https://github.com/clj-commons/aleph/pull/714#discussion_r1490788088), how about we turn the `log/error` calls into `log/trace` or completely drop them?

KingMob:
> While at it: if you agree with my rationale https://github.com/clj-commons/aleph/pull/714#discussion_r1490788088, how about we turn the log/error calls into log/trace or completely drop them?

There's no universal way to please everyone with logging. But I tend to view it as a signal to the reader as to what to do. TRACE is likely to be invisible by default. ERROR indicates "do something", and is maybe too strong for something partially outside the operator's control, like timeouts. INFO/DEBUG don't necessarily indicate a problem at all. Perhaps WARNING is the best.

src/aleph/netty.clj
arnaudgeiser: Since Aleph is not strongly encapsulating its internals, I tend to think : "If it's not documented, it's not supposed to be used".
If we remove the `:no-doc`,  we basically mention its part of the Aleph API.

I would say we can remove both the `no-doc`and we can commit on it.

src/aleph/http.clj
arnaudgeiser: Fine for me!

arnaudgeiser (approved):

src/aleph/netty.clj
DerGuteMoritz: Makes sense, pushed a commit to that effect :+1:

DerGuteMoritz:
FTR: I went ahead and demoted the error logs as suggested above now.

arnaudgeiser:
It makes sense, thank you @DerGuteMoritz!

PR #718, opened 2024-04-05
Update dependencies to latest versions

DerGuteMoritz:
Most importantly, update Netty to 4.1.108.Final which includes the fix for https://github.com/netty/netty/issues/13849 as well as various other fixes, particularly HTTP/2 related ones.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,49 +1,49 @@
 ;; DO NOT EDIT MANUALLY - generated from project.clj via deps/lein-to-deps
 {:paths ["src" "resources" "target/classes"],
  :deps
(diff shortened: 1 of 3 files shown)


arnaudgeiser (approved):

bitti:
Thanks for updating this, since we're currently getting flagged for [CVE-2024-29025](https://nvd.nist.gov/vuln/detail/CVE-2024-29025) in our builds. Can a new release be cut out so we don't have to update this transitive dependency ourselves in every project where we're using Aleph?

We're also eager to try out the effects #723, although it's unrelated to the CVE.

PR #719, opened 2024-04-05
Demote insecure/cleartext HTTP pipeline warn log

DerGuteMoritz:
In the common scenario of a plain HTTP server sitting behind a TLS terminating proxy, these warnings would show up for every newly established connection. In any case, warning the user about a setting they deliberately chose on every connection is a rather annoying default, so demote to `debug`.

### DIFF
--- a/src/aleph/http/server.clj
+++ b/src/aleph/http/server.clj
@@ -671,12 +671,12 @@
 
             use-h2c?
             (do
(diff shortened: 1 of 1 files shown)


arnaudgeiser (approved):
Yes, I recognize we went too far with this log, at the very least for HTTP1.0.

PR #721, opened 2024-04-05
Support cancellation during client connection establishment

DerGuteMoritz:
With #714 we added support for cancelling in-flight HTTP requests by putting the response deferred into an error state. However, this only worked once the underlying TCP connection was established. With this patch, it is now possible to cancel requests even while the connection is still being established (possible since Netty 4.1.108.Final via https://github.com/netty/netty/pull/13849). This also works for `aleph.tcp/client`.

Diff best viewed [without whitespace changes](files?w=1).

Completes #712.


src/aleph/http.clj
```diff
@@ -100,20 +100,23 @@
...
+                 (if ssl? 443 80))))
+              ssl?
+              (if on-closed
+                (assoc options :on-closed on-closed)
+                options))]
+    (doto (d/chain' conn middleware)
+      (d/catch' (fn [e]
+                  (log/trace e "Terminating creation of HTTP connection")
+                  (d/error! conn e)
+                  (d/error-deferred e))))))
```
KingMob, line 119: Can you explain this to me? It looks like you're using `doto` to return the deferred from `chain'`, but then you're returning an error-deferred from the `catch'` that will go unused, afaict.

If you don't care about the return value, and just want side effects, it may be clearer to replace the `catch'` with `on-realized`

src/aleph/http.clj
```diff
@@ -389,6 +392,9 @@
                                   ;; function.
                                   (reset! dispose-conn! (fn [] (flow/dispose pool k conn)))
 
+                                  ;; allow cancellation during connection establishment
+                                  (d/connect result (first conn))
```
KingMob, line 396: I assume the way this works is that, if result is already set, it will preempt the connection?

Can you expand the comment to explain this better?

src/aleph/http/client.clj
```diff
@@ -821,112 +821,126 @@
...
-                                (d/chain' (rsp-handler
-                                            {:ch                   ch
-                                             :keep-alive?          keep-alive? ; why not keep-alive?'
-                                             :raw-stream?          raw-stream?
-                                             :req                  req
-                                             :response-buffer-size response-buffer-size
-                                             :t0                   t0})))))))))))))
+               :connect-timeout     connect-timeout})
+
+        _ (attach-on-close-handler ch-d on-closed)
```
KingMob, line 826: Minor: Cleaner to use `doto` on `netty/create-client-chan` right above.

src/aleph/netty.clj
```diff
@@ -1529,8 +1537,8 @@
    complete."
   [{:keys [pipeline-builder
            bootstrap-transform
-           ^SocketAddress remote-address
-           ^SocketAddress local-address
```
KingMob, line 1533: ??? What's going on here?

src/aleph/netty.clj
```diff
@@ -1543,32 +1551,39 @@
...
+      (d/catch' (fn [e]
+                  (when-not (.isDone fut)
+                    (log/trace e "Cancelling Bootstrap#connect future")
+                    (when-not (.cancel fut true)
+                      (when-not (.isDone fut)
+                        (log/warn "Transport" transport "does not support cancellation of connection attempts."
+                                  "Instead, you have to wait for the connect timeout to expire for it to be terminated."
+                                  "Its current value is" connect-timeout "ms."
+                                  "It can be set via the `connect-timeout` option."))))
+                  (d/error-deferred e))))))
```
KingMob, line 1586: Same issue with the doto and catch, unless I'm misreading the indentation in github. (Also, I don't think `doto` should be indented like this, the `catch'` should be aligned with `(-> (wrap-future ....`

src/aleph/tcp.clj
```diff
@@ -204,13 +209,16 @@
...
+              {:pipeline-builder    pipeline-builder
+               :bootstrap-transform bootstrap-transform
+               :remote-address      remote-address
+               :local-address       local-address
+               :transport           (netty/determine-transport transport epoll?)
+               :connect-timeout     connect-timeout})]
+    (d/catch' ch-d #(d/error! s %))
+    (d/catch' s (fn [e]
+                  (d/error! ch-d e)
+                  (d/error-deferred e)))
```
KingMob, line 223: Again, that final `(d/error-deferred e)` is disappearing into the ether. Nothing binds the result of `(d/catch' s (fn [e] ...`

KingMob (commented):
I've reviewed half of it. It seems ok, but there's some weird uses of catch that needs some clarification.

src/aleph/http.clj
DerGuteMoritz: Oof, you're right, `on-realized` would make much more sense here to achieve the desired result, will adjust accordingly.

 The `error-deferred` return is in fact a leftover of an older version, thanks for noticing! Will also clean that up.

src/aleph/http.clj
DerGuteMoritz: Indeed, that's how that works, learned that trick from https://github.com/clj-commons/manifold/issues/167#issuecomment-457996116 - and your reaction confirms Oleksii's statement in the comment after that:

> The approach is not that obvious. Even having some experience with the library.

:sweat_smile: There's another occurrence of this idiom further down this function and I already introduced another instance with https://github.com/clj-commons/aleph/pull/714. Perhaps a util function like `preempt` (which would essentially just be an alias for `d/connect`) with a corresponding docstring would do the trick best?

Also, I just noticed that users could cause some havoc by manually resolving a result promise as successful at the moment. I guess we could say this is just improper usage but ideally we'd only propagate errors in this direction :thinking: Another case in point for a util function! Alright, I convinced myself to give this a shot!

DerGuteMoritz: Ah nevermind, this is in fact the only place where I used `d/connect` to propagate an error backwards like that. I will just convert it to  `d/on-realized`, too, which hopefully makes the intent more obvious. But perhaps a util for the `d/on-realized`-based error propagation is still reasonable. Will experiment and report back.

KingMob: I don't think we should worry about users messing with the deferreds they're given. Technically they could do that on _any_ unresolved deferred...but they won't, not if they want their results. :D

KingMob: One thing is still unclear; what is the advantage to using `connect` in this way over using `alt`?

DerGuteMoritz: I suppose you mean instead of `(d/connect conn result)` and then returning `result`, why not return `(d/alt conn result)`  instead? If so, the problem with that is that `d/alt` will return a *new* deferred. Putting e.g. a `d/timeout!` on that one will have no bearing on `conn` then, i.e. putting the returned deferred into an error state will not propagate back  to cancel the connection establishment. Does that make sense?

src/aleph/http/client.clj
DerGuteMoritz: Agreed, pushed a fixup commit to that effect!

src/aleph/http.clj
DerGuteMoritz: OK replaced all instances of this with `(d/on-realized x identity ...)` instead and dropped the pointless `error-deferred`.

src/aleph/netty.clj
DerGuteMoritz: > Same issue with the doto and catch, unless I'm misreading the indentation in github.

See https://github.com/clj-commons/aleph/pull/721#discussion_r1561040043

> Also, I don't think `doto` should be indented like this, the `catch'` should be aligned with `(-> (wrap-future ....`

Hmm this is how [`clojure-mode` indents it by default](https://github.com/clojure-emacs/clojure-mode/blob/af0e518a6b86f2c6f32dfb30b99c067071ed5cd4/clojure-mode.el#L1895) and it also corresponds with formatting in `clojure.core`, see e.g. [here](https://github.com/clojure/clojure/blob/527b330045ef35b47a968d80ed3dc4999cfa2623/src/clj/clojure/core.clj#L8022-L8033) :thinking: This is how all forms with a "head" of sorts are commonly indented. Which editor are you using that does it differently? I'd tend to file an issue with them about it.

src/aleph/tcp.clj
DerGuteMoritz: See https://github.com/clj-commons/aleph/pull/721#discussion_r1561040043 (resolving this one already).

DerGuteMoritz:
> I've reviewed half of it. It seems ok, but there's some weird uses of catch that needs some clarification.

Thank you for your feedback so far! I've addressed it all but am not yet satisfied with the current state. I'll see if I can somehow make the error propagation a bit easier to grok, possibly by extracting a util function as discussed above. Stay tuned.

src/aleph/netty.clj
KingMob: Nm, I'm used to using the arrow fns, which line up the chains and catches. doto is a body macro, and those are technically supposed to be indented only 2 spaces for the body. It just looks out of place, even though it matches the standard style...

src/aleph/http.clj
KingMob: Ahh, right, right, what was I thinking? nm

DerGuteMoritz:
BTW if you're wondering about these weird "fixup!" commit messages, see e.g. [this post for an explanation](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html). I've adopted that workflow now instead of amending commits & force pushing while working on a branch. That should be less confusing during review hopefully :pray:  I'll rebase & autosquash these before merge then!

KingMob:
> BTW if you're wondering about these weird "fixup!" commit messages, see e.g. [this post for an explanation](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html). I've adopted that workflow now instead of amending commits & force pushing while working on a branch. That should be less confusing during review hopefully 🙏 I'll rebase & autosquash these before merge then!

If you like doing that, you'll love Jujutsu. 

Want to squash the current revision into some older one? jj can do this in a single command: `jj squash -r some-older-revision`. It's less likely to have rebase conflicts than git, too.

Want to just _edit_ that older revision without worry? `jj edit -r some-older-revision` will make that revision the current one. Then, all your file changes will immediately update it, and everything downstream will be rebased, all while you type.

Not ready to just edit that older rev in-place, or need to see the diffs of your new change more clearly? `jj new -r some-older-revision` will immediately make a child revision of `some-older-revision`, and make it the current one. When you're done, `jj squash` puts your changes in `some-older-revision` and immediately rebases everything downstream.

Want to insert a new blank revision _between_ a parent and child? `jj new --insert-after parent`

Here's a wild one. Need to work on three branches simultaneously, but don't want to rebase them off each other? `jj new branch1 branch2 branch3` create a 3-way merged commit. You can use `jj squash --into branchX` to pick which branch to place changes into, and you can get rid of the merge commit when you're all done with `jj abandon`.

Notice I mention editing "revisions", not commits. A commit is immutable, right? Well, in jj, a revision is a stable, semantically-related change that has a history of different commits underneath the hood. If you rebase a branch, the commit IDs all change in git, yes? That's true for jj too, but the _revisions_ stay the same no matter what.

Plus, it's all git-compatible, so your coworkers don't even have to know.

src/aleph/util.clj
```diff
@@ -0,0 +1,20 @@
+(ns aleph.util
+  (:require [manifold.deferred :as d]))
+
+(defn on-error
+  [d f]
+  (d/on-realized d identity f))
```
KingMob, line 6: OK for now, but this should be in Manifold

src/aleph/util.clj
```diff
@@ -0,0 +1,20 @@
...
+  If the error was propagated (i.e. destination wasn't yet realized), on-propagate is invoked with
+  the error value.
+
+  Returns source."
+  ([source destination]
+   (propagate-error source destination identity))
+  ([source destination on-propagate]
+   (on-error source (fn [e]
+                      (when (d/error! destination e)
+                        (on-propagate e))))))
```
KingMob, line 20: `source` is stream terminology, not deferred terminology. We should call it something else to avoid confusion. I made that exact mistake on the first read-through.

test/aleph/util_test.clj
```diff
@@ -0,0 +1,45 @@
...
+            [manifold.deferred :as d]))
+
+(deftest test-propagate-error
+  (testing "Happy path"
+    (let [src (d/deferred)
+          dst (d/deferred)
+          prp (promise)]
+      (util/propagate-error src dst (fn [e] (deliver prp e)))
+      (d/error! src :boom)
+      (is (d/realized? dst))
```
KingMob, line 13: Isn't this a race condition? Might need to put them on same executor and thread.

test/aleph/http_test.clj
```diff
@@ -1480,17 +1480,21 @@
...
+      ;; NOTE: The atom indirection here is needed because `with-raw-handler` will run the body
+      ;; twice (for HTTP1 and HTTP2), so we need a new promise for each run.
+      (reset! conn-established (promise))
+      (reset! conn-closed (promise))
       (let [rsp (http-get "/")]
-        (is (= true (deref conn-established 1000 :timeout)))
+        (is (= true (deref @conn-established 1000 :timeout)))
         (http/cancel-request! rsp)
-        (is (= true (deref conn-closed 1000 :timeout)))
+        (is (= true (deref @conn-closed 1000 :timeout)))
```
KingMob, line 1497: `(deref @ ...` looks weird. I suggest switching to `unwrap/unwrap'`.

src/aleph/util.clj
```diff
@@ -0,0 +1,20 @@
+(ns aleph.util
+  (:require [manifold.deferred :as d]))
+
+(defn on-error
```
KingMob, line 4: I think it should be documented why you would use this over chain or on-realized.

KingMob (changes requested):
Looking good, but I think these utility fns should be in Manifold, not aleph.

PR #723, opened 2024-04-06
Make brotli and zstd optional again

DerGuteMoritz:
Since it came up a few times now, I looked into making zstd and brotli optional again. Turns out there was a pretty easy workaround (see second commit). With this branch, the repro from #703 doesn't crash anymore.


src/aleph/http/compression.clj
```diff
@@ -82,6 +77,16 @@
...
 
+;; We can't reference these options classes directly in `choose-codec` since the compiler would try
+;; to initialize them even when the necessary dependencies aren't available, resulting in an
+;; error. Indirectly referencing them like this works, though.  See
+;; https://github.com/clj-commons/aleph/issues/703
+(def brotli-options-class
+  BrotliOptions)
+
+(def zstd-options-class
+  ZstdOptions)
```
KingMob, line 88: I'm surprised this works; the compiler is still looking at the whole file

KingMob (commented):
I wonder if we're just shifting around the set of people who will complain :) 

Though truthfully, based on the number of downloads I see on Clojars, I don't think Aleph is used much anymore, so there probably won't be many complaints.

src/aleph/http/compression.clj
DerGuteMoritz: It seems to be due to how `fn` bodies are compiled. This for example also fails to compile:

```clojure
(def brotli-options-class
  (fn []
    BrotliOptions))
```

Perhaps this is a bug / unnecessary bit the compiler does when it analyzes a function body? I'll prepare a self-contained reproducer and file a question on https://ask.clojure.org/ about this.

DerGuteMoritz:
> I wonder if we're just shifting around the set of people who will complain :)

True. I guess we could have another vote about it but given the meager turnout last time...


> Though truthfully, based on the number of downloads I see on Clojars, I don't think Aleph is used much anymore, so there probably won't be many complaints.

Hmm indeed.... judging from these, at best there are maybe still 1000 users around? Perhaps it's time for a bit of a marketing campaign to re-ignite its popularity :smile:

src/aleph/http/compression.clj
KingMob: Hmmm, feels fragile to me. Like you're relying on internal behavior of the compiler.

The solution I originally contemplated, but figured would be a bunch of maintenance work, was to move all of these references into the Java class. That way the Clojure compiler can't blow up on it.

DerGuteMoritz: https://ask.clojure.org/index.php/13810/compilation-referencing-class-failing-static-initializer :sweat_drops:

arnaudgeiser (approved):

src/aleph/http/compression.clj
DerGuteMoritz: > Hmmm, feels fragile to me. Like you're relying on internal behavior of the compiler.

Yeah, I agree with your feeling. Meanwhile [Alex replied to my question](https://ask.clojure.org/index.php/13810/compilation-referencing-class-failing-static-initializer?show=13811#a13811) and he seems to agree that it's a bug in how `fn` bodies are compiled. Makes me feel a bit easier about the workaround.

 
> The solution I originally contemplated, but figured would be a bunch of maintenance work, was to move all of these references into the Java class. That way the Clojure compiler can't blow up on it.

Perhaps this is indeed the better way to go. I'll give that a shot to see!

DerGuteMoritz: OK pushed! That `instanceOf` issue mentioned in the comment derailed me for a bit there :sweat_smile: But overall this feels more robust. WDYT?

KingMob (approved):
Looks good. I think it's much safer to move this to Java.

Issue #727, opened 2024-06-10
No ssl-context via `SslContext` in Aleph 0.7.0 and later


David-Ongaro:
When trying to upgrade one of our services from Aleph 0.6.4 to 0.8.0 I saw this strange exception in the logs:

```
[...]
Caused by: clojure.lang.ExceptionInfo: Some desired HTTP versions are not part of ALPN config.
{:desired-http-versions [:http1], :alpn-protocols []}
 at aleph.http.common$assert_consistent_alpn_config_BANG_.invokeStatic (common.clj:139)
    aleph.http.common$assert_consistent_alpn_config_BANG_.invoke (common.clj:130)
    aleph.http.common$ensure_consistent_alpn_config.invokeStatic (common.clj:182)
    aleph.http.common$ensure_consistent_alpn_config.invoke (common.clj:143)
    aleph.http.client$client_ssl_context.invokeStatic (client.clj:698)
    aleph.http.client$client_ssl_context.invoke (client.clj:690)
    aleph.http.client$http_connection.invokeStatic (client.clj:799)
    aleph.http.client$http_connection.invoke (client.clj:752)
    aleph.http$create_connection.invokeStatic (http.clj:104)
    aleph.http$create_connection.invoke (http.clj:97)
    aleph.http$connection_pool$fn__21882.invoke (http.clj:239)
    aleph.flow$instrumented_pool$reify__12626.generate (flow.clj:47)
    io.aleph.dirigiste.Pool.addObject (Pool.java:273)
    io.aleph.dirigiste.Pool.acquire (Pool.java:466)
    aleph.flow$acquire$fn__12631.invoke (flow.clj:74)
    aleph.flow$acquire.invokeStatic (flow.clj:73)
    aleph.flow$acquire.invoke (flow.clj:68)
    aleph.http$eval21907$request__21911$fn__21915$fn__21916.invoke (http.clj:377)
    aleph.http$eval21907$request__21911$fn__21915.invoke (http.clj:371)
    aleph.http$eval21907$request__21911.invoke (http.clj:370)
[...]
```

When trying to pin this down, I could confirm that this already happens in Aleph 0.7.0 (even for the alpha versions) and it only happens when the pool is configured with an `io.netty.handler.ssl.SslContext` instance. I.e.

```clojure
(def ssl-context (.build (io.netty.handler.ssl.SslContextBuilder/forClient)))

(def pool (http/connection-pool {:connection-options {:ssl-context ssl-context}}))

[...]

PR #729, opened 2024-06-13
Improve docs for `http/connection-pool`'s HTTP/2 support and ALPN config

DerGuteMoritz:
Prompted by #727


src/aleph/http.clj
```diff
@@ -150,7 +157,7 @@
 
    Param key                   | Description
    | ---                       | ---
-   | `ssl-context`             | an `io.netty.handler.ssl.SslContext` object or a map of SSL context options (see `aleph.netty/ssl-client-context` for more details), only required if a custom context is required
+   | `ssl-context`             | an `io.netty.handler.ssl.SslContext` object or a map of SSL context options (see `aleph.netty/ssl-client-context` for more details), only required if a custom context is desired. When passing an `io.netty.handler.ssl.SslContext` object, it must have an ALPN config matching the `http-versions` option (see also `aleph.netty/application-protocol-config`).
```
arnaudgeiser, line 160: Can we create an example for this?
Because, as an end user, that's not something I'm expecting to have to do...

DerGuteMoritz: Why not! Where would such an example go so that users will find it, though? :thinking:

arnaudgeiser: Somewhere in this namespace? [1]
To be completely transparent with you, I'm not even sure what this comment is asking me to do.

[1] : https://github.com/clj-commons/aleph/blob/master/examples/src/aleph/examples/http.clj

KingMob:
Are you sure you want to mandate ALPN and update the docs, instead of updating the code?

arnaudgeiser:
> Are you sure you want to mandate ALPN and update the docs, instead of updating the code?

I think we should close the documentation PR and aims at updating the code with the snippet above.

DerGuteMoritz:
> Are you sure you want to mandate ALPN and update the docs, instead of updating the code?

No I would prefer to update the code, as well.  Will prepare a patch for that separately and once it's landed, we can get back to the doc update here.

> I think we should close the documentation PR and aim at updating the code with the snippet above.

The doc update is mostly orthogonal, only the wording about the ALPN config requirement would have to be changed once we've made it (more) optional again.

src/aleph/http.clj
DerGuteMoritz: Ah nice, that fits the bill! And in fact, the example there is broken due to the same issue at the moment :sweat_smile:  We should probably also re-test all the examples as part of the release preparation process...

> To be completely transparent with you, I'm not even sure what this comment is asking me to do.

Hmm well I've been assuming that a custom `SslContext` will only be provided by advanced users who should be aware of ALPN and how it relates to HTTP/2. It should be very rarely necessary anyway since we expose most (if not all?) SSL context settings via the options map, in which case you only need to specify which HTTP versions you care about (via the `http-versions` option) and Aleph takes care of ALPN for you then... So I would tend to leave it as-is and perhaps have the comment point to the example instead of `aleph.netty/application-protocol-config`? Thoughts?

src/aleph/netty.clj
```diff
@@ -938,7 +938,16 @@
...
+       Takes a vector of HTTP versions, in order of preference. E.g., `[:http2 :http1]`
+
+       Note that the returned config uses `SelectorFailureBehavior.NO_ADVERTISE`[1] and
+       `SelectedListenerFailureBehavior.ACCEPT`[2] since these are the only failure behaviors
+       supported by all SSL providers. See their documentation for details. One important
+       consequence of this is that it's not possible to completely opt out of HTTP/1.1 by way of
+       only specifying `[:http2]`.
+
+       1: https://netty.io/4.1/api/io/netty/handler/ssl/ApplicationProtocolConfig.SelectorFailureBehavior.html#NO_ADVERTISE
+       2: https://netty.io/4.1/api/io/netty/handler/ssl/ApplicationProtocolConfig.SelectedListenerFailureBehavior.html#ACCEPT"
```
KingMob, line 950: Maybe add a line to the effect of: "If you want this behavior, you will need to supply an SslContext with a different config."

src/aleph/http.clj
DerGuteMoritz: Heh with https://github.com/clj-commons/aleph/pull/730, the example *does* work without changes again, so it can't serve as an example anymore. [`aleph.examples.http2/http2-ssl-ctx`](https://github.com/clj-commons/aleph/blob/5caa1311db08c3f94c87871f2a57cb54e5987cdd/examples/src/aleph/examples/http2.clj#L33)  *could* work but it's a server SSL context so it wouldn't work in this particular case... I think I'll leave it as-is, advanced users should be able to figure out what to do based on the current docstrings. Otherwise we'll probably hear about it!

src/aleph/netty.clj
DerGuteMoritz: Done via 439214b -- I made it about `ApplicationProtocolConfig` instead since it can also be used with an `ssl-context` options map, custom `SslContext`  isn't even required.

src/aleph/http.clj
```diff
@@ -52,7 +52,7 @@
    | `socket-address`                  | A `java.net.SocketAddress` specifying both the port and interface to bind to. |
    | `bootstrap-transform`             | A function that takes an `io.netty.bootstrap.ServerBootstrap` object, which represents the server, and modifies it. |
    | `http-versions`                   | An optional vector of allowable HTTP versions to negotiate via ALPN, in preference order. Defaults to `[:http1]`. |
-   | `ssl-context`                     | An `io.netty.handler.ssl.SslContext` object or a map of SSL context options (see `aleph.netty/ssl-server-context` for more details) if an SSL connection is desired |
+   | `ssl-context`                     | An `io.netty.handler.ssl.SslContext` object or a map of SSL context options (see `aleph.netty/ssl-server-context` for more details) if an SSL connection is desired. When passing an `io.netty.handler.ssl.SslContext` object, it must have an ALPN config matching the `http-versions` option (see `aleph.netty/ssl-server-context` and `aleph.netty/application-protocol-config`).
```
KingMob, line 55: Hmmm... Maybe mention that no ALPN is needed for HTTP1?

How about "... have an ALPN config (if required) matching the ..." ?

KingMob (approved):
Overall LGTM, but see my minor doc suggestion

src/aleph/http.clj
DerGuteMoritz: Went with adding "If only HTTP/1.1 is desired, ALPN config is optional."

PR #730, opened 2024-06-13
Allow custom `SslContext` without ALPN again for HTTP/1.1-only case

### LINKED ISSUE #727: No ssl-context via `SslContext` in Aleph 0.7.0 and later

DerGuteMoritz:
With the introduction of HTTP/2 support, Aleph started to require a matching ALPN config to be present in custom `SslContext` objects. This needlessly broke existing HTTP/1.1-only uses. With this change, we now allow custom `SslContext` objects without ALPN config again if only HTTP/1.1 is desired (via the `http-versions` option). Since this still happens to be the default, existing uses should just work again.

Fixes #727

----

@KingMob your suggested patch *almost* worked as-is but was thwarted by the fact that `applicationProtocolNegotiator` is always present (see code comment).

Review welcome but don't merge, yet, as I also would like to add a `test-http-versions-client-config` test which mirrors `test-http-versions-server-config` for the client (as that's what #727 was originally about).

### DIFF
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -1339,7 +1339,7 @@
     (catch Exception e
       e)))
 
-(deftest test-http-versions-config
+(deftest test-http-versions-server-config
   (testing "ssl-context as options map"
 
     (testing "with different HTTP versions in ALPN config"
@@ -1424,12 +1424,18 @@
                                           (netty/application-protocol-config [:http2 :http1])))})]
         (is (= :started result))))
 
-    (testing "with no ALPN config"
+    (testing "with no ALPN config but desiring HTTP/2"
       (let [result (try-start-server
                     {:http-versions [:http2 :http1]
                      :ssl-context test-ssl/server-ssl-context})]
         (is (instance? ExceptionInfo result))
-        (is (= "Some desired HTTP versions are not part of ALPN config." (ex-message result))))))
+        (is (= "No ALPN supplied, but requested non-HTTP/1 versions that require ALPN." (ex-message result)))))
+
+    (testing "with no ALPN config but desiring only HTTP/1"
+      (let [result (try-start-server
+                    {:http-versions [:http1]
+                     :ssl-context test-ssl/server-ssl-context})]
+        (is (= :started result)))))
 
   (testing "HTTP/2 without ssl-context"
     (let [result (try-start-server
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -1429,7 +1429,7 @@
                     {:http-versions [:http2 :http1]
                      :ssl-context test-ssl/server-ssl-context})]
         (is (instance? ExceptionInfo result))
-        (is (= "No ALPN supplied, but requested non-HTTP/1 versions that require ALPN." (ex-message result)))))
+        (is (= "Supplied SslContext with no ALPN config, but requested secure non-HTTP/1 versions that require ALPN." (ex-message result)))))
 
     (testing "with no ALPN config but desiring only HTTP/1"
       (let [result (try-start-server
@@ -1583,7 +1583,7 @@
                        {:http-versions [:http2 :http1]
                         :ssl-context test-ssl/client-ssl-context}})]
           (is (instance? ExceptionInfo result))
-          (is (= "No ALPN supplied, but requested non-HTTP/1 versions that require ALPN." (ex-message result)))))
+          (is (= "Supplied SslContext with no ALPN config, but requested secure non-HTTP/1 versions that require ALPN." (ex-message result)))))
 
       (testing "with no ALPN config but desiring only HTTP/1"
         (let [result (try-request-with-pool
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -1464,6 +1464,154 @@
       (is (instance? IllegalArgumentException result))
       (is (= "use-h2c? may only be true when HTTP/2 is enabled." (ex-message result))))))
 
+(defn ex-root-cause [e]
+  (if-let [c (ex-cause e)]
+    (recur c)
+    e))
+
+(defn try-request-with-pool [options]
+  (try
+    (let [^Pool pool (http/connection-pool options)]
+      (try
+        @(http-get "/" {:pool pool})
+        :success
+        (finally
+          (.shutdown pool))))
+    (catch Exception e
+      ;; Exceptions raised during connection creation are wrapped in multiple levels of
+      ;; `RuntimeException` which we don't care about here.
+      (ex-root-cause e))))
+
+(deftest test-http-versions-client-config
+  (with-http-ssl-servers echo-handler {}
+    (testing "ssl-context as options map"
+
+      (testing "with different HTTP versions in ALPN config"
+        (let [result (try-request-with-pool
+                      {:connection-options
+                       {:http-versions [:http2]
+                        :ssl-context (assoc test-ssl/client-ssl-context-opts
+                                            :application-protocol-config
+                                            (netty/application-protocol-config [:http1]))}})]
+          (is (instance? ExceptionInfo result))
+          (is (= "Some desired HTTP versions are not part of ALPN config." (ex-message result)))))
+
+      (testing "with different preference order in ALPN config"
+        (let [result (try-request-with-pool
+                      {:connection-options
+                       {:http-versions [:http2 :http1]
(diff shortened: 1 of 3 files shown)


KingMob (approved):
👍🏻

arnaudgeiser (approved):

test/aleph/http_test.clj
```diff
@@ -1464,6 +1464,162 @@
...
+                                             (netty/application-protocol-config [:http2 :http1])))}})]
+          (is (= :success result))))
+
+      (testing "with no ALPN config but desiring HTTP/2"
+        (let [result (try-request-with-pool
+                      {:connection-options
+                       {:http-versions [:http2 :http1]
+                        :ssl-context test-ssl/client-ssl-context}})]
+          (is (instance? ExceptionInfo result))
+          (is (= "No ALPN supplied, but requested non-HTTP/1 versions that require ALPN." (ex-message result)))))
```
KingMob, line 1586: We should probably specify in the error message that they requested _secure_ HTTP2. 

On the rare chance they're using h2c, there will be no ALPN (or TLS, for that matter).

test/aleph/http_test.clj
```diff
@@ -1464,6 +1464,162 @@
...
+          (is (= :success result)))))
+
+    (testing "HTTP/2 without ssl-context"
+      (let [result (try-request-with-pool
+                    {:connection-options
+                     {:http-versions [:http2]
+                      ;; We only want assert that not providing `:ssl-context` is possible but since
+                      ;; the test server uses a self-signed certificate, we cannot verify it with
+                      ;; the default one. Since all we care about is ALPN here, this is fine.
+                      :insecure? true}})]
```
KingMob, line 1602: I hate this key, because it's so confusing.

`insecure?` here refers to using an insecure self-signed certificate, NOT using http or h2c.

KingMob (approved):

test/aleph/http_test.clj
DerGuteMoritz: Good idea. While at it, I also made it explicit that this is about the supplied `SslContext`.

test/aleph/http_test.clj
DerGuteMoritz: > refers to using an insecure self-signed certificate

Nit: Not just (any) self-signed certificates but also invalid certificates (e.g. expired ones or for different domains).

PR #733, opened 2024-06-18
Demote HTTPS connection info log to debug

DerGuteMoritz:
Logging every established HTTPS connection along with its ALPN protocols is unnecessarily verbose without adding much value (there's no way to identify the connection, for example), so demote these to debug level. Users can still get higher-value logs, e.g. via ring middleware.

Follow-up to #719.


arnaudgeiser (approved):
Yes, definitely.

PR #734, opened 2024-06-18
Update Netty to 4.1.111.Final

DerGuteMoritz:
This is a pure bug-fix release, so shouldn't hurt to include it.


arnaudgeiser (approved):

deps.edn
```diff
@@ -10,24 +10,24 @@
   org.clj-commons/dirigiste {:mvn/version "1.0.4"},
```
arnaudgeiser, line 10: That good old `deps.edn` that needs to be updated.

DerGuteMoritz: Yup, already done - the CI job reminded me :smile:

PR #741, opened 2024-12-13
test-bulk-requests: use dedicated pool

arnaudgeiser:
Use a dedicated connection pool instead of the shared one that could be closed by a concurrent test.

### DIFF
--- a/test/aleph/http_test.clj
+++ b/test/aleph/http_test.clj
@@ -523,16 +523,17 @@
     (repeatedly #(rand-nth charset))))
 
 (deftest test-bulk-requests
-  (with-handler basic-handler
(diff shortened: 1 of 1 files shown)


KingMob:
If something is closing the default pool, is there a problem elsewhere?

arnaudgeiser:
> If something is closing the default pool, is there a problem elsewhere?

We have various kind of tests, including some that are testing some edge cases.
If any problem, this is not a new issue that we introduced. Overall, it makes sense to have all the client tests run in isolation instead of having a single client (pool) that will be reused across all the tests.
Before investigating, I would think about revisiting how we bootstrap those clients for test execution.

PR #742, opened 2025-01-23
Fix HTTP client proxy support

### LINKED ISSUE #731: Proxy functionality broken in Aleph 0.7.0 and later
Since updating from 0.6.4 to 0.8.0 I see strange exceptions for proxied calls.

```
io.netty.handler.codec.DecoderException: io.netty.handler.ssl.NotSslRecordException: not an SSL/TLS record: 485454502f312e31203430302042616420526571756573740d0a636f6e74656e742d6c656e6774683a2031310d0a636f6e74656e742d747970653a20746578742f706c61696e0d0a646174653a205468752c203133204a756e20323032342032323a35353a323720474d540d0a7365727665723a20656e766f790d0a636f6e6e656374696f6e3a20636c6f73650d0a0d0a4261642052657175657374
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:500)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:290)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1407)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:440)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:918)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:788)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:724)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:650)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:562)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:994)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at manifold.executor$thread_factory$reify__1227$f__1228.invoke(executor.clj:71)
	at clojure.lang.AFn.run(AFn.java:22)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: io.netty.handler.ssl.NotSslRecordException: not an SSL/TLS record: 485454502f312e31203430302042616420526571756573740d0a636f6e74656e742d6c656e6774683a2031310d0a636f6e74656e742d747970653a20746578742f706c61696e0d0a646174653a205468752c203133204a756e20323032342032323a35353a323720474d540d0a7365727665723a20656e766f790d0a636f6e6e656374696f6e3a20636c6f73650d0a0d0a4261642052657175657374
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1314)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1387)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:530)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:469)
	... 19 common frames omitted
```

Decoding the hex string yields:

```console
$ xxd -r -p <<<485454502f312e31203430302042616420526571756573740d0a636f6e74656e742d6c656e6774683a2031310d0a636f6e74656e742d747970653a20746578742f706c61696e0d0a646174653a205468752c203133204a756e20323032342032323a35353a323720474d540d0a7365727665723a20656e766f790d0a636f6e6e656374696f6e3a20636c6f73650d0a0d0a4261642052657175657374
HTTP/1.1 400 Bad Request
content-length: 11
content-type: text/plain
date: Thu, 13 Jun 2024 22:55:27 GMT
server: envoy
connection: close

Bad Request
```
So it seems it tries to interpret an http response as https. (I try to reach a https endpoint via an http proxy, but maybe it's making an unproxied http call instead while still expecting https.)

It's difficult for me, though, to reproduce the proxy settings we use in prod locally. Therefore, I will use a local proxy for this bug report here:

```console
$ docker run --rm -p 3128:3128/tcp ghcr.io/tarampampam/3proxy
/bin/3proxy: Starting 3proxy
{"time_unix":1718326196, "proxy":{"type:":"PROXY", "port":3128}, "error":{"code":"00000"}, "auth":{"user":"-"}, "client":{"ip":"0.0.0.0", "port":3128}, "server":{"ip":"0.0.0.0", "port":0}, "bytes":{"sent":0, "received":0}, "request":{"hostname":"[0.0.0.0]"}, "message":"Accepting connections [17/4281620160]"}
{"time_unix":1718326196, "proxy":{"type:":"SOCKS", "port":1080}, "error":{"code":"00000"}, "auth":{"user":"-"}, "client":{"ip":"0.0.0.0", "port":1080}, "server":{"ip":"0.0.0.0", "port":0}, "bytes":{"sent":0, "received":0}, "request":{"hostname":"[0.0.0.0]"}, "message":"Accepting connections [17/4281566912]"}
```

After this, we can confirm that Aleph 0.6.4 works as expected

```clojure
(def pool (http/connection-pool {:connection-options {:proxy-options {:host "localhost" :port 3128}}}))

@(http/get "http://example.com" {:pool pool}) =>
{:request-time 52,
 :aleph/keep-alive? true,
 :headers
 {"x-cache" "HIT",
  "server" "ECAcc (sed/58AF)",
  "age" "245573",
  "content-type" "text/html; charset=UTF-8",
  "content-length" "1256",
  "accept-ranges" "bytes",
  "expires" "Fri, 21 Jun 2024 00:16:36 GMT",
  "etag" "\"3147526947+gzip\"",
  "date" "Fri, 14 Jun 2024 00:16:36 GMT",
  "vary" "Accept-Encoding",
  "last-modified" "Thu, 17 Oct 2019 07:18:26 GMT",
  "cache-control" "max-age=604800"},
 :status 200,
 :connection-time 96,
 :body
 #object[java.io.ByteArrayInputStream 0x36c1143f "java.io.ByteArrayInputStream@36c1143f"]}

@(http/get "https://example.com" {:pool pool}) =>
{:request-time 21,
 :aleph/keep-alive? true,
 :headers
 {"x-cache" "HIT",
  "server" "ECAcc (sed/5885)",
  "age" "271652",
  "content-type" "text/html; charset=UTF-8",
  "content-length" "1256",
  "expires" "Fri, 21 Jun 2024 00:16:40 GMT",
  "etag" "\"3147526947+gzip+ident\"",
  "date" "Fri, 14 Jun 2024 00:16:40 GMT",
  "vary" "Accept-Encoding",
  "last-modified" "Thu, 17 Oct 2019 07:18:26 GMT",
  "cache-control" "max-age=604800"},
 :status 200,
 :connection-time 194,
 :body
 #object[java.io.ByteArrayInputStream 0x2f625fd5 "java.io.ByteArrayInputStream@2f625fd5"]}
```

Trying the same in 0.7.0 or higher yields

```clojure
@(http/get "https://example.com" {:pool pool}) =>
Execution error (ExceptionInfo) at aleph.http.client/http-connection$setup-client$http-req-fn (client.clj:913).
Channel is inactive/closed.
   {:req
    {:path "/",
     :user-info nil,
     :pool
     #object[io.aleph.dirigiste.Pool 0x124e951f "io.aleph.dirigiste.Pool@124e951f"],
     :request-url "https://example.com",
     :server-port nil,
     :uri "/",
     :server-name "example.com",
     :query-string nil,
     :scheme :https,
     :request-method :get},
    :ch NioSocketChannel [8efb471b]():
   	Handlers:
   ,
    :open? false,
    :active? false}
                client.clj:  913  aleph.http.client/http-connection/setup-client/http-req-fn
     client_middleware.clj:  263  aleph.http.client-middleware/wrap-exceptions/fn/fn
                  AFn.java:  154  clojure.lang.AFn/applyToHelper
                  AFn.java:  144  clojure.lang.AFn/applyTo
                  core.clj:  667  clojure.core/apply
                  core.clj: 1990  clojure.core/with-bindings*
                  core.clj: 1990  clojure.core/with-bindings*
               RestFn.java:  142  clojure.lang.RestFn/applyTo
                  core.clj:  671  clojure.core/apply
                  core.clj: 2020  clojure.core/bound-fn*/fn
               RestFn.java:  408  clojure.lang.RestFn/invoke
              deferred.clj:  882  manifold.deferred/eval17318/chain'-
     client_middleware.clj:  263  aleph.http.client-middleware/wrap-exceptions/fn
     client_middleware.clj:  598  aleph.http.client-middleware/wrap-request-timing/fn
     client_middleware.clj:  976  aleph.http.client-middleware/wrap-request/fn
                  http.clj:  403  aleph.http/eval28275/request/fn/fn/fn
              deferred.clj:  900  manifold.deferred/eval17318/chain'-
              deferred.clj:  854  manifold.deferred/eval17318/subscribe/fn
              deferred.clj:  344  manifold.deferred.Listener/onSuccess
              deferred.clj:  523  manifold.deferred.Deferred/fn/fn
                  AFn.java:   22  clojure.lang.AFn/run
             Executor.java:  320  io.aleph.dirigiste.Executor$3/run
             Executor.java:   62  io.aleph.dirigiste.Executor$Worker$1/run
              executor.clj:   70  manifold.executor/thread-factory/reify/f
```

Or alternatively, after a timeout of a minute or so:

```clojure
@(http/get "https://example.com" {:pool pool}) =>

Unhandled io.netty.handler.proxy.ProxyConnectException
   http, none, localhost/127.0.0.1:3128 => null, timeout
         ProxyHandler.java:  201  io.netty.handler.proxy.ProxyHandler$2/run
          PromiseTask.java:   98  io.netty.util.concurrent.PromiseTask/runTask
  ScheduledFutureTask.java:  153  io.netty.util.concurrent.ScheduledFutureTask/run
AbstractEventExecutor.java:  173  io.netty.util.concurrent.AbstractEventExecutor/runTask
AbstractEventExecutor.java:  166  io.netty.util.concurrent.AbstractEventExecutor/safeExecute
SingleThreadEventExecutor.java:  469  io.netty.util.concurrent.SingleThreadEventExecutor/runAllTasks
         NioEventLoop.java:  569  io.netty.channel.nio.NioEventLoop/run
SingleThreadEventExecutor.java:  994  io.netty.util.concurrent.SingleThreadEventExecutor$4/run
    ThreadExecutorMap.java:   74  io.netty.util.internal.ThreadExecutorMap$2/run
              executor.clj:   70  manifold.executor/thread-factory/reify/f
                  AFn.java:   22  clojure.lang.AFn/run
FastThreadLocalRunnable.java:   30  io.netty.util.concurrent.FastThreadLocalRunnable/run
               Thread.java: 1583  java.lang.Thread/run
```

Trying http yields a strange 503:

```clojure
@(http/get "http://example.com" {:pool pool})
Execution error (ExceptionInfo) at aleph.http.client-middleware/wrap-exceptions$fn$fn$fn (client_middleware.clj:275).
status: 503
   {:aleph/keep-alive? false,
    :aleph/destroy-conn? #<SuccessDeferred@5a793212: false>,
    :headers
    {"content-type" "text/html; charset=UTF-8",
     "content-length" "1284",
     "connection" "close",
     "pragma" "no-cache",
     "expires" "Thu, 01 Jan 1970 00:00:00 GMT",
     "p3p" "CP=\"CAO PSA OUR\"",
     "cache-control"
     "no-store, no-cache, must-revalidate, post-check=0, pre-check=0"},
    :status 503,
    :body
    #object[java.io.ByteArrayInputStream 0x77490577 "java.io.ByteArrayInputStream@77490577"]}
     client_middleware.clj:  275  aleph.http.client-middleware/wrap-exceptions/fn/fn/fn
              deferred.clj:  882  manifold.deferred/eval17318/chain'-
              deferred.clj:  848  manifold.deferred/eval17318/subscribe/fn
              deferred.clj:  344  manifold.deferred.Listener/onSuccess
              deferred.clj:  523  manifold.deferred.Deferred/fn
              deferred.clj:  523  manifold.deferred.Deferred/success
              deferred.clj:  368  manifold.deferred/success!
              deferred.clj:  365  manifold.deferred/success!
     client_middleware.clj:  272  aleph.http.client-middleware/wrap-exceptions/fn/fn/f--auto--
                  AFn.java:   22  clojure.lang.AFn/run
             Executor.java:  320  io.aleph.dirigiste.Executor$3/run
             Executor.java:   62  io.aleph.dirigiste.Executor$Worker$1/run
              executor.clj:   70  manifold.executor/thread-factory/reify/f
```

If I try https afterwards it seems it still tries to intepret the result as http:

```clojure
@(http/get "https://example.com" {:pool pool}) =>
Execution error (IllegalArgumentException) at io.netty.handler.codec.http.HttpVersion/<init> (HttpVersion.java:123).
invalid version format: Ú¶&¿IÐQ¡_QIQ©ÎÜÕCÛDÁÎ6Ó§L³K·¹}§¾¿¶.¯}¦KSSÇ¶'}ÇÊÂÍ0I\RÊJ:ÑÚ­/CYÌ±Þ
          HttpVersion.java:  123  io.netty.handler.codec.http.HttpVersion/<init>
          HttpVersion.java:   85  io.netty.handler.codec.http.HttpVersion/valueOf
  HttpResponseDecoder.java:  194  io.netty.handler.codec.http.HttpResponseDecoder/createMessage
    HttpObjectDecoder.java:  354  io.netty.handler.codec.http.HttpObjectDecoder/decode
      HttpClientCodec.java:  320  io.netty.handler.codec.http.HttpClientCodec$Decoder/decode
 ByteToMessageDecoder.java:  530  io.netty.handler.codec.ByteToMessageDecoder/decodeRemovalReentryProtection
 ByteToMessageDecoder.java:  469  io.netty.handler.codec.ByteToMessageDecoder/callDecode
 ByteToMessageDecoder.java:  290  io.netty.handler.codec.ByteToMessageDecoder/channelRead
CombinedChannelDuplexHandler.java:  251  io.netty.channel.CombinedChannelDuplexHandler/channelRead
     HttpProxyHandler.java:  284  io.netty.handler.proxy.HttpProxyHandler$HttpClientCodecWrapper/channelRead
AbstractChannelHandlerContext.java:  444  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
AbstractChannelHandlerContext.java:  420  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
AbstractChannelHandlerContext.java:  412  io.netty.channel.AbstractChannelHandlerContext/fireChannelRead
       LoggingHandler.java:  280  io.netty.handler.logging.LoggingHandler/channelRead
AbstractChannelHandlerContext.java:  442  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
AbstractChannelHandlerContext.java:  420  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
AbstractChannelHandlerContext.java:  412  io.netty.channel.AbstractChannelHandlerContext/fireChannelRead
DefaultChannelPipeline.java: 1407  io.netty.channel.DefaultChannelPipeline$HeadContext/channelRead
AbstractChannelHandlerContext.java:  440  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
AbstractChannelHandlerContext.java:  420  io.netty.channel.AbstractChannelHandlerContext/invokeChannelRead
DefaultChannelPipeline.java:  918  io.netty.channel.DefaultChannelPipeline/fireChannelRead
AbstractNioByteChannel.java:  166  io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe/read
         NioEventLoop.java:  788  io.netty.channel.nio.NioEventLoop/processSelectedKey
         NioEventLoop.java:  724  io.netty.channel.nio.NioEventLoop/processSelectedKeysOptimized
         NioEventLoop.java:  650  io.netty.channel.nio.NioEventLoop/processSelectedKeys
         NioEventLoop.java:  562  io.netty.channel.nio.NioEventLoop/run
SingleThreadEventExecutor.java:  994  io.netty.util.concurrent.SingleThreadEventExecutor$4/run
    ThreadExecutorMap.java:   74  io.netty.util.internal.ThreadExecutorMap$2/run
              executor.clj:   70  manifold.executor/thread-factory/reify/f
                  AFn.java:   22  clojure.lang.AFn/run
FastThreadLocalRunnable.java:   30  io.netty.util.concurrent.FastThreadLocalRunnable/run
               Thread.java: 1583  java.lang.Thread/run
```

although, I sometimes can get this error right away, without trying http first, so it doesn't seem to be a matter of misusing a pool for different protocols (which we don't do in prod anyway).

So even though I can not reproduce the exact error I see in prod locally, I think we can conclude that there is a regression in 0.7.0 and later compared to 0.6.4.

DerGuteMoritz:
Fixes #731

### DIFF
--- a/src/aleph/http/client.clj
+++ b/src/aleph/http/client.clj
@@ -378,7 +378,7 @@
       (.setConnectTimeoutMillis ^ProxyHandler handler connection-timeout))
     handler))
 
-(defn pending-proxy-connection-handler [response-stream]
+(defn pending-proxy-connection-handler [early-response-d]
   (netty/channel-inbound-handler
     :exception-caught
     ([_ ctx cause]
@@ -396,7 +396,7 @@
 
                         :else
                         cause)]
-         (s/put! response-stream response)
+         (d/error! early-response-d response)
          ;; client handler should take care of the rest
          (netty/close ctx))))
 
@@ -408,7 +408,7 @@
 
 (defn- add-proxy-handlers
   "Inserts handlers for proxying through a server"
-  [^ChannelPipeline p response-stream proxy-options ssl?]
+  [^ChannelPipeline p early-response-d proxy-options ssl?]
   (when (some? proxy-options)
     (let [proxy (proxy-handler (assoc proxy-options :ssl? ssl?))]
       (.addFirst p "proxy" ^ChannelHandler proxy)
@@ -420,7 +420,7 @@
                    "proxy"
                    "pending-proxy-connection"
                    ^ChannelHandler
-                   (pending-proxy-connection-handler response-stream)))))
+                   (pending-proxy-connection-handler early-response-d)))))
   p)
 
 (defn- setup-http1-pipeline
@@ -458,7 +458,6 @@
                     false))
         (.addLast "streamer" ^ChannelHandler (ChunkedWriteHandler.))
         (.addLast "handler" ^ChannelHandler handler)
-        (add-proxy-handlers responses proxy-options ssl?)
         (common/add-non-http-handlers logger pipeline-transform))
 
     (log/debug "http1 client pipeline" pipeline)
@@ -473,9 +472,10 @@
 
    Can't use an ApnHandler/ApplicationProtocolNegotiationHandler here,
    because it's tricky to run Manifold code on Netty threads."
-  [{:keys [ssl? remote-address ssl-context ssl-endpoint-id-alg]}]
+  [{:keys [ssl? remote-address ssl-context ssl-endpoint-id-alg proxy-options early-response-d]}]
   (fn pipeline-builder*
     [^ChannelPipeline pipeline]
+    (add-proxy-handlers pipeline early-response-d proxy-options ssl?)
     (when ssl?
       (do
         (.addLast pipeline
@@ -803,16 +803,18 @@
                  (some? log-activity) (netty/activity-logger "aleph-client" log-activity)
                  :else nil)
 
+        early-response-d (d/deferred)
         pipeline-builder (make-pipeline-builder
-                           (assoc opts
-                                  :ssl? ssl?
-                                  :ssl-context ssl-context
-                                  :ssl-endpoint-id-alg ssl-endpoint-id-alg
-                                  :remote-address remote-address
-                                  :raw-stream? raw-stream?
-                                  :response-buffer-size response-buffer-size
-                                  :logger logger
-                                  :pipeline-transform pipeline-transform))
+                          (assoc opts
+                                 :early-response-d early-response-d
+                                 :ssl? ssl?
+                                 :ssl-context ssl-context
+                                 :ssl-endpoint-id-alg ssl-endpoint-id-alg
+                                 :remote-address remote-address
+                                 :raw-stream? raw-stream?
+                                 :response-buffer-size response-buffer-size
+                                 :logger logger
+                                 :pipeline-transform pipeline-transform))
 
         ch-d (netty/create-client-chan
               {:pipeline-builder    pipeline-builder
@@ -825,108 +827,110 @@
 
     (attach-on-close-handler ch-d on-closed)
 
-    (d/chain' ch-d
-              (fn setup-client
-                [^Channel ch]
-                (log/debug "Channel:" ch)
+    (d/alt'
+     early-response-d
+     (d/chain' ch-d
+               (fn setup-client
+                 [^Channel ch]
+                 (log/debug "Channel:" ch)
 
-                ;; We know the SSL handshake must be complete because create-client wraps the
-                ;; future with maybe-ssl-handshake-future, so we can get the negotiated
-                ;; protocol, falling back to HTTP/1.1 by default.
-                (let [pipeline (.pipeline ch)
-                      protocol (cond
-                                 ssl?
-                                 (or (-> pipeline
-                                         ^SslHandler (.get ^Class SslHandler)
-                                         (.applicationProtocol))
-                                     ApplicationProtocolNames/HTTP_1_1) ; Not using ALPN, HTTP/2 isn't allowed
+                 ;; We know the SSL handshake must be complete because create-client wraps the
+                 ;; future with maybe-ssl-handshake-future, so we can get the negotiated
+                 ;; protocol, falling back to HTTP/1.1 by default.
+                 (let [pipeline (.pipeline ch)
+                       protocol (cond
+                                  ssl?
+                                  (or (-> pipeline
+                                          ^SslHandler (.get ^Class SslHandler)
+                                          (.applicationProtocol))
+                                      ApplicationProtocolNames/HTTP_1_1) ; Not using ALPN, HTTP/2 isn't allowed
 
-                                 force-h2c?
-                                 (do
-                                   (log/info "Forcing HTTP/2 over cleartext. Be sure to do this only with servers you control.")
-                                   ApplicationProtocolNames/HTTP_2)
+                                  force-h2c?
+                                  (do
+                                    (log/info "Forcing HTTP/2 over cleartext. Be sure to do this only with servers you control.")
+                                    ApplicationProtocolNames/HTTP_2)
 
-                                 :else
-                                 ApplicationProtocolNames/HTTP_1_1) ; Not using SSL, HTTP/2 isn't allowed unless h2c requested
-                      setup-opts (assoc opts
-                                        :authority authority
-                                        :ch ch
-                                        :server? false
-                                        :keep-alive? keep-alive?
-                                        :keep-alive?' keep-alive?'
-                                        :logger logger
-                                        :non-tun-proxy? non-tun-proxy?
-                                        :pipeline pipeline
-                                        :pipeline-transform pipeline-transform
-                                        :raw-stream? raw-stream?
-                                        :remote-address remote-address
-                                        :response-buffer-size response-buffer-size
-                                        :ssl-context ssl-context
-                                        :ssl? ssl?)]
+                                  :else
+                                  ApplicationProtocolNames/HTTP_1_1) ; Not using SSL, HTTP/2 isn't allowed unless h2c requested
+                       setup-opts (assoc opts
+                                         :authority authority
+                                         :ch ch
+                                         :server? false
+                                         :keep-alive? keep-alive?
+                                         :keep-alive?' keep-alive?'
+                                         :logger logger
+                                         :non-tun-proxy? non-tun-proxy?
+                                         :pipeline pipeline
+                                         :pipeline-transform pipeline-transform
+                                         :raw-stream? raw-stream?
+                                         :remote-address remote-address
+                                         :response-buffer-size response-buffer-size
+                                         :ssl-context ssl-context
+                                         :ssl? ssl?)]
 
-                  (log/debug (str "Using HTTP protocol: " protocol)
-                             {:authority authority
-                              :ssl? ssl?
-                              :force-h2c? force-h2c?})
+                   (log/debug (str "Using HTTP protocol: " protocol)
+                              {:authority authority
+                               :ssl? ssl?
+                               :force-h2c? force-h2c?})
 
-                  ;; can't use ApnHandler, because we need to coordinate with Manifold code
-                  (let [http-req-handler
-                        (cond (.equals ApplicationProtocolNames/HTTP_1_1 protocol)
-                              (setup-http1-client setup-opts)
+                   ;; can't use ApnHandler, because we need to coordinate with Manifold code
+                   (let [http-req-handler
+                         (cond (.equals ApplicationProtocolNames/HTTP_1_1 protocol)
+                               (setup-http1-client setup-opts)
 
-                              (.equals ApplicationProtocolNames/HTTP_2 protocol)
-                              (do
-                                (http2/setup-conn-pipeline setup-opts)
-                                (http2-req-handler setup-opts))
+                               (.equals ApplicationProtocolNames/HTTP_2 protocol)
+                               (do
+                                 (http2/setup-conn-pipeline setup-opts)
+                                 (http2-req-handler setup-opts))
 
-                              :else
-                              (do
-                                (let [msg (str "Unknown protocol: " protocol)
-                                      e (IllegalStateException. msg)]
-                                  (log/error e msg)
-                                  (netty/close ch)
-                                  (throw e))))]
+                               :else
+                               (do
+                                 (let [msg (str "Unknown protocol: " protocol)
+                                       e (IllegalStateException. msg)]
+                                   (log/error e msg)
+                                   (netty/close ch)
+                                   (throw e))))]
 
-                    ;; Both Netty and Aleph are set up, unpause the pipeline
-                    (when (.get pipeline "pause-handler")
-                      (log/debug "Unpausing pipeline")
-                      (.remove pipeline "pause-handler"))
+                     ;; Both Netty and Aleph are set up, unpause the pipeline
+                     (when (.get pipeline "pause-handler")
+                       (log/debug "Unpausing pipeline")
+                       (.remove pipeline "pause-handler"))
 
-                    (fn http-req-fn
-                      [req]
-                      (log/trace "http-req-fn fired")
-                      (log/debug "client request:" (pr-str req))
+                     (fn http-req-fn
+                       [req]
+                       (log/trace "http-req-fn fired")
+                       (log/debug "client request:" (pr-str req))
 
-                      ;; If :aleph/close is set in the req, closes the channel and
-                      ;; returns a deferred containing the result.
-                      (if (or (contains? req :aleph/close)
-                              (contains? req ::close))
-                        (-> ch (netty/close) (netty/wrap-future))
+                       ;; If :aleph/close is set in the req, closes the channel and
+                       ;; returns a deferred containing the result.
+                       (if (or (contains? req :aleph/close)
+                               (contains? req ::close))
+                         (-> ch (netty/close) (netty/wrap-future))
 
-                        (let [t0 (System/nanoTime)
-                              ;; I suspect the below is an error for http1
-                              ;; since the shared handler might not match.
-                              ;; Should work for HTTP2, though
-                              raw-stream? (get req :raw-stream? raw-stream?)]
+                         (let [t0 (System/nanoTime)
+                               ;; I suspect the below is an error for http1
+                               ;; since the shared handler might not match.
+                               ;; Should work for HTTP2, though
+                               raw-stream? (get req :raw-stream? raw-stream?)]
 
-                          (if (or (not (.isActive ch))
-                                  (not (.isOpen ch)))
+                           (if (or (not (.isActive ch))
+                                   (not (.isOpen ch)))
 
-                            (d/error-deferred
+                             (d/error-deferred
                               (ex-info "Channel is inactive/closed."
                                        {:req     req
                                         :ch      ch
                                         :open?   (.isOpen ch)
                                         :active? (.isActive ch)}))
 
-                            (-> (http-req-handler req)
-                                (d/chain' (rsp-handler
+                             (-> (http-req-handler req)
+                                 (d/chain' (rsp-handler
                                             {:ch                   ch
                                              :keep-alive?          keep-alive? ; why not keep-alive?'
                                              :raw-stream?          raw-stream?
                                              :req                  req
                                              :response-buffer-size response-buffer-size
-                                             :t0                   t0})))))))))))))
+                                             :t0                   t0}))))))))))))))
 
 
 
--- a/src/aleph/http/http2.clj
+++ b/src/aleph/http/http2.clj
@@ -1251,16 +1251,12 @@
   [{:keys [^ChannelPipeline pipeline
            handler
            server?
-           proxy-options
            logger
            http2-stream-pipeline-transform
            max-request-body-size
            compression?
            compression-options]}]
   (log/trace "setup-stream-pipeline called")
-
-  (when (some? proxy-options)
-    (throw (IllegalArgumentException. "Proxying HTTP/2 messages not supported yet")))
 
   (cond-> pipeline
 


DerGuteMoritz:
FYI: There's a huge diff there due to wrapping + reindent, so I recommend to [hide whitespace](https://github.com/clj-commons/aleph/pull/742/files?diff=unified&w=1) when looking at the diff once more!

src/aleph/http/client.clj
```diff
@@ -396,7 +396,7 @@
 
                         :else
                         cause)]
-         (s/put! response-stream response)
+         (d/error! early-response-d response)
```
DerGuteMoritz, line 399: Oh yeah, so far, I was unable to even reach this code path and have it result in an error response anyway. It would probably be a good idea to come up with a bunch of test cases for the proxy functionality (currently there are none).

KingMob: Given how untested proxy code is, I'd strongly suggest writing tests before approving.

Maybe whoever needed the proxies could lend a hand?

DerGuteMoritz: I was able to come up with a test that fails as expected without the patch - sanity check welcome :pray: 

Using the test as a basis, I should be able to see how we can reach the error case referenced above. Stay tuned.

test/aleph/http_test.clj
```diff
@@ -1656,6 +1658,55 @@
...
         (is (thrown? RequestCancellationException (deref rsp 1000 :timeout)))))))
 
+(deftest test-client-proxy-support
+  (with-http-ssl-servers basic-handler {}
+    (let [proxy-request (d/deferred)
+          proxy-server (tcp/start-server
+                        (fn [stream _]
+                          (d/future
+                            (try
+                              (let [stream-ch (:aleph/channel (meta stream))
```
KingMob, line 1668: Doesn't have to be fixed here, but I always hated this. IIRC, sticking channels in metadata is the sole reason Manifold supports metadata abuse like this.

test/aleph/http_test.clj
```diff
@@ -1656,6 +1658,55 @@
...
+                                    req (http.core/netty-request->ring-request
+                                         @(d/timeout! (s/take! stream) 1000)
+                                         false
+                                         stream-ch
+                                         nil)]
+                                (assert (instance? LastHttpContent @(d/timeout! (s/take! stream) 1000)))
```
KingMob, line 1674: IIRC, LastHttpContent is Netty HTTP1-only. You'll need a new test if proxying is added to HTTP2.

KingMob (approved):
Looks ok to me, but I'm rusty, and never did anything with proxies even when I was working on Aleph, so take with a grain of salt :D

arnaudgeiser (approved):

src/aleph/http/client.clj
DerGuteMoritz: OK turns out there was a bit of breakage in the proxy connection error handling and propagation due to the various recent changes in the client response pipeline. Pushed 9673ac980e8b2eca8be8f6377c22a59f10b89155 for fixing those (see commit message for details) and added two proxy connection error test cases which indeed do hit those code paths now :sweat_drops:

test/aleph/http_test.clj
DerGuteMoritz: Ah yes, see also our conversation about this not too long ago [in the Clojurians Slack](https://clojurians.slack.com/archives/C02H9GF74CF/p1731042930920799?thread_ts=1730280543.361439&cid=C02H9GF74CF)

test/aleph/http_test.clj
DerGuteMoritz: The request towards the proxy server is indeed still HTTP/1 even for HTTP/2! Note how the test uses the `with-http-ssl-servers` macro which already exercises both protocol versions and it works just fine. This is because for TLS connections, the proxy acts as a dumb tunnel. In other words: beyond this point, it only shoves opaque bytes back and forth and has no clue about which HTTP version is contained within. Not sure whether this also works for h2c but I am willing to leave that for some adventurous soul to find out :yum:

KingMob: True...but what about the initial CONNECT? There's still one http request before shoveling bytes, and if the ALPN negotiates HTTP/2, it won't work. It should probably be documented to not allow http1 during ALPN if you're going to proxy.

DerGuteMoritz: The initial CONNECT request currently will always use (plain) HTTP/1 as there is no option to make it connect to the proxy itself via HTTPS (see [`proxy-options`](https://github.com/clj-commons/aleph/blob/9673ac980e8b2eca8be8f6377c22a59f10b89155/src/aleph/http.clj#L195-L202) where there would have to be an `ssl-context` option for this to work). Also, Netty's support for this [is still quite buggy](https://github.com/netty/netty/issues/11418) and [`HttpProxyHandler` is hardcoded to HTTP/1.1](https://github.com/netty/netty/blob/c1d0fd2dbf7907a99ab702795e35d9de8068c1cb/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java#L177-L180). So tl;dr I think this is fine? :fire: :coffee: :dog: :fire:

src/aleph/http/client.clj
```diff
@@ -473,9 +474,10 @@
 
    Can't use an ApnHandler/ApplicationProtocolNegotiationHandler here,
    because it's tricky to run Manifold code on Netty threads."
-  [{:keys [ssl? remote-address ssl-context ssl-endpoint-id-alg]}]
+  [{:keys [ssl? remote-address ssl-context ssl-endpoint-id-alg proxy-options proxy-connected]}]
   (fn pipeline-builder*
     [^ChannelPipeline pipeline]
+    (add-proxy-handlers pipeline proxy-connected proxy-options ssl?)
```
KingMob, line 480: Should be after the SSL handlers, for security, no?

test/aleph/http_test.clj
KingMob: Ahh, I hadn't realized Aleph only supported unencrypted client<->proxy connections. I think ssl support should be mandatory, but I defer to you. 

FWIW, I'm pretty sure think we'll also have the problem that netty issue did, with putting two ssl handlers in the pipeline. We have some code that looks for the handler by class instead of name, expecting only one.

---

To me, not supporting an encrypted client<->proxy connection isn't safe, since a client may not be encrypting the proxied stream (e.g. when the proxy is the gateway between the internet and a trusted VPC). 

The point I was trying to raise is that, when Aleph acts as a secure http proxy (not the client), ALPN will occur before the CONNECT is sent, so it might negotiate http2 before it could learn that won't work.

### Suggested changes
#### When Aleph is  the client
When Aleph is the client, it should really support an encrypted client<->proxy connection. 

Optionally, we could add proxy-specific ALPN options in `proxy-options` that default to accepting _only_ http1, for better compatibility when the client and proxy are both Aleph. 

#### When Aleph is the proxy 
We should support ssl. If we only support it with http1, we should probably:
1. Return an informative error and kill the connection if (1) https and ALPN are used, (2) a client negotiated http2, and then (3) sent a CONNECT
2. Document these limits.

DerGuteMoritz: > Ahh, I hadn't realized Aleph only supported unencrypted client<->proxy connections.

Heh yeah I only learned about the ins and outs of the client proxy support myself while looking into this fix, never used the feature myself so far.

> I think ssl support should be mandatory, but I defer to you.

I wouldn't go that far - AFAIK HTTP proxies are mainly used from within trusted networks for connecting to the outside world and might not even offer HTTPS. So I'd leave that as optional just like we do for direct connections. @David-Ongaro Do you have a perspective on this, too?

> To me, not supporting an encrypted client<->proxy connection isn't safe, since a client may not be encrypting the proxied stream (e.g. when the proxy is the gateway between the internet and a trusted VPC).

I don't quite understand your point here. A client would always encrypt the proxied stream when the destination address is `https` (since the proxy only acts as a tunnel, not as an endpoint of a TLS session in this case). Only the initial `CONNECT` request towards the proxy is unencrypted in that scenario. This means a MITM could tamper with the host being connected to but then hostname verification should fail and the connection should be aborted :crossed_fingers:. OK and there's the possibility of MITMs snooping which hosts clients are connecting to. But given that the network context is assumed to be trusted, both issues should be no problem in practice (except for defense in depth, of course, which is why I still find it worth supporting).

> The point I was trying to raise is that, when Aleph acts as a secure http proxy (not the client), ALPN will occur before the CONNECT is sent, so it might negotiate http2 before it could learn that won't work.

Aaah :bulb: Right, using Aleph as a proxy *server* is quite a different situation and there's no first-class support for that really (note how Iow-level I had to go in the test). IMHO first-class proxy support is out of scope for Aleph's HTTP server. There's enough special-purpose implementations of this already, I don't see how a (customized?) Clojure-based adds much to the table.

In any case, I would like to keep this PR focused on restoring the status quo ante and created https://github.com/clj-commons/aleph/issues/743 for adding HTTPS proxy support. Thanks your input and for thinking along! :pray:

src/aleph/http/client.clj
DerGuteMoritz: As in `.addAfter`? Or code ordering wise? The former won't do the right thing and the latter I'm not sure makes a difference since the pipeline builder is run atomically before anything else happens. Since the proxy handlers are added to the front of the pipeline (via `.addFirst`) and the SSL handler is added to the back, I would tend to leave the code ordering as it is since it mirrors the pipeline order.

test/aleph/http_test.clj
KingMob: > A client would always encrypt the proxied stream when the destination address is https (since the proxy only acts as a tunnel, not as an endpoint of a TLS session in this case).

Sure, but what if the destination address is insecure http? The whole session will be in cleartext. I think the default should be encryption in 2025.

> IMHO first-class proxy support is out of scope for Aleph's HTTP server. There's enough special-purpose implementations of this already, I don't see how a (customized?) Clojure-based adds much to the table.

I don't recall, what will an Aleph server do at the moment when it gets a CONNECT?

I agree there's no particular need for Aleph to do this, though, other than for completeness. I doubt many servers are simultaneously Clojure web apps and proxies.

src/aleph/http/client.clj
KingMob: > Since the proxy handlers are added to the front of the pipeline (via .addFirst)

Dammit. This is one of the most annoying things about netty. I thought we consistentnly used .addLast.

I guess I should say, if we supported encrypted client<->proxy connections, we'd have to ensure there was a second SslHandler before the proxy handlers in the pipeline.

test/aleph/http_test.clj
DerGuteMoritz: > Sure, but what if the destination address is insecure http? The whole session will be in cleartext. I think the default should be encryption in 2025.

Right, in that case the best you could do is to encrypt the client <-> proxy connection while the proxy <-> destination connection would remain unencrypted. I don't see how we could change anything else about this within the scope of Aleph? Seems more like job for https://www.eff.org/https-everywhere :smile: 

> I don't recall, what will an Aleph server do at the moment when it gets a CONNECT?

Interesting question. AFAICT it treats it as any other HTTP method and just passes it to the ring handler. For example, when starting a server with  `aleph.http-test/basic-handlers`, this is what an exchange using `curl` looks like:

```
$ curl -v -X CONNECT http://localhost:8082/string
* Host localhost:8082 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8082...
* Connected to localhost (::1) port 8082
* using HTTP/1.x
> CONNECT /string HTTP/1.1
> Host: localhost:8082
> User-Agent: curl/8.11.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Server: Aleph/0.8.2
< Date: Mon, 24 Feb 2025 15:35:02 GMT
< Connection: Keep-Alive
< content-length: 7
< 
* Connection #0 to host localhost left intact
String!
```

Given that Aleph or Ring doesn't enforce semantics of HTTP methods in any capacity, I think this is basically fine?

However, using Aleph's HTTP client to send a `CONNECT` request directly (i.e. not by setting a proxy) *does* seem to trigger some special handling in Netty. This results in the following error:

```
610929 tid=459 WARN client - error in HTTP client
java.lang.ClassCastException: io.netty.buffer.PooledUnsafeDirectByteBuf cannot be cast to io.netty.handler.codec.DecoderResultProvider
	at aleph.http.common$decoder_failed_QMARK_.invokeStatic(common.clj:237)
	at aleph.http.common$decoder_failed_QMARK_.invoke(common.clj:237)
	at aleph.http.client$http1_client_handler$reify__19484.channelRead(client.clj:209)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
```

Not sure why anyone would do that but given that we do expose [`aleph.http/connect`](https://github.com/clj-commons/aleph/blob/9673ac980e8b2eca8be8f6377c22a59f10b89155/src/aleph/http.clj#L550) as a first-class API, we probably shouldn't blow up like that :sweat_smile:  Filed as https://github.com/clj-commons/aleph/issues/745

src/aleph/http/client.clj
DerGuteMoritz: > Dammit. This is one of the most annoying things about netty. I thought we consistentnly used .addLast.

Ah, I see! OK AFAICT it doesn't make a difference whether we use `.addFirst` or `.addLast` in this particular case since the proxy handlers are the first thing to be added anyway. I've changed it accordingly to reduce mental overhead for future readers :smile: 

> I guess I should say, if we supported encrypted client<->proxy connections, we'd have to ensure there was a second SslHandler before the proxy handlers in the pipeline.

Indeed! Updated the description of https://github.com/clj-commons/aleph/issues/743 to mention this more explicitly.

test/aleph/http_test.clj
KingMob: >Right, in that case the best you could do is to encrypt the client <-> proxy connection while the proxy <-> destination connection would remain unencrypted. 

Yes, this is what I'm arguing for. If Aleph is the proxy, we can't control how the client wants to talk to the ultimate endpoint, but we should be encrypting the client<->proxy connection as much as possible. 

----

>AFAICT it treats it as any other HTTP method and just passes it to the ring handler.

Probably for the best. I've read about custom CONNECT implementations that do complicated but transparent proxying, and handling CONNECT yourself seems like the way to do that if you want. 

But default proxy behavior (forward the bytes) should probably be opt-in when starting an Aleph server, otherwise, that's a massive security risk.

DerGuteMoritz: > Yes, this is what I'm arguing for. If Aleph is the proxy, we can't control how the client wants to talk to the ultimate endpoint, but we should be encrypting the client<->proxy connection as much as possible.

Agreed. As we have https://github.com/clj-commons/aleph/issues/743 for this now, gonna resolve this thread!

> Probably for the best. I've read about custom CONNECT implementations that do complicated but transparent proxying, and handling CONNECT yourself seems like the way to do that if you want.

:+1: 

> But default proxy behavior (forward the bytes) should probably be opt-in when starting an Aleph server, otherwise, that's a massive security risk.

Ah yes, open proxies on the public internet are indeed a big no-no. Thanks for thinking along once more!

src/aleph/http/client.clj
DerGuteMoritz: Resolved as per you guys' approval of the changes :pray:

bitti:
Thanks for the great effort! I'm sorry that I didn't have time to contribute some tests. Maybe next time.

Also congragulations to the 10²th tag and the 19²th closed issue! These are nice round numbers.

PR #744, opened 2025-02-14
Update Netty to 4.1.118.Final

DerGuteMoritz:
Fixes [CVE-2025-24970](https://github.com/netty/netty/security/advisories/GHSA-4g8c-wm8x-jfhw).

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -10,24 +10,24 @@
   org.clj-commons/dirigiste {:mvn/version "1.0.4"},
   org.clj-commons/primitive-math {:mvn/version "1.0.1"},
   potemkin/potemkin {:mvn/version "0.4.7"},
-  io.netty/netty-transport {:mvn/version "4.1.115.Final"},
+  io.netty/netty-transport {:mvn/version "4.1.118.Final"},
   io.netty/netty-transport-native-epoll$linux-x86_64
-  {:mvn/version "4.1.115.Final"},
+  {:mvn/version "4.1.118.Final"},
   io.netty/netty-transport-native-epoll$linux-aarch_64
-  {:mvn/version "4.1.115.Final"},
+  {:mvn/version "4.1.118.Final"},
   io.netty/netty-transport-native-kqueue$osx-x86_64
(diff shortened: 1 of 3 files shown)


arnaudgeiser (approved):

KingMob (approved):
Did you see anything else in the Netty changelog that might impact Aleph?

DerGuteMoritz:
> Did you see anything else in the Netty changelog that might impact Aleph?

It was all seemingly innocuous bug fixes BUT the tests just failed horribly in CI. Just scheduled a retry but I think there might be more to it this time :grimacing: Stay tuned.

DerGuteMoritz:
OK please see https://github.com/clj-commons/aleph/pull/744/commits/1b433ab090c26d19cf6822046c0c0b222705413c - it's technically a breaking change but it didn't even make the list list of [most important changes](https://netty.io/news/2024/12/17/4-1-116-Final.html) for the Netty release that introduced it. Maybe we should still call it out in ours, though.

KingMob:
I see you updated the test, but no code. What's the breaking change for users?

Does https://github.com/netty/netty/pull/14519 affect things somehow?

KingMob:
Ahh! 

But iiuc, current netty only enables brotli and zstd if it finds them, right? If the user has them on the path, they probably want them anyway (or were at least willing), and if they're not on the path, nothing would break, correct?

So from the user perspective, this is the mildest of breaking changes, hopefully. Still, I agree it it should be documented.

DerGuteMoritz:
> But iiuc, current netty only enables brotli and zstd if it finds them, right? If the user has them on the path, they probably want them anyway (or were at least willing), and if they're not on the path, nothing would break, correct?

That's should indeed be the case, good point!

 
> So from the user perspective, this is the mildest of breaking changes, hopefully. Still, I agree it it should be documented.

Yeah, our test at least was very explicit about this expectation so maybe some users will hit it, as well. I'll make a note about it in the changelog then!

Thanks for your feedback, merging!

Issue #728, opened 2024-06-13
Using InputStream as :ssl-context :trust-store is buggy


DerGuteMoritz:
Quoting the original report by @David-Ongaro from the addendum of https://github.com/clj-commons/aleph/issues/727:

> As per the docs, instead of `java.io.File` instances, `java.io.InputStream` instances are also supported as keys for the `:ssl-context` map. But I can't figure out how this is supposed to be used, as I regularly get `java.lang.IllegalArgumentException: Input stream does not contain valid certificates.` exceptions. I.e., preparing a single request just works fine, but preparing them in quick succession may fail:
> 
> ```clojure
> (def ssl-context {:trust-store (io/input-stream client-ca)})
> 
> (def pool (http/connection-pool {:connection-options {:ssl-context ssl-context}}))
> 
> (http/get "https://example.com" {:pool pool}) => #<Deferred@6fb6283a: :not-delivered>
> 
> [(http/get "https://example.com" {:pool pool}) (http/get "https://example.com" {:pool pool})] =>
> [#<Deferred@58ac8429: Error printing return value (CertificateException) at io.netty.handler.ssl.PemReader/readCertificates (PemReader.java:114).
> Error printing return value (CertificateException) at io.netty.handler.ssl.PemReader/readCertificates (PemReader.java:114).
> found no certificates in input stream
>             PemReader.java:  114  io.netty.handler.ssl.PemReader/readCertificates
>            SslContext.java: 1263  io.netty.handler.ssl.SslContext/toX509Certificates
>     SslContextBuilder.java:  276  io.netty.handler.ssl.SslContextBuilder/trustManager
>                  netty.clj:  917  aleph.netty/eval23500/add-ssl-trust-manager!
>                  netty.clj: 1026  aleph.netty/eval23500/ssl-client-context
>                  netty.clj: 1193  aleph.netty/coerce-ssl-context
>                  netty.clj: 1179  aleph.netty/coerce-ssl-context
>                   core.clj: 2641  clojure.core/partial/fn
>                 client.clj:  699  aleph.http.client/client-ssl-context
>                 client.clj:  690  aleph.http.client/client-ssl-context
>                 client.clj:  799  aleph.http.client/http-connection
>                 client.clj:  752  aleph.http.client/http-connection
>                   http.clj:  104  aleph.http/create-connection
>                   http.clj:   97  aleph.http/create-connection
>                   http.clj:  239  aleph.http/connection-pool/fn
[...]

PR #746, opened 2025-03-01
Lift `ssl-context` coercion to `connection-pool` fn

PawelStroinski:
Addresses #728.


DerGuteMoritz:
Hey @PawelStroinski, sorry for the delay! Responding to the main commit message:

> Keeping the `client-ssl-context` call in `http-connection` as is,
> even though it might seem superfluous considering the code path taken in
> the test, but `http-connection` is a public API, so we have to keep
> the call (which for us is a no-op, if we ignore the repeated ALPN check)
> even for our case when the protocol is https and `ssl-context` is supplied.

Hmm I wonder how public that API is really - note how the `aleph.http.client` has `^:no-doc` metadata :thinking: Also, the docstring doesn't explain any of the options and there are no dedicated tests for it either. So IMHO we can consider this an internal API in practice and drop the `client-ssl-context` call here.

 
> NOTE: This highlights a difference we are introducing here. Previously,
> if we specified ssl-context, but the protocol wasn't https, we would just
> ignore the ssl-context. Currently, we are coercing it ahead-of-time,
> before knowing the request protocol. This could be alleviated by wrapping
> the coercion in a `delay`, so it wouldn't happen until needed. Yet, given
> how unlikely this scenario seems, I have doubts whether it'd be worth it.

I agree. If anyone runs into it, they can easily fix it since such a combination of options doesn't make sense to begin with :sweat_smile: 


> I slightly dislike the repetition of `[:http1]` default value,
> but since it serves as documentation in `http-connection`,
> I decided to keep it as is rather than to extract it out.

See above - we can get rid of this then, too.


> Also, I slightly dislike the repetition of a pattern to call
> `ensure-consistent-alpn-config` and then `coerce-ssl-client-context`
> but it's only now in 2 places, which I think is a better alternative
> than adding yet another ssl-coercion layer/wrapping function.
> Obviously, we cannot just move `ensure-consistent-alpn-config` to
> `ssl-client-context`, since ALPN is only for HTTP.

Technically, ALPN can also be used for other purposes but in the context of Aleph's out-of-the-box support, this is true :smile: But as per above, we can get rid of this then, too.

Thanks for your thoughtful notes as always! :pray:

PawelStroinski:
Thank you for the feedback, @DerGuteMoritz. Certainly, `http-connection` didn't even have a docstring until a recent (well... relatively, considering long Aleph history) refactoring and the `^:no-doc` metadata on that namespace is a good hint regarding the intended usage (or lack of it), indeed. I've updated my PR taking advantage of that fact.

PR #749, opened 2025-04-09
Fix typos

bitti:
I thought if nothing else, I at least can contribute some spelling fixes. (Only one of these affect actual code). I guess that they are so few and hard to find speaks to the quality of the project.


arnaudgeiser (approved):
Thank you David!

PR #754, opened 2025-06-18
Update dependencies

DerGuteMoritz:
This pulls in Netty 4.1.122 - update to 4.2.x pending.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,7 +1,7 @@
 ;; DO NOT EDIT MANUALLY - generated from project.clj via deps/lein-to-deps
 {:paths ["src" "resources" "target/classes"],
  :deps
- {org.clojure/clojure {:mvn/version "1.11.3"},
+ {org.clojure/clojure {:mvn/version "1.12.1"},
   org.clojure/tools.logging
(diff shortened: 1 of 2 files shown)


DerGuteMoritz:
> A good old Renovate could be useful at some point for the project. Compared to dependabot, it's working with Clojure (both lein and deps.edn)

Indeed. Currently, I am subscribed to Netty updates (which is the only critical dependency we have really). When there's a security fix, I'll immediately update, otherwise I'll opportunistically do it whenever I am about to tag a new release. But automating this in general would indeed be convenient. So if you happen to have some experience with this, patches always welcome :pray:

arnaudgeiser:
> But automating this in general would indeed be convenient. So if you happen to have some experience with this, patches always welcome 🙏

We probably need to ask Erik Aslum(?) to enable Renovate on our project. If I remember correctly, that's something that has to be done at the organization level. Once done, it will create an onboarding PR.

DerGuteMoritz:
> We probably need to ask Erik Aslum(?) to enable Renovate on our project. If I remember correctly, that's something that has to be done at the organization level. Once done, it will create an onboarding PR.

Ah, very good! That would be @slipset then :wave: Would you be able to? :pray:

PR #778, opened 2026-02-08
fix(req-headers): Set headers instead of adding duplicates when key is a singleton 

mpenet:
The HTTP specifications define (and in some cases hints at) several headers as singletons (should appear only once) in responses, and aleph will create duplicates if we have multiple keys with the different casing. 
While I understand the user should be careful, the server could also prevent that to happen.


mpenet:
I am tempted to close it. Maybe it's just a case of "garbage-in, garbage-out". I ll let you decide.

PR #779, opened 2026-02-16
Update Netty to 4.2.10.Final

### LINKED ISSUE #756: KQueue class loading issues with netty 4.2.2.Final
I realize the aleph repo isn't officially using 4.2.x yet, but aleph works perfectly for our use cases on 4.2.1.Final at least so figured it'd be helpful to report this issue here.

With 4.2.2.Final, building uberjars on Linux systems throws errors like this:
```
Syntax error macroexpanding def at (netty.clj:1326:1).
Execution error (IllegalStateException) at io.netty.channel.kqueue.Native/loadNativeLibrary (Native.java:150).
Only supported on OSX/BSD
```

Looks like the issue is that these functions force KQueue classes to be loaded at build time:
https://github.com/clj-commons/aleph/blob/bb3a1f1a10acb50ff68b87185f04d3f5cd2eccc4/src/aleph/netty.clj#L1327-L1332
https://github.com/clj-commons/aleph/blob/bb3a1f1a10acb50ff68b87185f04d3f5cd2eccc4/src/aleph/netty.clj#L1398-L1410

This starts failing on 4.2.2.Final because that version added some new static fields to `AbstractKQueueChannel` that force io.netty.channel.kqueue.Native` to be loaded here, even on non-OSX systems.

I've reported the root cause upstream in https://github.com/netty/netty/issues/15392 already since this seems like a netty regression (and they've fixed similar issues in the past).

While I wait on a response from the netty maintainers, would the maintainers here be open to a PR to guard against issues like this? 

I've worked around this locally by replacing the class names in the functions above with calls like `(eval 'io.netty.channel.kqueue.KQueueServerSocketChannel)` to ensure that the classes aren't loaded until they're actually needed, but wasn't sure on the right approach to adding a test here since this issue doesn't show up in the netty version currently set in `project.clj`.

### LINKED ISSUE #760: Update to Netty 4.2.x
See https://github.com/netty/netty/wiki/Netty-4.2-Migration-Guide

robinlahtinen:
This pull request upgrades Aleph to `Netty 4.2.10.Final`, modernizes transport support, and updates SSL certificate logic for compatibility with newer JDKs.

Resolves #760 and fixes #756.

## Changes

**Netty version upgrade and transport improvements**

* Upgrades Netty dependencies from `4.1.130.Final` to `4.2.10.Final`. **This is a breaking change.**
* Promotes the `io_uring` transport from the incubator package to the stable `io.netty` package and updates imports.
* Adds `kqueue` native transport support for macOS on Apple Silicon (`osx-aarch_64`).
* Refactors event loop group and channel management to use the new `IoHandler` and `MultiThreadIoEventLoopGroup` APIs in Netty 4.2.
* Implements lazy initialization for transport availability checks to resolve KQueue class-loading issues.

**SSL context and certificate generation**

* Updates `self-signed-ssl-context` to use `CertificateBuilder` for compatibility with JDK 24+.
* Retains fallback support for `SelfSignedCertificate` on older JDKs.
* Disables default HTTPS endpoint identification in the Netty 4.2 SSL context builder, as Aleph handles hostname verification independently.

**Deprecations and cleanup**

* Removes usage of the deprecated `ChannelOption/MAX_MESSAGES_PER_READ` option.
* Updates Bouncy Castle exclusions to prevent dependency conflicts.

**Build and testing**

* Adds a `:test-unsafe-deny` profile for testing environments with strict memory access flags.
* Improves leak detector reliability by using the `WeakReference` sentinel pattern in `force-leak-detection!` for deterministic GC confirmation across JDK 8 and JDK 21.
* Updates `CHANGES.md` with a summary of breaking changes.


robinlahtinen:
Note on the single failing check: `patch_hold_test_with_dropped_error_deferred_detection_status`.

I don't think this is an actual test failure, but I could be mistaken. It appears to be the CircleCI job that patches the GitHub commit status for the manual approval gate, and it seems to be getting a `401 Unauthorized` response from the GitHub API. My guess is that the `GITHUB_ACCESS_TOKEN` configured in CircleCI may have expired or no longer has the `repo:status` scope.

As far as I can tell, all the actual test jobs are green (leak detection and dependency cache on JDK 8 and 21).

Would refreshing the token in CircleCI likely resolve this? If helpful, I could also add a small guard so a missing or expired token doesn't fail the build.

Please let me know if you think anything is needed on the code side, or if I'm missing something.

.circleci/config.yml
```diff
@@ -72,6 +84,36 @@ jobs:
...
+            - ~/.m2
+          key: v2-dependencies-jdk_21-{{ checksum "project.clj" }}
+  test_with_leak_detection_jdk_21:
+    <<: *job_defaults_jdk_21
+    steps:
+      - checkout
+      - restore_deps_cache
+      - run:
+          name: Run tests with leak detection (JDK 21)
+          command: lein do clean, with-profile +leak-detection,+jdk-21 test :default+leak
```
DerGuteMoritz, line 116: Ah very good, updating the CI pipeline to use more recent versions was on my radar, as well. Ideally, we'd use [matrix jobs](https://circleci.com/docs/guides/orchestrate/using-matrix-jobs/) for testing combinations of supported Clojure and JDK versions but we can do that in a follow-up. Filed https://github.com/clj-commons/aleph/issues/780 as a placeholder.

src/aleph/http/http2.clj
```diff
@@ -1399,6 +1399,21 @@
...
+    ;; channelActive is never replayed, so the flushPreface guard in
+    ;; PrefaceDecoder.channelActive never fires. Without this explicit flush,
+    ;; the preface sits in SslHandler.pendingUnencryptedWrites and under
+    ;; certain allocator/timing conditions (e.g., unpooled allocator on
+    ;; JDK 21), a SETTINGS ACK can be flushed before the initial SETTINGS,
+    ;; causing the peer to reject the connection.
+    ;; See https://github.com/netty/netty/issues/12089
+    (let [ch (.channel pipeline)]
+      (when (.isActive ch)
+        (.flush ch)))
```
DerGuteMoritz, line 1415: Hm, how did you figure this out and are you sure this is indeed a possible explanation? FYI: I've shared this with Netty upstream and [they are not convinced](https://github.com/netty/netty/issues/16142#issuecomment-4067796527) so far 🤔 

Note that we have https://github.com/clj-commons/aleph/issues/772 to track this particular issue.

project.clj
```diff
@@ -66,6 +70,8 @@
                                          "-Dio.netty.allocator.type=unpooled"]}
              :dropped-error-deferred-detection {:jvm-opts ["-Dorg.slf4j.simpleLogger.log.manifold.debug=warn"
                                                            "-Daleph.testutils.detect-dropped-error-deferreds=true"]}
+             :test-unsafe-deny    {:jvm-opts ["--sun-misc-unsafe-memory-access=deny"]}
+             :jdk-21              {:javac-options ^:replace ["--release" "11"]}
```
DerGuteMoritz, line 74: What's the reason for this?

src/aleph/netty.clj
```diff
@@ -1164,21 +1168,79 @@
...
   "A self-signed SSL context for servers.
 
    Never use in production. Even if you control all clients, and want to use
-   a self-signed cert internally, do not use this fn, because Netty's
-   SelfSignedCertificate class is only for testing, and uses an insecure PRNG."
+   a self-signed cert internally, do not use this fn, because the generated
+   certificate uses weak algorithms and is only for testing.
+
+   Prefers Netty 4.2's `CertificateBuilder` from `io.netty/netty-pkitesting`
+   when available (works on all JDKs including 24+). Falls back to Netty's
```
DerGuteMoritz, line 1221: No need to keep around the `SelfSignedCertificate` fallback - AFAIUI `io.netty/netty-pkitesting` also works with older JDK versions, right? See also [my other comment](https://github.com/clj-commons/aleph/pull/779/changes#r2940663969).

src/aleph/netty.clj
```diff
@@ -1164,21 +1168,79 @@
...
+          get-cert-m    (.getMethod bundle-class "getCertificate" (into-array Class []))
+          certificate   (.invoke get-cert-m bundle (object-array []))]
+      (log/debug "Using CertificateBuilder from netty-pkitesting for self-signed cert")
+      {:private-key       priv-key
+       :certificate-chain [certificate]})
+    (catch ClassNotFoundException _
+      nil)
+    (catch Exception e
+      (log/debug e "CertificateBuilder reflection failed, falling back to SelfSignedCertificate")
+      nil)))
```
DerGuteMoritz, line 1211: Since this will be a breaking change release anyway, I suggest we move `self-signed-ssl-context` into a separate namespace `aleph.netty.test` which people can separately load when needed and then be expected to provide `io.netty/netty-pkitesting`. That saves us from having to do this cumbersome reflection dance and also makes it more explicit that this function is not intended for production use at the same time.

src/aleph/netty.clj
```diff
@@ -1287,29 +1365,51 @@
...
+   creates its own threads using the factory."
+  ^MultiThreadIoEventLoopGroup [transport ^long thread-count thread-factory-or-executor]
+  (let [^IoHandlerFactory handler-factory
+        (case transport
+          :nio     (NioIoHandler/newFactory)
+          :epoll   (EpollIoHandler/newFactory)
+          :kqueue  (KQueueIoHandler/newFactory)
+          :io-uring (let [cls (Class/forName "io.netty.channel.uring.IoUringIoHandler")
+                          m   (.getMethod cls "newFactory" (into-array Class []))]
+                      (.invoke m nil (object-array []))))]
```
DerGuteMoritz, line 1381: I suggest to change it so that we make the caller pass in the factory instance directly instead of re-dispatching on the `transport` key here, thus keeping `create-io-event-loop-group` decoupled from the transport-specific shenanigans.

test/aleph/http/pipeline_stress_test.clj
```diff
@@ -0,0 +1,146 @@
...
+    {:connection-options {:insecure?     true
+                          :http-versions [:http2 :http1]}}))
+
+(defmacro ^:private with-stress-server
+  "Starts a server with the given handler and options, runs body, then cleans up."
+  [handler opts & body]
+  `(let [server# (http/start-server ~handler (merge server-options ~opts))
+         pool# (make-pool)]
+     (try
+       (let [~'pool pool#]
```
DerGuteMoritz, line 60: Holy hygiene breakage 🦇 😬 Please pull out into a dynamic variable like in `aleph.http-test` or make it an explicit binding passed in by the caller 🙏 In fact, couldn't you simply move these tests to `aleph.http-test`, as well? I appreciate the impetus to split this monster test namespace up but if we do it, we should use a common set of test utilities.

test/aleph/http/pipeline_stress_test.clj
```diff
@@ -0,0 +1,146 @@
...
+  (testing "N concurrent HTTP/2 requests over a single connection"
+    (with-stress-server echo-handler {}
+                        (let [n 100
+                              url (str "https://localhost:" stress-port "/echo")
+                              results (doall
+                                        (for [i (range n)]
+                                          (http/get url
+                                                    {:pool      pool
+                                                     :insecure? true
+                                                     :body      (str "req-" i)})))]
```
DerGuteMoritz, line 77: Note that this will max out at 8 concurrent requests due to the [default `connections-per-host`](https://github.com/clj-commons/aleph/blob/142f009e820078d4e6a85186709e7114d152901f/src/aleph/http.clj#L216). Might be worth making explicit here.

test/aleph/resource_leak_detector.clj
```diff
@@ -91,9 +92,27 @@
...
+
+(def max-gc-polls
+  "Maximum number of GC polls before giving up in force-leak-detection!."
+  10)
+
 (defn force-leak-detection! []
-  (System/gc)
-  (System/runFinalization)
+  ;; Use a WeakReference sentinel to confirm GC has actually run and
+  ;; processed references. This is the standard OpenJDK ForceGC pattern.
```
DerGuteMoritz, line 106: Interesting, I didn't come across this pattern during my research for the initial implementation of the leak detector. In what sense is it a "standard"? A quick web search pointed me to [a `jlibs` project](https://github.com/santhosh-tekuri/jlibs/blob/59c28719f054123cf778278154e1b92e943ad232/core/src/main/java/jlibs/core/lang/RuntimeUtil.java#L148) but that seems more like that guy's bag of utils.

In any case, AFAICT it basically amounts to the same idea, just leaking a `WeakReference` instead of a Netty buffer? What do we gain from doing this in addition? Shouldn't it also suffice to just bump `max-probe-gc-runs` instead to achieve better detection probability when needed?

CHANGES.md
```diff
@@ -1,3 +1,18 @@
+### Unreleased
+
+* BREAKING: Bump Netty to 4.2.10.Final
+* Promote io_uring transport from incubator (`io.netty.incubator` → `io.netty`)
+* Load io_uring transport classes lazily via reflection (requires Java 9+; gracefully unavailable on Java 8)
+* Add kqueue `osx-aarch_64` native transport for Apple Silicon
+* Update `self-signed-ssl-context` for JDK 24+ compatibility via Netty `CertificateBuilder`, with `SelfSignedCertificate` fallback
+* Fix deprecated `Class.newInstance()` usage in `CertificateBuilder` reflection (use `Constructor.newInstance()`)
```
DerGuteMoritz, line 8: No need to mention a fix for something that was never released.

DerGuteMoritz (changes requested):
Thanks a lot for your contribution 🙏  Let me know if you have the capacity to address the feedback - otherwise I'll see to it myself!

DerGuteMoritz:
> Note on the single failing check: `patch_hold_test_with_dropped_error_deferred_detection_status`.
> 
> I don't think this is an actual test failure, but I could be mistaken. It appears to be the CircleCI job that patches the GitHub commit status for the manual approval gate, and it seems to be getting a `401 Unauthorized` response from the GitHub API. My guess is that the `GITHUB_ACCESS_TOKEN` configured in CircleCI may have expired or no longer has the `repo:status` scope.

Ah I think the problem is that the environment variable is only available when the pipeline was triggered by a branch on the main repository or so. I just tried to manually re-run it. But yeah, this is indeed not a test failure. It only serves as a workaround to mark the pipeline as completed. Otherwise, it would be indefinitely pending. See https://federicoterzi.com/blog/solving-github-status-stuck-on-pending-with-circlecis-approvals/ if you are interested in the gory details :sweat_smile: 


> As far as I can tell, all the actual test jobs are green (leak detection and dependency cache on JDK 8 and 21).

:+1: 


> If helpful, I could also add a small guard so a missing or expired token doesn't fail the build.

Right, we could make it not fail in this case but then the pipeline would instead be considered pending (see above). Not sure if that's really an improvement :shrug:

PR #782, opened 2026-04-08
Fix HTTP client response body coercion for error status responses

DerGuteMoritz:
When using the HTTP client's response body coercion middleware (via the `:as` request parameter) in combination with `:throw-exceptions? true` (the default), one would run into an error when the coercion result is not a manifold stream or byte array. Specifically, the commonly used `:as :string` would result in such an error because the error status handler would assume a manifold stream as the default case.

This patch fixes this by turning manifold streams into a special case and passing all other response body types through untouched. A basic dedicated test case for the response body coercion middleware is included (there was none so far) while the offending interaction with `:throw-exceptions? true` is covered by `test-client-throw-error-responses-as-exceptions`.

### DIFF
--- a/src/aleph/http/client_middleware.clj
+++ b/src/aleph/http/client_middleware.clj
@@ -278,15 +278,15 @@
                   (error-status-deferred
                    (assoc rsp :body (ByteArrayInputStream. body)))))
 
-      (nil? body)
-      (error-status-deferred rsp)
-
-      :else
+      (s/stream? body)
       (d/chain'
        (s/reduce conj [] body)
        (fn [body]
          (error-status-deferred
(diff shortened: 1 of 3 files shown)


src/aleph/http/client_middleware.clj
buhhh: If I am not mistaken, this can be removed as well including the corresponding destructuring – it seems, it dates back to a prior `:default` method. It's somewhat related to this PR, but, of course, not actually touched here.

DerGuteMoritz: Ah indeed, well spotted. Will drop the dead branch while at it :-)

